PR #259, opened 2015-01-26
Pe sorption

Paulie14:
Exception introduced in sorption:
- input exception in record reading
- exceptions in isotherm - boost solver and negative total mass


Paulie14:
Feel free to delete PE_sorption branch after merging these changes.

Issue #270, opened 2015-02-04
Adjust time step of TransportDG to output times and time-dependent input fields


jstebel:
Similarly as in DarcyFlow and ConvectionTransport, update appropriate time marks to respect times when output is generated and the input fields which are changed in time.

jbrezmorf:
Merged in  10d23cb.

Issue #146, opened 2014-04-11
TimeGovernor:: init_dt behavior


jbrezmorf:
First problem 
Current behavior is:
no init_dt - time step choosen by equation 
given init_dt = fixed dt (override fixed times from input)
given init_dt and max_dt or min_dt - variable dt, init_dt just suggestion

This is misleading since init_dt should be interpreted as a suggestion of initial time step. This is useful in for equations with some sort of time step adaptivity.

Solution:
Use init_dt only as initial time step.
Max_dt, is used for two purposes: 
1) max_dt is used as first suggested dt after avery fixed time.
2) max_dt limits any choice of dt given by equation min_dt ( if the suggstion can only decrease time step than the second is implied by the first )
Always match input time and output time marks possibly shorten timestep below min_dt limit.

* implement this behaviour
* Make clear documentation.
* Integration tests for various setting.

jbrezmorf:
Merged in 10d23cb.
test of Init_dt is part test 14, short_pulse_implicit.

Issue #287, opened 2015-02-15
apply new time governor interface


jbrezmorf:
- [ ] make separate unit test for time marks
- [ ] minimize time governor interface using step() and marks() getters (can remove some methods)
- [ ] update unit test for time governor (in particular test: constructors, estimate_dt, next_time)
- [ ] apply changed interface through the code

PR #291, opened 2015-02-15
Jb time frames

jbrezmorf:
Fix issues #270 and #146


jbrezmorf:
Further cleanup in time_governor is necessary. Part of issue #287.

PR #303, opened 2015-02-26
Df multifield

jbrezmorf:


src/input/json_to_storage.cc
```diff
@@ -336,21 +336,37 @@ StorageBase * JSONToStorage::make_storage(JSONPath &p, const Type::Record *recor
...
+    	std::set<string> keys_to_process;
         json_spirit::mObject::const_iterator map_it;
         std::set<string>::iterator set_it;
 
         for( map_it = j_map.begin(); map_it != j_map.end(); ++map_it) {
-           keys_to_processed.insert(map_it->first);
+           keys_to_process.insert(map_it->first);
+        }
+
+        if (keys_to_process.find("TYPE") != keys_to_process.end()) {
```
jbrezmorf: use record->has_key_iterator(..) to check that record have TYPE key. Only in that case try a conversion.

src/input/json_to_storage.cc
```diff
@@ -336,21 +336,37 @@ StorageBase * JSONToStorage::make_storage(JSONPath &p, const Type::Record *recor
...
+            string value;
+            string ref_address;
+            p.down( record->key_iterator("TYPE")->key_ );
```
jbrezmorf: Rather use same approach as in AbstractRecord. Use try block and catch exception for unknown selection key.
Need not to deal with ref_path etc.

src/input/storage.hh
```diff
@@ -80,16 +80,17 @@ public:
...
     virtual bool is_null() const;
-    virtual StorageBase *deep_copy();
+    virtual StorageBase *deep_copy() const;
     virtual void print(std::ostream &stream, int pad=0) const;
     virtual ~StorageArray();
 private:
     /// Forbids default constructor to have array set to NULL.
     StorageArray();
-    std::vector<StorageBase *> array_;
+    mutable std::vector<StorageBase *> array_;
```
jbrezmorf: Is mutable necessary? If so, comment why.

src/input/json_to_storage.cc
```diff
@@ -336,21 +336,39 @@ StorageBase * JSONToStorage::make_storage(JSONPath &p, const Type::Record *recor
...
+           keys_to_process.insert(map_it->first);
+        }
+
+        Type::Record::KeyIter key_it;
+        if ( record->has_key_iterator("TYPE", key_it) && record->auto_conversion_key_iter() != record->end() ) {
+            JSONPath type_path(p);
+            if (type_path.down( "TYPE" ) != NULL) {
+                try {
+                	if ( type_path.head()->get_str() != record->type_name() ) {
+                		xprintf(UsrErr, "Invalid value of TYPE key of record %s.", record->type_name().c_str());
```
jbrezmorf: 1) Don't use xprintf for error messages, use exceptions.
2) The condition is not necessary, I can imagine conversion to subrecord with
explicitly given TYPE:
	lower = IT::Record("lower","")
		.has_obligatory_type_key()
		.finish()
	IT::Record("Upper","")
		.has_obligatory_type_key()
		.declare_key("rec",lower, "")
		.finish()

	input="{TYPE=\"Lower\"}"

3) This automatic conversion is very tricky, we have to check it by unit tests. I have pushed one proposal which actually fails (not clear how to fix it, we can not provide default value to TYPE in has_obligatory_type_key,
possibly exclude TYPE key from check in allow_auto_conversion

jbrezmorf: Resolution:
* set default value of TYPE key to the Record name v has_obligatory_type_key
* Record::finish - check that autoconvertible key is of type Record, in such a case check that TYPE key is declared. Optionally introduce flag: "is_autoconvertible_from_record" to simplify similar test in JSONToStorage.

	type_id(*data_->keys[data_->autoconvertible_key_idx_].type_) == Record

PR #322, opened 2015-03-10
Pe sorption test

Paulie14:
Resolve limited solubility-precipitation for sorption "none".

Add test20 - two simple test:
- test exact values for all isotherm (no flow,transport..)
- test sorption "none" and precipitation (which was not tested before)

Improve tutorial problem and its documentation.


src/reaction/sorption.cc
```diff
@@ -42,8 +42,20 @@ void SorptionSimple::isotherm_reinit(std::vector<Isotherm> &isotherms_vec, const
 		Isotherm & isotherm = isotherms_vec[i_subst];
 
 		//scales are different for the case of sorption in mobile and immobile pores
-		double scale_aqua = por_m, 
-               scale_sorbed = (1 - por_m) * rock_density * molar_masses_[i_subst];
+		double scale_aqua = por_m,
+               scale_sorbed = (1 - por_m) * rock_density * substances_[substance_global_idx_[i_subst]].molar_mass();
+
+        if(adsorption_type[i_subst] == Isotherm::none)       // means no sorption
```
jbrezmorf: I see no reason for explicit condition since Isotherm::none is passed to the Isotherm::reinit anyway.
This further enlarge code duplicity in SorptionMobiel and SorptionImmobile.

PR #329, opened 2015-03-16
JHy profiler update

x3mSpeedy:
**JSON output** and **MPI independence**


CMake/Modules/Flow123d_config_header.cmake
```diff
@@ -33,14 +33,18 @@ MACRO (FLOW_DEFINE_CONSTANT variable_name numeric_value)
     MESSAGE (STATUS "Adding: '${variable_name}' ${numeric_value}")
     LIST (APPEND DEFINITIONS_LIST "FLOW123D_${variable_name}=${numeric_value}")
     SET (DEFINITIONS_CONTENT "${DEFINITIONS_CONTENT}#define FLOW123D_${variable_name} ${numeric_value}\n")
-ENDMACRO (FLOW_DEFINE variable_name)
+ENDMACRO (FLOW_DEFINE_CONSTANT variable_name numeric_value)
```
jbrezmorf: Next time, put this into separate commit or pull request as it is an unrelated fix.

CMakeLists.txt
```diff
@@ -183,6 +183,43 @@ else ()
...
+    # if success save theese values via flow_define and flow_define_constant macros
+    if (HAVE_TIMER_QUERY_PERFORMANCE_COUNTER AND TIMER_RESOLUTION)
+        flow_define           (HAVE_TIMER_QUERY_PERFORMANCE_COUNTER)
+        # flow_define_constant  (TIMER_RESOLUTION "${TIMER_RESOLUTION}")
+    else()
+        MESSAGE (WARNING "Couldn't access QueryPerformanceCounter timer")   
+    endif()
+endif()
+
+if (NOT TIMER_RESOLUTION)
```
jbrezmorf: So on UNIX we always use std high resolution timer? Then why the other timer tests?
It would be helpful to present observer resolution of individual timers in a comment.

src/system/sys_profiler.cc
```diff
@@ -36,11 +36,23 @@
...
+#include <boost/property_tree/ptree.hpp>
+#include <boost/property_tree/json_parser.hpp>
 
 #include "system/file_path.hh"
+#include "mpi.h"
+#include "timer_data.hh"
 
 
-#include "mpi.h"
+using boost::property_tree::ptree;
```
jbrezmorf: Rater use name space renaming:
e.g. namespace boost_ptree =boost::property_tree::ptree;
or use full names if it is used on few places. This is better for readability.

src/system/timer_data.hh
```diff
@@ -0,0 +1,145 @@
...
+#define TIMERDATA_HH_
+
+#include "config.h"
+#include <string>
+#include <time.h>
+
+/**
+ * Include either Windows or chrono lib
+ */
+#ifdef FLOW123D_HAVE_TIMER_QUERY_PERFORMANCE_COUNTER
```
jbrezmorf: Is it necessary to make includes here. 
Rather move them to the source file (faster builds).

src/system/timer_data.hh
```diff
@@ -0,0 +1,145 @@
...
+        /**
+         * Static method for initializing timers (necessary for Windows timer only)
+         */
+        static void init (void);
+
+        /**
+         * Static method which measures current's timer resolution.
+         * Returned value is average shortest time step in microseconds with double precision
+         */
+        static double get_resolution ();
```
jbrezmorf: Rather name it measure_resolution.

src/system/timer_data.hh
```diff
@@ -0,0 +1,145 @@
...
+
+        /**
+         * Static method which generates TimerData object and sets its ticks value to current time
+         */
+        static TimerData get_time (void);
+
+        /**
+         * Static method for initializing timers (necessary for Windows timer only)
+         */
+        static void init (void);
```
jbrezmorf: Rather use static variable to automatize the initizalization:
e.g. make this method private, returning int. Introduce a static const int member
initialized in the source file:
TimerData::_enforce_initizalization_ = TimeData::init();

src/system/timer_data.hh
```diff
@@ -0,0 +1,145 @@
...
+
+        /**
+         * Method returns time in seconds in double precision
+         */
+        double to_time (void) const;
+
+        /**
+         * Debug-only method returning double value from 'toTime' method as string
+         */
+        string to_string (void);
```
jbrezmorf: Rather overload << operator or remove it from public interface.

src/system/timer_data.hh
```diff
@@ -0,0 +1,145 @@
...
+ * for time measurements.
+ * Maximum resolution is 1000 ns (limited by value of CLOCKS_PER_SEC).
+ */
+class TimerData {
+    public:
+
+        /**
+         * Method returns time in seconds in double precision
+         */
+        double to_time (void) const;
```
jbrezmorf: Possibly use type cast overload:
double operator double() const;

src/system/timer_data.hh
```diff
@@ -0,0 +1,145 @@
...
+        /**
+         * Static method which measures current's timer resolution.
+         * Returned value is average shortest time step in microseconds with double precision
+         */
+        static double get_resolution ();
+
+        /**
+         * Overloaded operator for addition. Allows adding TimerData objects using '+' sign
+         */
+        TimerData operator+ (const TimerData &right);
```
jbrezmorf: Does it make sense? If TimerData represents a point in the time, difference of these points means a time period, but addition makes no sense.

src/system/sys_profiler.cc
```diff
@@ -87,8 +99,8 @@ double MPI_Functions::max(double* val, MPI_Comm comm) {
  */
 
 Timer::Timer(const CodePoint &cp, int parent)
-: start_time(0),
-  cumul_time(0),
+: start_time(TimerData::get_time()),
+  cumul_time(start_time - start_time),
```
jbrezmorf: Rather make cumul_time just double, time in seconds. See also notes in TimerData.

src/system/timer_data.hh
```diff
@@ -0,0 +1,145 @@
...
+        /**
+         * Internal variable for storing actual ticks
+         */
+        long long ticks_;
+
+#ifdef FLOW123D_HAVE_TIMER_QUERY_PERFORMANCE_COUNTER
+        /**
+         * Variable for storing CPU Frequency
+         */
+        static LARGE_INTEGER frequency;
```
jbrezmorf: Where LARGE_INTEGER comes from?

src/system/timer_data.hh
```diff
@@ -0,0 +1,145 @@
...
+        /**
+         * Constructor will populate object with current time
+         */
+        TimerData ();
+
+        /**
+         * Method which returns current time ticks. Depending on used timer value may represent
+         * nanoseconds or internal ticks / CPU frequency
+         */
+        long long get_ticks (void) const;
```
jbrezmorf: No reason for private getters/setters.

src/system/timer_data.cc
```diff
@@ -0,0 +1,131 @@
...
+#include <chrono>
+#endif //FLOW123D_HAVE_TIMER_CHRONO_HIGH_RESOLUTION
+
+bool TimerData::inited = false;
+
+/*
+ * Constructor
+ */
+
+TimerData::TimerData () :
```
jbrezmorf: Why not move code from get_current_ticks here?

src/system/timer_data.hh
```diff
@@ -0,0 +1,145 @@
+/*
```
jbrezmorf: After simplifications I suggest we can make two independent implementations of this class and just 
have one conditional compilation in the Timer class containing include of the right implementation.

src/system/sys_profiler.hh
```diff
@@ -63,6 +63,10 @@
 #include "global_defs.h"
 #include "system/system.hh"
 #include <mpi.h>
+#include <boost/property_tree/ptree.hpp>
+#include "time_point.hh"
+
+using boost::property_tree::ptree;
```
jbrezmorf: Use just namespace rename:
e.g. namespace ptree=boost::property_tree::ptree

This improves readability since reader notice that some class or function is related to some library.

src/system/sys_profiler.cc
```diff
@@ -36,11 +36,23 @@
...
+#include "mpi.h"
+#include "time_point.hh"
 
 
-#include "mpi.h"
+using boost::property_tree::ptree;
+using boost::property_tree::read_json;
+using boost::property_tree::write_json;
+
+#define FLOW123D_JSON_PRETTY        1
```
jbrezmorf: Why these defines.  Nobody knows their meaning without documentation. But better to use typed constants in particular functions, where the meaning could be clear from context. So two morals:
1) prefer const ints over defines
2) make global only what have to be global

src/system/sys_profiler.cc
```diff
@@ -133,8 +132,7 @@ void Timer::start() {
 
 
 void Timer::update() {
-    TimeData time = get_time();
-    cumul_time += time - start_time;
+    cumul_time += (TimePoint() - start_time);
```
jbrezmorf: Rather use so simple code diractly since it is better then function named "update" where is not clear what it updates.

src/system/sys_profiler.hh
```diff
@@ -351,11 +391,44 @@ protected:
     size_t total_deallocated_;
 
 
+    /**
+     * common part in filepath
```
jbrezmorf: Better documentation. Not clear.

src/system/sys_profiler.cc
```diff
@@ -176,6 +173,7 @@ string Timer::code_point_str() const {
     return boost::str( boost::format("%s:%d, %s()") % code_point_->file_ % code_point_->line_ % code_point_->func_ );
 }
 
+string Timer::common_path = "";
```
jbrezmorf: If do not understand meaning of this static !!.

src/system/sys_profiler.cc
```diff
@@ -350,51 +348,78 @@ void Profiler::notify_free(const size_t size) {
...
+    // get timer and check preconditions
+    Timer &timer = timers_[timer_idx];
     ASSERT( timer_idx >=0, "Wrong timer index %d.\n", timer_idx);
     ASSERT( timer.parent_timer >=0 , "Inconsistent tree.\n");
 
-    int numproc;
-    MPI_Comm_size(comm, &numproc);
+    // fix path
+    string filepath = timer.code_point_->file_;
+    filepath.erase (0, Timer::common_path.size());
```
jbrezmorf: If common_path is used only in this method, then move it here.

src/system/sys_profiler.cc
```diff
@@ -350,51 +348,78 @@ void Profiler::notify_free(const size_t size) {
...
+        result += t2 - t1;
+    }
 
-void Profiler::add_timer_info(MPI_Comm comm, vector<vector<string> > &timers_info, int timer_idx, int indent, double parent_time) {
+    return (result / measurements) * 1000; // ticks to seconds to microseconds conversion
+}
+
+
+
+std::string common_prefix( std::string a, std::string b ) {
```
jbrezmorf: Where this is used?

src/system/sys_profiler.hh
```diff
@@ -512,6 +608,13 @@ private:
      */
     void update_running_timers();
 
+    /**
```
jbrezmorf: This should be protected or private method.

jbrezmorf:
Completely done and ready to merge as soon as Python formater will be ready in the subsequent development.

PR #340, opened 2015-03-26
Jb apply frames

jbrezmorf:


jbrezmorf:
Apply TimeStep through the code, in particular in fields.
Create double porosity test that somehow reproduce problems with jump times.

Issue #23, opened 2013-10-27
reuse linear algebra classes


jbrezmorf:
Reuse linear algebra wrapper classes from some other project either deal.ii or LibMesh.
Deal.ii seems to be better, supporting block vectors and having its own sequential implementation.

- [x] make our fork
- [x] make branch with our changes
- [x] make submodule in Flow123d repository for our fork
- [x] make files compile 

... then pass through every file, enhance documentation and possibly modify to our purpose. This would be another issue.

jbrezmorf:
merged in a7328af

PR #350, opened 2015-04-09
Js deal ii

jstebel:
For review (#23). See also submodule pull-request at https://github.com/jstebel/dealii/compare/v8.2.1...jstebel:flow123d_fork_v8.2.1

Issues:
- How to properly update the deal.II submodule and track a given branch (git submodule foreach -q --recursive 'branch="$(git config -f <path>.gitmodules submodule.$name.branch)"; git checkout $branch')?
- Deal.II still depends on 2 bundled packages: muparser and tbb. Should they be shipped with Flow123d instead?
- Use whole base and lac directories or just selected files?
- Clean unused directories and files?


jbrezmorf:
- the official deal.ii fork for flow123d should be under flow123d organisation
- do not support muparser
- have TBB as part of flow123d in third party until we have support for dynamic libraries (make a not in appropriate issue)

jstebel:
We want to keep master branch synchronized to official deal.II repository. The actual branch used in Flow123d is renamed to flow123d_fork.

Issue #201, opened 2014-08-11
PythonLoader exceptions


jbrezmorf:
PythonLoader class should extract error messages from Python and throw exeption holding this error string. Then we can catch them in calling functions reporting also context of the error.

Problem may be that there is no simple way how to extract Python error message withou printing it to std:err.

jbrezmorf:
merged in 43e55cbc1b8c

PR #351, opened 2015-04-14
Df python exception

dflanderka:


src/fields/field_python.impl.hh
```diff
@@ -109,18 +109,12 @@ void FieldPython<spacedim, Value>::set_func(const string &func_name)
...
 
-    }
     if (! PyCallable_Check(p_func_)) {
-        xprintf(UsrErr, "Field '%s' from the python module: %s is not callable.\n", func_name.c_str(), PyModule_GetName(p_module_) );
-        Py_XDECREF(p_func_);
-        Py_XDECREF(p_module_);
-
+    	stringstream ss;
+    	ss << "Field '" << func_name << "' from the python module: " << PyModule_GetName(p_module_) << " is not callable." << endl;
+    	THROW(PythonLoader::ExcPythonError() << PythonLoader::EI_PythonMessage( ss.str() ));
```
jbrezmorf: This error is not an exception raised in Python so it should not be reported through the PythonLoader exception. 
Move these checks into:
PyObject *PythonLoader::get_callable(PyObject *module, string name);
and raise dedicated exceptions defined in PythonLoader.

src/fields/field_python.impl.hh
```diff
@@ -131,15 +125,12 @@ void FieldPython<spacedim, Value>::set_func(const string &func_name)
...
-    if (p_value_ == NULL) {
-        PyErr_Print();
-        // TODO: use cout to output also field arguments
-        xprintf(Err,"Failed to call field '%s' from the python module: %s\n", func_name.c_str(), PyModule_GetName(p_module_) );
-    }
+    PythonLoader::check_error();
 
     if ( ! PyTuple_Check( p_value_)) {
-        xprintf(UsrErr, "Field '%s' from the python module: %s doesn't return Tuple.\n", func_name.c_str(), PyModule_GetName(p_module_) );
+    	stringstream ss;
```
jbrezmorf: This error is not an exception raised in Python so it should not be reported through the PythonLoader exception. 
Just use ExcMessage.

src/system/python_loader.hh
```diff
@@ -43,6 +43,14 @@ public:
 class PythonLoader {
 public:
 
+	/**
+	 * Definition of exception thrown by python compiler.
```
jbrezmorf: Doc: ... by python compiler or interpreter.

jbrezmorf:
Nearly done. It seems practical to move part of FieldPython implementation to the general purpose method of PythonLoader. It could be later reused in Profiler etc.

jbrezmorf:
Close issue  #201.

PR #352, opened 2015-04-14
Df apply factory

dflanderka:


src/coupling/hc_explicit_sequential.cc
```diff
@@ -73,7 +73,15 @@ it::Record HC_ExplicitSequential::input_type
  */
 HC_ExplicitSequential::HC_ExplicitSequential(Input::Record in_record)
 {
-    START_TIMER("HC constructor");
+	FLOW123D_FORCE_LINK_IN_PARENT(transportOperatorSplitting);
```
jbrezmorf: Rather move this out of any function say just after includes.

src/flow/darcy_flow_mh.cc
```diff
@@ -1794,5 +1808,10 @@ void DarcyFlowLMH_Unsteady::postprocess() {
   VecAssemblyEnd(schur0->get_solution());
 }
 
+
+FLOW123D_FORCE_LINK_IN_CHILD(steady_MH);
```
jbrezmorf: Move these to beginning of the file.

src/fields/field_set.hh
```diff
@@ -216,6 +216,7 @@ public:
      * Collective interface to @p FieldCommonBase::output().
      */
     void output(OutputTime *stream);
```
jbrezmorf: Can we remove this output method, same for field and multi_field.

src/system/global_defs.h
```diff
@@ -210,6 +210,15 @@ static const int debug_asserts_view = 0;
 
 #endif
 
+
+/**
+ * Usage in Input::Factory class.
+ * Necessary for linking static input_type records.
```
jbrezmorf: Better document why we need these macros.

PR #364, opened 2015-05-11
Df type repository

jbrezmorf:


src/fields/field.impl.hh
```diff
@@ -99,16 +100,11 @@ Field<spacedim,Value> &Field<spacedim,Value>::operator=(const Field<spacedim,Val
...
-	 * List of AbstratRecord types created by make_input_tree() in get_input_type() implementation.
-	 * We have to return reference, which may be reference to not yet initialized static object.
-	 *
-	 * TODO: Have method to get persistent copy of an Input Type (which exists nevertheless)
-	 */
 	static vector<it::AbstractRecord> ar_list;
 
 	if (is_enum_valued) {
 		ar_list.push_back(make_input_tree());
+		boost::shared_ptr<it::AbstractRecord> ptr = Input::TypeRepository<it::AbstractRecord>::getInstance().add_type( ar_list.back() );
```
jbrezmorf: ??

src/coupling/hc_explicit_sequential.cc
```diff
@@ -53,26 +53,36 @@ FLOW123D_FORCE_LINK_IN_PARENT(unsteady_LMH)
 
 namespace it = Input::Type;
 
-it::AbstractRecord CouplingBase::input_type
-    = it::AbstractRecord("Problem",
+it::AbstractRecord & CouplingBase::get_input_type() {
+	static it::AbstractRecord type = it::AbstractRecord("Problem",
```
jbrezmorf: Remove static (here and in all generators).

src/fields/field.impl.hh
```diff
@@ -13,6 +13,7 @@
 #include "field.hh"
 #include "mesh/region.hh"
 #include "input/json_to_storage.hh"
+#include "input/type_repository.hh"
```
jbrezmorf: Should not be necessary. Try remove.

src/fields/field_algo_base.impl.hh
```diff
@@ -53,26 +53,31 @@ string FieldAlgorithmBase<spacedim, Value>::template_name() {
...
+		const it::Selection * sel_type = dynamic_cast<const it::Selection *>(element_input_type);
+		if (sel_type != NULL ) element_input_sel = sel_type;
+	}
+
+	it::AbstractRecord type= it::AbstractRecord("Field:"+template_name(), "Abstract record for all time-space functions.")
+    	.allow_auto_conversion("FieldConstant")
+		.set_element_input(element_input_sel)
+		.close();
+
+	if ( !type.is_finished() ) {
```
jbrezmorf: Why this? Should it be !type.is_closed()?

src/fields/field_set.hh
```diff
@@ -111,8 +111,10 @@ public:
 	 *
 	 * Input::Type::Record SomEquation::input_type=
 	 * 		Record("SomeEquation","equation's description")
-	 * 		.declare_key("data",Input::Type::Array(EqData().make_field_descriptor_type()),"List of field descriptors.");
+	 * 		.declare_key("data",Input::Type::Array(EqData().make_field_descriptor_type().close()),"List of field descriptors.");
```
jbrezmorf: Why this close? make_field_descriptor should return closed type

src/flow/darcy_flow_mh.cc
```diff
@@ -104,64 +106,100 @@ it::Selection DarcyFlowMH::EqData::bc_type_selection =
...
+				"Linear solver for MH problem.")
+		.declare_key("output", DarcyFlowMHOutput::get_input_type(), it::Default::obligatory(),
+				"Parameters of output form MH module.")
+		.declare_key("mortar_method", DarcyFlowMH::get_mh_mortar_selection(), it::Default("None"),
+				"Method for coupling Darcy flow between dimensions." )
+		.declare_key("balance", Balance::get_input_type(), it::Default::obligatory(),
+				"Settings for computing mass balance.")
+		.declare_key("time",         TimeGovernor::get_input_type(), it::Default::obligatory(),
+									"Time governor setting for the unsteady Darcy flow model.")
+		.copy_keys(DarcyFlowMH_Steady::get_input_type())
```
jbrezmorf: Move copy keys at the beginning.

src/flow/darcy_flow_mh.cc
```diff
@@ -184,7 +222,7 @@ DarcyFlowMH::EqData::EqData()
     	water_source_density.units( UnitSI().s(-1) );
     
     ADD_FIELD(bc_type,"Boundary condition type, possible values:", "\"none\"" );
-        bc_type.input_selection(&bc_type_selection);
+        bc_type.input_selection( &get_bc_type_selection() );
```
jbrezmorf: Mark this as UGLY, we should try to get rid of pointer to the selection after having generic types.

src/input/type_record.cc
```diff
@@ -442,8 +445,6 @@ Record &Record::declare_key(const string &key, const KeyType &type,
         data_->declare_key(key, boost::shared_ptr<const TypeBase>(), &type, default_value, description);
     } else {
         // for Array, Double, Integer, we assume no static variables
-    	if (const_cast<KeyType *>( &type )->finish()) check_key_default_value(default_value, type, key);
```
jbrezmorf: Why removing this? We have to check default value somewhere.

src/input/type_record.cc
```diff
@@ -526,19 +527,21 @@ AbstractRecord::AbstractRecord(const string & type_name_in, const string & descr
...
+TypeBase::TypeHash AbstractRecord::content_hash() const
 {
-    std::size_t seed=0;
+	TypeHash seed=0;
     boost::hash_combine(seed, "Abstract");
-    boost::hash_combine(seed, data_.get());
-    boost::hash_combine(seed, full_type_name());
+    boost::hash_combine(seed, type_name());
     boost::hash_combine(seed, data_->description_);
+    if (child_data_->element_input_selection != nullptr) {
```
jbrezmorf: Mark this as HACK, that should be removed after generic types.

src/input/type_record.cc
```diff
@@ -634,6 +637,44 @@ unsigned int AbstractRecord::child_size() const {
...
+	data_->finished = true;
+	no_more_descendants();
+
+	return (data_->finished);
+}
+
+
+AbstractRecord &AbstractRecord::close() {
+	data_->closed_=true;
+    return *( Input::TypeRepository<AbstractRecord>::getInstance().add_type( *this ).get() );
```
jbrezmorf: Why you are still using explicit call of get()?

return *( Input::TypeRepository<AbstractRecord>::getInstance().add_type( *this ));
.. should work as well.

src/input/type_record.hh
```diff
@@ -509,6 +514,8 @@ protected:
          * Vector of derived Records (proxies) in order of derivation.
          */
         vector< Record > list_of_childs;
+
+        const Selection * element_input_selection;
```
jbrezmorf: Mark as HACK

src/input/type_record.hh
```diff
@@ -613,6 +623,41 @@ public:
...
+
+		 /// implementation of registration variable
+		 const int SomeDescendant::reg =
+			 Input::register_class< SomeDescendant >("SomeDescendant") +
+			 SomeBase::get_input_type().add_child(SomeDescendant::get_input_type());
+	 @endcode
+     */
+    int add_child(const Record &subrec);
+
+    AbstractRecord &set_element_input(const Selection * element_input);
```
jbrezmorf: Mark as HACK

src/input/type_repository.hh
```diff
@@ -0,0 +1,82 @@
...
+class TypeRepository {
+public:
+	/// Template parameter can be only descendant of TypeBase class.
+	static_assert(std::is_base_of<Type::TypeBase, T>::value,
+	        "T must be a descendant of Input::Type::TypeBase"
+	    );
+
+    typedef std::map< Type::TypeBase::TypeHash, boost::shared_ptr<T> > TypeRepositoryMap;
+
+    static TypeRepository & getInstance() {
```
jbrezmorf: rename to get_instance

jbrezmorf:
Very good job, just few questions.

src/fields/field_algo_base.impl.hh
dflanderka: It's necessary. Only 'is_closed' test adds childs repeated for each call and throws exception.
This solution is temporary.

src/fields/field_set.hh
dflanderka: In some cases (transport, unit tests) method declare_key is called on returned record. Declare_key can't be called on closed record.

src/input/type_record.cc
dflanderka: Method check_key_default_value is now called only in finish. In some cases call ( &type )->finish() causes finishing of some incomplete descendants of 'type'.

jbrezmorf: Checking default value here produce much better error message, since one can find exact reason for the error from the stack trace. Is it necessary to finish type first? Isn't close enough? Are there some principal reasons why the check can not be done here?

src/fields/field_algo_base.impl.hh
jbrezmorf: OK.

src/fields/field_set.hh
jbrezmorf: I.e. make_filed_descriptor returns unclosed record? That is potential source of errors, since close do the TypeRepository stuff. I propose to close the record and use copy constructor or copy_keys or something new to perform adding further keys ... but in completely new record.

PR #380, opened 2015-06-17
Js dg with linear sorption

jstebel:
Implemented linear sorption.
- Mass matrix is modified by adding a retardation coefficient computed from Model::compute_retardation_coefficient()
- each substance has its own TransportDG::mass_matrix
- cumulative sources adjusted to mass exchange between rock and solvent.
- sorption input fields added to SoluteTransport_DG_Data record, which seems easier than reusing class SorptionBase (too complicated initialization for our purpose)
- updated documentation of Transport model
- initial condition for sorbed substance is assumed to be in equilibrium with the substance in liquid

TODO:
- add integration test for sorption with diffusion?

### DIFF
--- a/src/transport/concentration_model.cc
+++ b/src/transport/concentration_model.cc
@@ -46,6 +46,8 @@ using namespace Input::Type;
 
 
 
+
+
 ConcentrationTransportModel::ModelEqData::ModelEqData()
 : TransportBase::TransportEqData()
 {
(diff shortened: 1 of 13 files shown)


src/transport/concentration_model.cc
```diff
@@ -78,6 +80,24 @@ ConcentrationTransportModel::ModelEqData::ModelEqData()
             .units( UnitSI().m(2).s(-1) )
             .input_default("0.0")
             .flags_add( in_main_matrix & in_rhs );
+    *this+=solvent_density
```
jbrezmorf: Solvent density should be just constant not a field.

PR #382, opened 2015-06-20
Df input clean

dflanderka:


src/input/type_base.cc
```diff
@@ -167,36 +109,7 @@ bool Array::ArrayData::finish()
...
-			boost::shared_ptr<const TypeBase> type_copy = boost::make_shared<const Selection>(*s);
-			type_of_values_ = type_copy;
-			p_type_of_values = 0;
-		}
-		else if (dynamic_cast<const Array *>(p_type_of_values) != 0)
-		    xprintf(PrgErr, "Should not happen!\n");
-	}
-
-	return (finished = true);
+	return (finished = const_cast<TypeBase *>( type_of_values_.get() )->finish() );
```
jbrezmorf: const_cast smells....
Possibly we can make finish method constant and make finished flag mutable.

src/input/type_base.cc
```diff
@@ -238,21 +151,10 @@ Array::Array(const ValueType &type, unsigned int min_size, unsigned int max_size
...
-         && ! TypeBase::was_constructed(&type) ) {
-        data_->p_type_of_values = &type;
-        TypeBase::lazy_type_list().push_back( boost::make_shared<Array>( *this ) );
-    } else {
-        data_->p_type_of_values = NULL;
-        boost::shared_ptr<const TypeBase> type_copy = boost::make_shared<ValueType>(type);
-        data_->type_of_values_ = type_copy;
-        data_->finished=true;
-    }
+	boost::shared_ptr<const TypeBase> type_copy = boost::make_shared<ValueType>(type);
```
jbrezmorf: Here the copy may be non-const, is n't it?

src/input/type_base.hh
```diff
@@ -73,6 +73,13 @@ public:
     virtual bool is_finished() const
     {return true;}
 
+    /**
+     * Returns true if the type is closed.
+     *
+     */
+    virtual bool is_closed() const
```
jbrezmorf: Is this method correctly overloaded in descendants?

src/input/type_record.cc
```diff
@@ -124,9 +120,21 @@ void Record::make_derive_from(AbstractRecord &parent) {
     if (data_->derived_) return;
```
jbrezmorf: Make a note , that this could by simplified if we manage to have TYPE key in every record. Currently, it is not possible due to new autoconversion rules for record transpose.

src/input/type_record.cc
```diff
@@ -184,26 +191,11 @@ void Record::make_copy_keys(Record &origin) {
...
-		Record tmp(*ptr);
-		make_copy_keys(tmp);
-		ptr = NULL;
-	}
-	data_->copy_from_ptr.clear();
-}
-
-
 
 Record &Record::derive_from(AbstractRecord &parent) {
```
jbrezmorf: - if no key defined -> define key TYPE with empty shared ptr to actula type
- ASSERT that only TYPE key is defined (all other keys must com later)
- in finish check for TYP key and set proper selection from parent

src/input/type_record.cc
```diff
@@ -554,7 +508,7 @@ TypeBase::TypeHash AbstractRecord::content_hash() const
 
 void AbstractRecord::add_descendant(const Record &subrec)
```
jbrezmorf: merge with add_child

src/input/type_record.cc
```diff
@@ -574,13 +528,12 @@ void AbstractRecord::no_more_descendants()
 {
```
jbrezmorf: Make private, or rather put into finish. Fix unit_tests.

src/input/type_record.cc
```diff
@@ -574,13 +528,12 @@ void AbstractRecord::no_more_descendants()
...
     child_data_->selection_of_childs->close();
     // check validity of possible default value of TYPE key
-    if (data_->keys.size() > 0 ) { // skip for empty records
-        Default &dflt = data_->keys[0].default_;
-        if ( dflt.has_value_at_declaration() ) {
+    if ( !child_data_->selection_default_.is_obligatory() ) { // skip for empty records
+        if ( child_data_->selection_default_.has_value_at_declaration() ) {
```
jbrezmorf: Rather ASSERT for has_value_at_declaration.

src/input/type_record.cc
```diff
@@ -574,13 +528,12 @@ void AbstractRecord::no_more_descendants()
 {
     child_data_->selection_of_childs->close();
     // check validity of possible default value of TYPE key
-    if (data_->keys.size() > 0 ) { // skip for empty records
-        Default &dflt = data_->keys[0].default_;
-        if ( dflt.has_value_at_declaration() ) {
+    if ( !child_data_->selection_default_.is_obligatory() ) { // skip for empty records
```
jbrezmorf: Better comment.

src/input/type_record.cc
```diff
@@ -590,10 +543,9 @@ void AbstractRecord::no_more_descendants()
 
 bool AbstractRecord::valid_default(const string &str) const
 {
-    if (data_->keys.size() != 0)  { // skip for empty records
-        Default &dflt = data_->keys[0].default_;
+    if ( !child_data_->selection_default_.is_obligatory() )  { // skip for empty records
         if (! child_data_->selection_of_childs->is_finished()) return false;
```
jbrezmorf: Rather ASSERT.

src/input/type_record.cc
```diff
@@ -590,10 +543,9 @@ void AbstractRecord::no_more_descendants()
 
 bool AbstractRecord::valid_default(const string &str) const
```
jbrezmorf: Improve documentation in header file.

src/input/type_record.cc
```diff
@@ -617,10 +569,9 @@ const Record  & AbstractRecord::get_descendant(unsigned int idx) const
 
 
 const Record * AbstractRecord::get_default_descendant() const {
-    if (data_->keys.size() != 0 )  { // skip for empty records
-        Default &dflt = data_->keys[0].default_;
-        if ( dflt.has_value_at_declaration() ) {
-            return &( get_descendant( dflt.value() ) );
+    if ( !child_data_->selection_default_.is_obligatory() )  { // skip for empty records
```
jbrezmorf: Possibly introduce private method have_default_descendant in order to reduce code duplication.

src/input/type_record.cc
```diff
@@ -641,9 +592,6 @@ unsigned int AbstractRecord::child_size() const {
 
 int AbstractRecord::add_child(const Record &subrec)
```
jbrezmorf: Put add_descendant here.

jbrezmorf:
Make Abstract derived from TypeBase directly.

src/input/type_record.cc
```diff
@@ -184,25 +169,25 @@ void Record::make_copy_keys(Record &origin) {
...
-    } else { //postponed
-        data_->p_parent_ = &parent;
+	//ASSERT( ! data_->parent_ptr_.size() , "Record has been already derived.\n");
+	ASSERT( parent.is_closed(), "Parent AbstractRecord '%s' must be closed!\n", parent.type_name().c_str());
+	ASSERT( data_->keys.size() == 0 || (data_->keys.size() == 1 && data_->keys[0].key_ == "TYPE"),
+			"Derived record '%s' can have defined only TYPE key!\n", this->type_name().c_str() );
+
+	bool contains = false;
+	for (auto it = data_->parent_ptr_.begin(); it != data_->parent_ptr_.end(); ++it) {
+	    if ((*it)->type_name() == parent.type_name()) {
```
jbrezmorf: I suggest to rather keep all pointers and make this list unique during finish. 
Then during finish we have to use hash of the parent abstract, otherwise we are not able to distinguish abstracts comming from common generic type.

jbrezmorf:
Just one minor issue.

src/input/type_record.cc
dflanderka: Done.

PR #387, opened 2015-06-25
Pe removemhfe

Paulie14:
Suggesting partial merge of PE_mhfevalues.

Get rid of MH_FEValues class.
- add FE and mapping objects for velocity into DarcyFlowMH
- replace MH_FEValues computation with FEValues
- assemble local_mh_matrix templated by dimension
- replace MH_FEValues in DarcyFlowOutput

DoFHandler and FEValues:
- add function get_loc_dof_indices() to get local indices
- add some getter for FEValues


src/fem/dofhandler.cc
```diff
@@ -363,6 +363,26 @@ void DOFHandlerMultiDim::get_dof_indices(const CellIterator &cell, unsigned int
...
+        break;
+    case 2:
+        for (unsigned int k=0; k<fe2d_->n_object_dofs(dim,DOF_SINGLE); k++)
+            indices[k] = object_dofs[cell.index()][dim][k] - loffset_;
+        break;
+    case 3:
+        for (unsigned int k=0; k<fe3d_->n_object_dofs(dim,DOF_SINGLE); k++)
+            indices[k] = object_dofs[cell.index()][dim][k] - loffset_;
+        break;
+    }
```
jstebel: Hint to simplify code: In the switch, set a variable n_object_dofs, and then put the for-cycle out of the switch.

src/fem/fe_values.cc
```diff
@@ -216,7 +216,17 @@ const Quadrature<dim> * FEValuesBase<dim,spacedim>::get_quadrature() const
...
+const FiniteElement<dim,spacedim> * FEValuesBase<dim,spacedim>::get_fe() const
+{
+    return fe;
+}
 
+template<unsigned int dim, unsigned int spacedim> inline
+const Mapping<dim,spacedim> * FEValuesBase<dim,spacedim>::get_mapping() const
+{
+    return mapping;
+}
```
jstebel: "inline" makes sense only in header file.

jbrezmorf: So we should move implementation of critical getters into the header file.

src/flow/darcy_flow_mh_output.cc
```diff
@@ -243,29 +245,58 @@ void DarcyFlowMHOutput::make_element_scalar() {
...
+                              * fv_rt2.shape_vector(li,0); //fe_values.RT0_value( ele, ele->centre(), li )
+                }
+                break;
+            case 3: 
+                fv_rt3.reinit(ele);
+                for (unsigned int li = 0; li < ele->n_sides(); li++) {
+                    flux_in_center += mhdh.side_flux( *(ele->side( li ) ) )
+                              * fv_rt3.shape_vector(li,0); //fe_values.RT0_value( ele, ele->centre(), li )
+                }
+                break;
```
jbrezmorf: This and other code involving different dimensions is terribly redundant. 
I suggest create an internal class template DimensionData<dim>, move all dimension dependent data members there
and move parts of assembly routine there.

PR #388, opened 2015-06-25
Pe balance

Paulie14:
- correct bug - add condition on balance presence
- add tests for balance switched off


jbrezmorf:
Did you check also other equations for similar errors?

Paulie14:
I went through the code in darcy flow - steady, unsteady MH/LMH, and transport - TOS, convectional, DG and heat flow. I have found no other bugs regarding balance pointer. All these equations are supposed to be tested from now on (except heatflow,because the balance is computed on the DG level)

PR #389, opened 2015-06-30
Df attributes

dflanderka:


src/input/type_base.cc
```diff
@@ -28,6 +28,7 @@
 #include "type_record.hh"
 #include "type_output.hh"
 #include "type_repository.hh"
+#include "json_to_storage.hh"
```
jbrezmorf: Rather just include of JSON parser, i.e. "json_spirit/json_spirit.h".

src/input/type_base.hh
```diff
@@ -154,6 +156,12 @@ public:
      */
     virtual TypeHash content_hash() const =0;
 
+    /// Add attribute to map
+    void add_attribute(std::string name, json_string val);
+
+    /// Print JSON output of attributes to @p stream.
+    void print_json(ostream& stream);
```
jbrezmorf: Rename to "write_attributes".

src/input/type_output.cc
```diff
@@ -1293,7 +1293,8 @@ void OutputJSONMachine::print_impl(ostream& stream, const Record *type, unsigned
     stream << "\"input_type\" : \"Record\"," << endl;
     stream << "\"type_name\" : \"" << type->type_name() << "\"," << endl;
     stream << "\"type_full_name\" : \"" << type->full_type_name() << "\"," << endl;
-    stream << endl;
+    const_cast<Record *>(type)->print_json(stream);
```
jbrezmorf: Avoid this const_cast by making print_json const method.

src/input/type_base.cc
```diff
@@ -44,11 +45,13 @@ using namespace std;
 
 
 
-TypeBase::TypeBase() {}
+TypeBase::TypeBase()
+: attributes_( boost::make_shared<attribute_map>() ) {}
```
jbrezmorf: Not necessary, this is automatic initiation of attributes_.

src/input/type_record.cc
```diff
@@ -475,6 +476,7 @@ RECORD_DECLARE_KEY(AdHocAbstractRecord);
 AbstractRecord::AbstractRecord()
 : child_data_( boost::make_shared<ChildData>( "EmptyAbstractRecord", "" ) )
 {
+	attributes_ = boost::make_shared<attribute_map>();
```
jbrezmorf: Not necessary. delete.

src/input/type_record.cc
```diff
@@ -483,13 +485,17 @@ AbstractRecord::AbstractRecord()
 
 AbstractRecord::AbstractRecord(const AbstractRecord& other)
 : child_data_(other.child_data_)
-{}
+{
+	attributes_ = other.attributes_;
```
jbrezmorf: Rather call TypeBase copy constructor in initializer list. Need also add to other copy constructors, are there any for Record or Selection?

jbrezmorf:
Proposed few simplifications. Possible missing call of TypeBase copy constructor for derived classes has to be added.

src/input/type_base.cc
dflanderka: It can't be removed, under cygwin throws error.

jbrezmorf: Of course. I overlooked the shared pointer.

PR #392, opened 2015-07-10
Jb generic types

jbrezmorf:


src/input/type_base.cc
```diff
@@ -46,12 +47,12 @@ using namespace std;
 
 
 TypeBase::TypeBase()
-: attributes_( boost::make_shared<attribute_map>() ) {}
+: attributes_( boost::make_shared<attribute_map>() ), generic_(false) {}
```
jbrezmorf: This flag may not be necessary. See not on is_generic

src/input/type_base.cc
```diff
@@ -159,6 +169,9 @@ bool Array::ArrayData::finish()
 {
 	if (finished) return true;
 
+    ASSERT(typeid( *type_of_values_ ) != typeid(Parameter), "Finished Array can't be of type Parameter '%s'.\n",
```
jbrezmorf: possibly, we have to support finish of generic types.

src/input/type_base.cc
```diff
@@ -190,6 +203,39 @@ bool Array::valid_default(const string &str) const {
...
+		json_string val = ss.str();
+		if (this->validate_json(val)) {
+			(*arr.attributes_)["parameters"] = val;
+		} else {
+			xprintf(PrgErr, "Invalid JSON format of attribute 'parameters'.\n");
+		}
+		return boost::make_shared<Array>(arr);
+	}
+
+	return boost::make_shared<Array>(*this);
```
jbrezmorf: Need to call make_instance recursively for the type_of_values_. Adjust attributes of the array according to attributes of returned  type.

src/input/type_base.cc
```diff
@@ -190,6 +203,39 @@ bool Array::valid_default(const string &str) const {
 }
 
 
+boost::shared_ptr<TypeBase> Array::make_instance(std::vector<ParameterPair> vec) const {
+	// Replace only if type_of_values_ is parameter
+	if ( typeid( *(data_->type_of_values_) ) == typeid(Parameter) ) {
+		ASSERT( vec.size() == 1, "Parameter vector of Input::Type::Array '%s' must have size 1!", this->type_name().c_str());
+		std::vector<ParameterPair>::iterator vec_it=vec.begin();
+		// Create copy of array, we can't set type from parameter vector directly (it's TypeBase that is not allowed)
+		Array arr = Array(Integer(), this->data_->lower_bound_, this->data_->upper_bound_);
```
jbrezmorf: We should have proper copy constructor for making copies.

src/input/type_base.cc
```diff
@@ -190,6 +203,39 @@ bool Array::valid_default(const string &str) const {
...
+		// Set parameters as attribute
+		std::stringstream ss;
+		ss << "[";
+		for (std::vector<ParameterPair>::iterator vec_it=vec.begin(); vec_it!=vec.end(); vec_it++) {
+			if (vec_it != vec.begin()) ss << "," << endl;
+			ss << "{ \"" << (*vec_it).first << "\" : \"" << (*vec_it).second->content_hash() << "\" }";
+		}
+		ss << "]";
+		json_string val = ss.str();
+		if (this->validate_json(val)) {
```
jbrezmorf: make this just an ASSERT.

src/input/type_base.cc
```diff
@@ -190,6 +203,39 @@ bool Array::valid_default(const string &str) const {
...
+			(*arr.attributes_)["parameters"] = val;
+		} else {
+			xprintf(PrgErr, "Invalid JSON format of attribute 'parameters'.\n");
+		}
+		return boost::make_shared<Array>(arr);
```
jbrezmorf: Missing attribute with link to original generic type

src/input/type_base.hh
```diff
@@ -162,6 +168,14 @@ public:
     /// Print JSON output of attributes to @p stream.
     void write_attributes(ostream& stream) const;
 
+    /// Create instance of generic type
+    virtual boost::shared_ptr<TypeBase> make_instance(std::vector<ParameterPair> vec = std::vector<ParameterPair>()) const
```
jbrezmorf: TypeBase is abstract class anyway, so just make this method abstract as well:
make_instance(...) = 0;

src/input/type_base.hh
```diff
@@ -162,6 +168,14 @@ public:
...
+    { ASSERT(false, "make_instance() may not be called for TypeBase class.\n"); };
+
+    /// Return if type is generic.
+    inline bool is_generic() const
```
jbrezmorf: It seems that this method is not used. We will see if it is necessary somewhere during finish phase.

src/input/type_base.hh
```diff
@@ -261,6 +278,9 @@ public:
     /**
      * Constructor with a @p type of array items given as pure reference. In this case \p type has to by descendant of \p TypeBase different from
      * 'complex' types @p Record and @p Selection. You can also specify minimum and maximum size of the array.
+     *
+     * Possibly modification: If we need set attributes to Array, we can't need TypeBase::add_attribute method (requires non-closed type).
+     * Solution: We add to constructor optional parameter of type vector< pair<std::string, json_string> > what allows set all attributes.
```
jbrezmorf: Seems that finally we can live without that auxiliary parameter.

src/input/type_generic.cc
```diff
@@ -0,0 +1,92 @@
...
+TypeBase::TypeHash Parameter::content_hash() const {
+	TypeHash seed=0;
+	boost::hash_combine(seed, "Parameter");
+	boost::hash_combine(seed, type_name());
+
+	return seed;
+}
+
+
+bool Parameter::valid_default(const string &str) const {
```
jbrezmorf: Where we need to call this method? This method should be used only in JSONToStorage which works only with IST where Parameter type should not appear. So, this method should throw always.

src/input/type_generic.cc
```diff
@@ -0,0 +1,92 @@
...
+	for (std::vector<TypeBase::ParameterPair>::const_iterator it = parameters_.begin(); it!=parameters_.end(); it++) {
+		boost::hash_combine(seed, (*it).first );
+		boost::hash_combine(seed, (*it).second->content_hash() );
+	}
+
+	return seed;
+}
+
+
+bool Instance::valid_default(const string &str) const {
```
jbrezmorf: This should throw as well.

src/input/type_base.cc
```diff
@@ -80,6 +81,15 @@ void TypeBase::lazy_finish() {
 
 
 
+boost::shared_ptr<TypeBase> TypeBase::substitute_instance_type(boost::shared_ptr<TypeBase> type_ptr) {
```
jbrezmorf: Where this is used? I suppose in lazy_finish. But there we should go trough TypeRepository<Instance> and we know that Instance types are there. So we can call make_instance directly.

src/input/type_generic.cc
```diff
@@ -0,0 +1,92 @@
...
+}
+
+
+const Instance &Instance::close() const {
+	return *( Input::TypeRepository<Instance>::get_instance().add_type( *this ) );
+}
+
+
+// Implements @p TypeBase::make_instance.
+boost::shared_ptr<TypeBase> Instance::make_instance(std::vector<ParameterPair> vec) const {
```
jbrezmorf: This is root of the generic subtree materialization. If we save in attribute only used parameters we can compare this list to the parameters_ and throw in the case of unused parameters. It would be nice to mark the root of the instance by some particular attribute (e.g.  'root_of_instance' : 'true').

src/input/type_record.cc
```diff
@@ -230,6 +231,8 @@ bool Record::check_key_default_value(const Default &dflt, const TypeBase &type,
 
 bool Record::finish()
 {
+	if (this->generic_) return false;
```
jbrezmorf: Not so simple.

src/input/type_record.cc
```diff
@@ -265,6 +268,9 @@ bool Record::finish()
     data_->finished = true;
     for (vector<Key>::iterator it=data_->keys.begin(); it!=data_->keys.end(); it++)
     {
+        ASSERT(typeid( *(it->type_) ) != typeid(Parameter), "Finished Record '%s' can't contain key '%s' of type Parameter.\n",
```
jbrezmorf: 1) Finish of generic types has to be called when they are instantiated.
2) Finish of generic types is an error when called during pass through IST.
Thus we want to distinguish two different favour of finish. This can be done by optional parameter generic_allowed=false.

When called during instantiation we call finish(true), otherwise generic is no allowed. 

This way we doesn't need generic_ flag. Just use this ASSERT only in case generic_allowed == false.

src/input/type_record.cc
```diff
@@ -359,6 +365,48 @@ Record &Record::has_obligatory_type_key() {
 }
 
 
+boost::shared_ptr<TypeBase> Record::make_instance(std::vector<ParameterPair> vec) const {
+	Record rec = Record(this->type_name(), this->data_->description_);
```
jbrezmorf: Use copy constructor.

src/input/type_record.cc
```diff
@@ -359,6 +365,48 @@ Record &Record::has_obligatory_type_key() {
...
+					found = true;
+					key_it->type_ = (*vec_it).second;
+				}
+			}
+			if (!found) xprintf(Warn, "Parameterized key '%s' in make_instance method of '%s' Record wasn't replaced!\n",
+					key_it->key_.c_str(), this->type_name().c_str());
+		}
+	}
+	// Copy attributes
+	for (attribute_map::iterator it=attributes_->begin(); it!=attributes_->end(); it++) {
```
jbrezmorf: Can we just say:  rec.attributes_ = this->attributes_?

src/input/type_record.cc
```diff
@@ -663,6 +718,72 @@ bool AbstractRecord::have_default_descendant() const {
 
 
 
+boost::shared_ptr<TypeBase> AbstractRecord::make_instance(std::vector<ParameterPair> vec) const {
```
jbrezmorf: Reuse record::make_instance

src/input/type_generic.cc
dflanderka: Would be better ASSERT?

src/input/type_record.cc
dflanderka: attributes_ are stored in shared_ptr and we need copy (created instance have added new attributes, original generic record not)

jbrezmorf: Ok, still we can use assignment operator for the map:
 rec.attributes_ = std::make_shared(*attributes_).

I assume 
1) call of Record copy constructor (which in turn call TypeBase copy constructor)
    after that both original and new record attributes_ points to the same map.
2) use make_shared to make new pointer to copy of the map.

src/input/type_base.cc
```diff
@@ -190,6 +194,40 @@ bool Array::valid_default(const string &str) const {
 }
 
 
+boost::shared_ptr<TypeBase> Array::make_instance(std::vector<ParameterPair> vec) const {
+	// Replace only if type_of_values_ is parameter
+	if ( typeid( *(data_->type_of_values_) ) == typeid(Parameter) ) {
+		ASSERT( vec.size() == 1, "Parameter vector of Input::Type::Array '%s' must have size 1!", this->type_name().c_str());
```
jbrezmorf: vec.size()>1 is valid; need parameter name lookup same as in Record::make_instance ...
rather implement Parameter::make_instance which makes parameter name lookup.

Carefully think what we have to do with the type substituted for the parameter. Surly call finish, maybe call make_instance?

src/input/type_record.cc
```diff
@@ -663,6 +729,68 @@ bool AbstractRecord::have_default_descendant() const {
...
+	abstract.close();
+
+	// make instances of all descendant records and add them into instance of abstract
+	for (ChildDataIter child_it = begin_child_data(); child_it != end_child_data(); ++child_it) {
+		Record rec = Record((*child_it).type_name(), (*child_it).data_->description_);
+		// Add parent Abstracts, generic (this) record is replaced by created instance
+		TypeHash hash = this->content_hash();
+		for (auto it = (*child_it).data_->parent_ptr_.begin(); it != (*child_it).data_->parent_ptr_.end(); ++it) {
+			if ( (*it)->content_hash() == hash ) {
+				rec.derive_from( abstract );
```
jbrezmorf: Possible solution, Record can forgot its parent after registration, then Abstract has to call private method Record::add_parent() during its Abstract::finish()

PR #394, opened 2015-07-29
Df generic types new

dflanderka:


src/input/type_base.hh
```diff
@@ -65,6 +65,10 @@ public:
 	typedef std::size_t TypeHash;
 	typedef std::string json_string;
 	typedef std::map<std::string, json_string> attribute_map;
+	typedef std::pair< std::string, boost::shared_ptr<TypeBase> > ParameterPair;
```
jbrezmorf: Add doc for all types.

src/input/type_base.hh
```diff
@@ -139,7 +143,7 @@ public:
      * Finish try to convert all raw pointers pointing to lazy types into smart pointers to valid objects. If there
      * are still raw pointers to not constructed objects the method returns false.
      */
```
jbrezmorf: Documentation for new parameter.

src/input/type_record.cc
```diff
@@ -228,45 +229,21 @@ bool Record::check_key_default_value(const Default &dflt, const TypeBase &type,
...
     for (vector<Key>::iterator it=data_->keys.begin(); it!=data_->keys.end(); it++)
     {
-        if (it->key_ != "TYPE") {
-            data_->finished = data_->finished && const_cast<TypeBase *>( it->type_.get() )->finish();
+    	if (it->key_ != "TYPE") {
+    		ASSERT(is_generic || typeid( *(it->type_) ) != typeid(Parameter),
+    			"Finished non-generic Record '%s' can't contain key '%s' of type Parameter.\n",
+        		this->type_name().c_str(), it->type_->type_name().c_str());
+			if (typeid( *(it->type_.get()) ) == typeid(Instance)) it->type_ = it->type_->make_instance().first;
+            data_->finished = data_->finished && it->type_->finish(is_generic);
```
jbrezmorf: !! This makes data_->finished == false forever. Why?

src/input/type_base.cc
```diff
@@ -149,17 +168,21 @@ TypeBase::TypeHash Array::content_hash() const
...
-bool Array::ArrayData::finish()
+bool Array::ArrayData::finish(bool is_generic)
 {
 	if (finished) return true;
 
-	return (finished = type_of_values_->finish() );
+    ASSERT(is_generic || typeid( *type_of_values_ ) != typeid(Parameter),
+    		"Finished non-generic Array can't be of type Parameter '%s'.\n",
+    		type_of_values_->type_name().c_str());
+
```
jbrezmorf: Remove this assert (as well as similar in Record and AbstractRecord). Replace it by overloaded 
Parameter::finish(...) with an ASSERT(is_generic, ....) reporting actual name of the parameter which is the only helpful information anyway.

src/input/type_base.hh
```diff
@@ -162,6 +166,9 @@ public:
     /// Print JSON output of attributes to @p stream.
     void write_attributes(ostream& stream) const;
 
+    /// Create instance of generic type
+    virtual MakeInstanceReturnType make_instance(std::vector<ParameterPair> vec = std::vector<ParameterPair>()) const =0;
```
jbrezmorf: Document parameters and return value.

src/input/type_base.hh
```diff
@@ -205,9 +212,22 @@ protected:
...
+
     /// map of type attributes (e. g. input_type, name etc.)
     boost::shared_ptr<attribute_map> attributes_;
 
+    /**
+     * Map of used parameters in generic type (for non-generic types is empty).
+     *
+     * Contains all parameter pairs used in type and in all descendants (in subtree of type).
+     */
+    mutable ParameterMap parameter_map_;
```
jbrezmorf: Why this have to be mutable. Put reason into comment.

src/input/type_base.hh
```diff
@@ -297,6 +322,12 @@ public:
      */
     virtual bool valid_default(const string &str) const;
 
+    // Implements @p TypeBase::make_instance.
+    virtual MakeInstanceReturnType make_instance(std::vector<ParameterPair> vec = std::vector<ParameterPair>()) const;
```
jbrezmorf: use override insead of virtual here and for other implementations of virtual methods.

src/input/type_generic.cc
```diff
@@ -0,0 +1,116 @@
...
+	TypeHash seed=0;
+	boost::hash_combine(seed, "Parameter");
+	boost::hash_combine(seed, type_name());
+
+	return seed;
+}
+
+
+bool Parameter::valid_default(const string &str) const {
+    ASSERT(false, "Method valid_default can't be called for Parameter type.\n");
```
jbrezmorf: Report parameter name.

src/input/type_generic.cc
```diff
@@ -0,0 +1,116 @@
...
+
+
+TypeBase::MakeInstanceReturnType Parameter::make_instance(std::vector<ParameterPair> vec) const {
+	for (std::vector<ParameterPair>::iterator vec_it=vec.begin(); vec_it!=vec.end(); vec_it++) {
+		if ( (*vec_it).first == this->name_ ) {
+			this->parameter_map_[(*vec_it).first] = (*vec_it).second->content_hash();
+			return std::make_pair( (*vec_it).second, this->parameter_map_ );
+		}
+	}
+	ASSERT(false, "Parameterized key '%s' in make_instance method wasn't replaced!\n", this->type_name().c_str());
```
jbrezmorf: Imagine you get this error and you know only interface and nothing about implementation ...

I suggest to convert this assert into exception:
"No input type substitution for input type parameter "%s" found during creation of instance with parameter list: ..."
Provide parameter name through EI object.
Then catch this exception in Instance::make_instance and add EI object with list of parameter in the parameter list. This can really help to track down the error.

src/input/type_generic.cc
```diff
@@ -0,0 +1,116 @@
...
+}
+
+
+// Implements @p TypeBase::make_instance.
+TypeBase::MakeInstanceReturnType Instance::make_instance(std::vector<ParameterPair> vec) const {
+	TypeBase::MakeInstanceReturnType ret = generic_type_.make_instance(parameters_);
+#ifdef FLOW123D_DEBUG_ASSERTS
+	for (std::vector<TypeBase::ParameterPair>::const_iterator vec_it = parameters_.begin(); vec_it!=parameters_.end(); vec_it++) {
+		ParameterMap::iterator map_it = ret.second.find( vec_it->first );
+		ASSERT(map_it != ret.second.end(), "Parameter '%s' in make_instance method of '%s' type must be used.\n",
```
jbrezmorf: Do not report type, e.g.
Unused parameter '%s'  in input type instance with parameters: ...
Report list of instance parameters.

src/input/type_generic.hh
```diff
@@ -0,0 +1,82 @@
...
+    const Instance &close() const;
+
+    /// Finish declaration of the Instance type. Call finish of stored @p generic_type_
+    bool finish(bool is_generic = false);
+
+    // Implements @p TypeBase::make_instance.
+    virtual MakeInstanceReturnType make_instance(std::vector<ParameterPair> vec = std::vector<ParameterPair>()) const;
+
+protected:
+	TypeBase &generic_type_;
```
jbrezmorf: documentation.

src/input/type_generic.hh
```diff
@@ -0,0 +1,82 @@
...
+    virtual string type_name() const;
+
+    TypeHash content_hash() const  override;
+
+    virtual bool valid_default(const string &str) const;
+
+    virtual MakeInstanceReturnType make_instance(std::vector<ParameterPair> vec = std::vector<ParameterPair>()) const;
+
+protected:
+	const string name_;
```
jbrezmorf: documentation

src/input/type_generic.hh
```diff
@@ -0,0 +1,82 @@
...
+    /// Finish declaration of the Instance type. Call finish of stored @p generic_type_
+    bool finish(bool is_generic = false);
+
+    // Implements @p TypeBase::make_instance.
+    virtual MakeInstanceReturnType make_instance(std::vector<ParameterPair> vec = std::vector<ParameterPair>()) const;
+
+protected:
+	TypeBase &generic_type_;
+
+	std::vector<TypeBase::ParameterPair> parameters_;
```
jbrezmorf: documentation of data members.

src/input/type_record.cc
```diff
@@ -180,7 +181,7 @@ Record &Record::derive_from(AbstractRecord &parent) {
 	data_->parent_ptr_.push_back( boost::make_shared<AbstractRecord>(parent) );
 
 	if (data_->keys.size() == 0) {
-    	data_->declare_key("TYPE", boost::shared_ptr<TypeBase>(NULL), Default::obligatory(), "Sub-record Selection.");
+    	data_->declare_key("TYPE", boost::shared_ptr<TypeBase>(NULL), Default( type_name() ), "Sub-record Selection.");
```
jbrezmorf: Why this change. Why declare_type_key can not be used?

jbrezmorf: Why this change? I 'm suspicious that this can allow missing TYPE key in abstract records.
Why declare_type_key can not be used?

src/input/type_record.cc
```diff
@@ -228,45 +229,21 @@ bool Record::check_key_default_value(const Default &dflt, const TypeBase &type,
...
-    	data_->keys[0].default_ = Default( type_name() );
-    }
-
     data_->finished = true;
     for (vector<Key>::iterator it=data_->keys.begin(); it!=data_->keys.end(); it++)
     {
-        if (it->key_ != "TYPE") {
-            data_->finished = data_->finished && const_cast<TypeBase *>( it->type_.get() )->finish();
+    	if (it->key_ != "TYPE") {
+    		ASSERT(is_generic || typeid( *(it->type_) ) != typeid(Parameter),
```
jbrezmorf: remove the assert (see comment in Array::make_instance)

src/input/type_record.cc
```diff
@@ -228,45 +229,21 @@ bool Record::check_key_default_value(const Default &dflt, const TypeBase &type,
...
-    	data_->keys[0].type_ = type_copy;
-    	data_->keys[0].default_ = Default( type_name() );
-    }
-
     data_->finished = true;
     for (vector<Key>::iterator it=data_->keys.begin(); it!=data_->keys.end(); it++)
     {
-        if (it->key_ != "TYPE") {
-            data_->finished = data_->finished && const_cast<TypeBase *>( it->type_.get() )->finish();
+    	if (it->key_ != "TYPE") {
```
jbrezmorf: Is it necessary to treat "TYPE" key differently?

src/input/type_record.cc
```diff
@@ -359,6 +337,61 @@ Record &Record::has_obligatory_type_key() {
...
+	rec.attributes_ = boost::make_shared<attribute_map>(*attributes_);
+	return rec;
+}
+
+
+const Record &Record::add_parent(AbstractRecord &parent) const {
+	ASSERT( parent.is_closed(), "Parent AbstractRecord '%s' must be closed!\n", parent.type_name().c_str());
+
+	// check if parent exists in parent_ptr_ vector
+	if ( data_->parent_ptr_.size() ) {
```
jbrezmorf: Uneccessary condition.
Rename parent_ptr_ (since it is not 'ptr' anymore.

src/input/type_record.cc
```diff
@@ -359,6 +337,61 @@ Record &Record::has_obligatory_type_key() {
...
+		TypeHash hash = parent.content_hash();
+		for (auto it = data_->parent_ptr_.begin(); it != data_->parent_ptr_.end(); ++it) {
```
jbrezmorf: Use 'for each' notation: for(auto &parent)

src/input/type_record.cc
```diff
@@ -359,6 +337,61 @@ Record &Record::has_obligatory_type_key() {
...
+			}
+		}
+	}
+
+	data_->parent_ptr_.push_back( boost::make_shared<AbstractRecord>(parent) );
+
+	// finish inheritance
+	ASSERT( data_->keys.size() > 0 && data_->keys[0].key_ == "TYPE",
+				"Derived record '%s' must have defined TYPE key!\n", this->type_name().c_str() );
+	boost::shared_ptr<TypeBase> type_copy = boost::make_shared<Selection>( parent.get_type_selection() );
```
jbrezmorf: This will make trouble in multiple inheritance.  TYPE can not have type of child selection of the parent abstract record since there can be more of them. On the other hand we want to store the actual type in form of index from the selection of particular parent. To make it compatible with the type of the TYPE key, we can declare this key as Interger in the Record, and change implementation in the JSONTo Storage so that the TYPE key is processed in AbstractRecord and skipped in Record. But there are much more places to check, so we postpone all this. Please copy this comment into appropriate issue,

src/input/type_record.cc
```diff
@@ -375,6 +408,22 @@ Record::RecordData::RecordData(const string & type_name_in, const string & descr
 }
 
 
+Record::RecordData::RecordData(const RecordData &other)
```
jbrezmorf: Such a copy constructor should be created automatically by the compiler. Try it.

src/input/type_record.cc
```diff
@@ -420,7 +469,7 @@ Record &Record::declare_key(const string &key, const KeyType &type,
     if (data_->closed_)
         xprintf(PrgErr, "Can not add key '%s' into closed record '%s'.\n", key.c_str(), type_name().c_str());
 
-	check_key_default_value(default_value, type, key);
+    check_key_default_value(default_value, type, key);
```
jbrezmorf: wrong change of indent.

src/input/type_record.cc
```diff
@@ -600,6 +651,13 @@ bool AbstractRecord::finish() {
 	child_data_->finished_ = true;
 
 	child_data_->selection_of_childs->close();
+
+	for (std::vector< Record >::iterator child_it = child_data_->list_of_childs.begin();
```
jbrezmorf: Use short for.

src/input/type_record.cc
```diff
@@ -600,6 +651,13 @@ bool AbstractRecord::finish() {
...
+			child_it != child_data_->list_of_childs.end(); ++child_it) {
+		(*child_it).add_parent(*this);
+		child_data_->finished_ = child_data_->finished_ && (*child_it).finish(is_generic);
```
jbrezmorf: ?? again as in Record::finish.

src/input/type_record.cc
```diff
@@ -663,6 +721,38 @@ bool AbstractRecord::have_default_descendant() const {
 
 
 
+TypeBase::MakeInstanceReturnType AbstractRecord::make_instance(std::vector<ParameterPair> vec) const {
+	AbstractRecord abstract = this->deep_copy();
+
+	// Set close flag - only for add descendants
+	abstract.child_data_->closed_ = true;
```
jbrezmorf: Not clear to me why this. Improve comment.

src/input/type_record.cc
```diff
@@ -663,6 +721,38 @@ bool AbstractRecord::have_default_descendant() const {
...
+	// make instances of all descendant records and add them into instance of abstract
+	for (ChildDataIter child_it = begin_child_data(); child_it != end_child_data(); ++child_it) {
```
jbrezmorf: short for

src/input/type_record.cc
```diff
@@ -663,6 +721,38 @@ bool AbstractRecord::have_default_descendant() const {
...
+		MakeInstanceReturnType inst = (*child_it).make_instance(vec);
+		abstract.add_child( static_cast<Record &>( *(inst.first) ) );
+		abstract.add_to_parameter_map(inst.second);
+	}
+	// Unset close flag - necessary for set parameters
+	abstract.child_data_->closed_ = false;
+
+	// Set parameters and generic type as attributes
+	abstract.add_attribute("parameters", abstract.print_parameter_map_to_json());
```
jbrezmorf: This code seems to repeat in Array, Record and here, could we make appropriate method in TypeBase.

jbrezmorf:
Still good load of comments, but mainly simplifications and documentation requests.

src/input/type_record.cc
```diff
@@ -600,6 +651,13 @@ bool AbstractRecord::finish() {
 	child_data_->finished_ = true;
```
jbrezmorf: Move this line just before the loop.

dflanderka: Done.

src/input/type_record.cc
dflanderka: Done.

src/input/type_record.cc
dflanderka: Yes, TYPE is selection that is set to NULL in first step. Not-null selection is set later from AbstractRecord and it can be finished after call of Record::finish().

src/input/type_record.cc
dflanderka: - set of Default is moved from Record::finish
- abstract records haven't keys, record is registered into abstract record in add_child method what is called during Record::close()
- declare_type_key set Default::obligatory

src/input/type_record.cc
jbrezmorf: OK, setting  type  of the key TYPE to NULL in derive_from is error prone and needs checks like this one. What about to set it to an empty selection. Then we still can distinguish completed TYPE keys if it is necessary but make it litle bit more safe. This change however needs careful check of all places dealing with the type of TYPE.

PR #406, opened 2015-08-10
Pe numbering

Paulie14:
Testing pull request to see, what happens when VF reference element is applied.
Only some corrections suggested.
Is branched from PE_removemhfe.


src/flow/darcy_flow_mh.cc
```diff
@@ -929,8 +929,11 @@ void P1_CouplingAssembler::assembly(LinSys &ls) {
         // base fce = a0 + a1*t0 + a2*t1
         //         a0     a1      a2
         base_2D << 1.0 << 0.0 << -2.0 << arma::endr //point on side 0
```
Paulie14: I don't know where this matrix is coming from, but it should be made independent of ref. element numbering.

jbrezmorf:
merged into master with JB_numbering

PR #407, opened 2015-08-11
JHy Windows Python embedded test

x3mSpeedy:
Template now contains step **python_test** which will using [this](https://github.com/x3mSpeedy/windows-install-flow123d/blob/master/install.py) script, run test ```03_transport_small_12d``` on windows testing node. If Python is Embedded correctly there will be  profiler generated also in txt format. Condition is to test packgage from bacula which contains new python changes, so far testing branch ```0.0.master```

Other changes:
 - remove whitespace in document
 - fix comment indentation

### DIFF
--- a/unit_tests/jenkins_jobs_template.yaml
+++ b/unit_tests/jenkins_jobs_template.yaml
@@ -19,11 +19,11 @@
     name: "Multijob project debug"
     platform: linux
     build-type: debug
-    unit-dir: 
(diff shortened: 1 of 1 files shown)


jbrezmorf:
Please, remove white space changes.

jbrezmorf:
Seems that this commit is already part of HTML pull request. Will merge there.

PR #408, opened 2015-08-11
JHy ist formatter

x3mSpeedy:
Pull which handles HTML documentation creation (and publishing to server). Major changes:
 1. HTML creation
    There is now new CMake lib which handles creation of html documentation (actually only part of it)
    During flow123d build file is created in ```${CMAKE_BINARY_DIR}/htmldoc``` which contains following steps:
    *  **downloading** zip package which is uploaded on random server
    * **unpack** zip package into given location

    There is new **makefile** step ```make html-doc``` which
    * runs cmake which downloads ands unpacks css/js/fonts zip ball from server
    * runs flow123d bin and generates JSON input specification file
    * calls python to transforms JSON to HTML

 2. HTML publish
    In order to publish HTML, shell script ```prepare_to_publish.sh``` now copies html resources to ```publish_dir```

 3. Documentation string change
    In order to produce valid JSON some doc string were slightly edited (talking about escaping) some doc string were also edited but in order to fully support markdown syntax on docs strings (lists, code blocks, ...)

 4. Change of ```flow123d --JSON_machine output_file.json```
    JSON file is now generated into location specified during call. Other debug python-related output messages during execution of flow were muted


CMake/Modules/InstallHTMLDOC.cmake
```diff
@@ -0,0 +1,30 @@
+# create CMakeLists.txt for external project
+# required variables:
+## HTMLDOC_INSTALL_URL - from where to download file (zip/tarball)
+## HTMLDOC_INSTALL_DIR - directory where helper files will be stored such as CMakeLists.cmake file makefile
+## HTMLDOC_SOURCE_DIR - direcotry which will contain unpacked files from server
```
jbrezmorf: Is it REALLY necessary to set these variables out of this file? Its better to put this setting here since you can find it easier then in the already huge CMakeLists.txt

CMakeLists.txt
```diff
@@ -657,6 +657,17 @@ message(STATUS "YamlCpp_INCLUDE_DIR: ${YamlCpp_INCLUDE_DIR}")
 message(STATUS "\n ======================================================= ")
 
 
+set(HTMLDOC_INSTALL_URL \"http://hybs.nti.tul.cz/test/ist.zip\")
```
jbrezmorf: Move into InstallHTMLDOC.

makefile
```diff
@@ -157,9 +170,9 @@ checkout-submodule-branches:
 #	  > $(DOC_DIR)/flow_version.tex
 
 # call flow123d and make raw input_reference file
-$(DOC_DIR)/input_reference_raw.tex: update-build-tree build-flow123d	 	
-	$(BUILD_DIR)/bin/flow123d --latex_doc | grep -v "DBG" | \
-	sed 's/->/$$\\rightarrow$$/g' > $(DOC_DIR)/input_reference_raw.tex
+$(DOC_DIR)/input_reference_raw.tex: update-build-tree build-flow123d
```
jbrezmorf: Just remove this rule.

src/fields/field_constant.impl.hh
```diff
@@ -28,9 +28,9 @@ const Input::Type::Record & FieldConstant<spacedim, Value>::get_input_type(
                                     "Value of the constant field.\n"
                                     "For vector values, you can use scalar value to enter constant vector.\n"
                                     "For square NxN-matrix values, you can use:\n"
-                                    "* vector of size N to enter diagonal matrix\n"
-                                    "* vector of size (N+1)*N/2 to enter symmetric matrix (upper triangle, row by row)\n"
-                                    "* scalar to enter multiple of the unit matrix." )
+                                    " - vector of size N to enter diagonal matrix\n\n"
+                                    " - vector of size (($(N+1)*N/2$)) to enter symmetric matrix (upper triangle, row by row)\n"
+                                    " - scalar to enter multiple of the unit matrix." )
         .allow_auto_conversion("value")
```
jbrezmorf: replace NxN with (($N\timesN$))
replace N with (($N$))
replace (($(N+1)*N/2$)) with (($\frac12N(N+1)$))

... in order to test formulas support in both HTML and Latex little bit.

src/fields/field_formula.impl.hh
```diff
@@ -30,11 +30,11 @@ const Input::Type::Record & FieldFormula<spacedim, Value>::get_input_type(
...
-                                        "For square NxN-matrix values, you can use:\n"
-                                        "* array of strings of size N to enter diagonal matrix\n"
-                                        "* array of strings of size (N+1)*N/2 to enter symmetric matrix (upper triangle, row by row)\n"
-                                        "* just one string to enter (spatially variable) multiple of the unit matrix.\n"
-                                        "Formula can contain variables x,y,z,t and usual operators and functions." )
+                                        "For square NxN-matrix values, you can use:\n\n"
+                                        " - array of strings of size N to enter diagonal matrix\n"
+                                        " - array of strings of size (($(N+1)*N/2$)) to enter symmetric matrix (upper triangle, row by row)\n"
+                                        " - just one string to enter (spatially variable) multiple of the unit matrix.\n"
+                                        "Formula can contain variables ```x,y,z,t``` and usual operators and functions." )
```
jbrezmorf: Same as for FieldConstant.

src/flow/darcy_flow_mh.cc
```diff
@@ -104,7 +104,7 @@ const it::Selection & DarcyFlowMH::EqData::get_bc_type_selection() {
                        "Specify the pressure head through the 'bc_pressure' field "
                        "or the piezometric head through the 'bc_piezo_head' field.")
                .add_value(neumann, "neumann", "Neumann boundary condition. Prescribe water outflow by the 'bc_flux' field.")
-               .add_value(robin, "robin", "Robin boundary condition. Water outflow equal to $\\sigma (h - h^R)$. "
+               .add_value(robin, "robin", "Robin boundary condition. Water outflow equal to (($sigma (h - h^R)$)). "
```
jbrezmorf: need back slash before sigma:
valid Latex formula is:
\sigma(h-h^R)
how to deal with it? I suggest writing single backslash here and do correct escaping in JSONmachine output if it is not done yet.

src/main.cc
```diff
@@ -232,8 +232,12 @@ void Application::parse_cmd_line(const int argc, char ** argv) {
     }
 
     if (vm.count("JSON_machine")) {
+        // write ist to json file
+        string json_filename = vm["JSON_machine"].as<string>();
+        ofstream json_stream(json_filename);
```
jbrezmorf: have to check that file is open, and trow with correct error message.

src/transport/transport_dg.cc
```diff
@@ -71,7 +71,7 @@ const Selection & TransportDG<Model>::EqData::get_bc_type_selection() {
                        //"or the piezometric head through the 'bc_piezo_head' field."
                       )
                .add_value(neumann, "neumann", "Neumann boundary condition. Prescribe water outflow by the 'bc_flux' field.")
-               .add_value(robin, "robin", "Robin boundary condition. Water outflow equal to $\\sigma (h - h^R)$. "
+               .add_value(robin, "robin", "Robin boundary condition. Water outflow equal to (($sigma (h - h^R)$)). "
```
jbrezmorf: Same as before.

unit_tests/jenkins_jobs_template.yaml
```diff
@@ -19,11 +19,11 @@
     name: "Multijob project debug"
```
jbrezmorf: remove whitspace changes

CMake/Modules/InstallHTMLDOC.cmake
x3mSpeedy: moved to cmake module

x3mSpeedy:
Whitespace changes were reverted (they were automatically caused by GitHub's editor Atom)

x3mSpeedy:
@jbrezmorf  In order for JSON to be parsed, validity is needed. Which means single slash would case parsing error. Correct way to write slash ```\``` char in JSON file is to escape slash with another slash ```\\```

Combining with c++ language where single slash is written as two. Four slashes are needed for one JSON slash
```cc
// C++
string foo = "(($\\\\sigma (h - h^R)$))";
```
will produce desired result:
```js
// JSON
{
 "foo": "(($\\sigma (h - h^R)$))"
}
```
which when read will as one slash in every parser due to JSON specification:
```python
# python
print foo
$> "(($\sigma (h - h^R)$))"
```

Two slashes in c++ would cause parsing error so before parsing there would have to be script which would fix this escaping issue.

jbrezmorf:
Concerning escaping: The JSON serialization (JSONmachine output) should do an escaping. So in C++ code one should write:
"\\sigma" which is than escaped during output to "\\\\sigma" which will be correctly parsed later on.
I thought that backslash escaping is done during JSONmachine output, but I may by wrong.

jbrezmorf:
Checking src/input/type_output.cc reveald that:
1. escaping of description strings is done in method: OutputJSONMachine::escape_description
2. escaping is incomplete

Please, modify this method to make correct escaping of backslash. Is there any other character that needs escaping?
Then simplify formulas  in descriptions using just two backslash notation, which is quite acceptable.

x3mSpeedy:
According to [json specification](http://json.org/) slash ```\``` is only character thus cannot by left alone (hence two slashes are needed). There are other special characters but since they are escapable by slash there should be no problem. 
> any Unicode character except ```"``` or ```\``` or control character
>\"
>\\\\
>\/
>\b \f \n \r \t
>\u four-hex-digits

x3mSpeedy:
@jbrezmorf current status is that following example in C++
 ```cc
std::string description = "N\\times N \n continue here";
```
will generate **valid** JSON which will contain following expression
```json
{
    "description": "N\\times N \n continue here"
}
```

If we want to use other special chars in description (```\t \b \r " ```) they also need to be added. I also propose to create ```boost:regex``` variables beforehand to avoid unnecessary computation since these regular expression are built for every description.

PR #410, opened 2015-08-14
Df yaml reader

dflanderka:


src/input/json_to_storage.cc
```diff
@@ -465,18 +871,18 @@ StorageBase * JSONToStorage::make_storage(JSONPath &p, const Type::AbstractRecor
...
             THROW( ExcInputError() << EI_Specification("The value should be 'JSON object', but we found: ")
-                << EI_ErrorAddress(p) << EI_JSON_Type( json_type_names[ p.head()->type() ] ) << EI_InputType(abstr_rec->desc()) );
+                << EI_ErrorAddress(p) << EI_JSON_Type( p->get_node_type() ) << EI_InputType(abstr_rec->desc()) );
         } else { // auto conversion
         	return abstract_rec_automatic_conversion(p, abstr_rec);
         }
     }
 
-    return NULL;
+    return NULL;*/
```
dflanderka: Remove unused code.

src/fields/field.impl.hh
```diff
@@ -437,7 +437,7 @@ void Field<spacedim,Value>::check_initialized_region_fields_() {
     	// has to deal with fact that reader can not deal with input consisting of simple values
     	string default_input=input_default();
     	auto input_type = get_input_type();
-        Input::JSONToStorage reader( default_input, input_type );
+        Input::JSONToStorage reader( default_input, input_type, Input::FileFormat::format_JSON ); // TODO set correct Input::FileFormat
```
jbrezmorf: Remove TODO comment.

src/input/json_to_storage.hh
```diff
@@ -37,24 +38,22 @@ namespace Input {
...
- *
- * This class keeps whole path from the root of the JSON tree to the current node. We store nodes along path in \p nodes_
- * and address of the node in \p path_.
- *
- * The class also contains methods for processing of special keys 'REF' and 'TYPE'. The reference is record with only one key
- * 'REF' with a string value that contains address of the reference. The string with the address is extracted by \p JSONToStorage::get_ref_from_head
- * then the JSONPath corresponding to the address is provided by method \p JSONtoStorage::find_ref_node.
- */
-class JSONPath {
+/// Result type have sense only for larger Value types like vectors and tensors.
```
jbrezmorf: Do not understand meaning of the comment.

src/input/json_to_storage.hh
```diff
@@ -67,55 +66,83 @@ public:
...
+    virtual double get_double_value() const =0;
+
+    /// Get string value of head node or throw exception
+    virtual std::string get_string_value() const =0;
+
+    /// Get short string description of head type, method is used for printout of messages
+    virtual std::string get_node_type() const =0;
+
+    /// Get set of keys of head type record, if head type is not record return false
+    virtual bool get_record_key_set(std::set<std::string> &) const =0;
```
jbrezmorf: Why not map?

src/input/json_to_storage.hh
```diff
@@ -67,55 +66,83 @@ public:
...
+
+    /// Get size of array (sequence type), if object is not array return -1
+    virtual int get_array_size() const =0;
+
+    /// Check if type of head node is map (record)
+    virtual bool is_map_type() const =0;
```
jbrezmorf: rename to is_record_type

src/input/json_to_storage.hh
```diff
@@ -67,55 +66,83 @@ public:
...
+
+    /// Check if type of head node is sequence (array)
+    virtual bool is_sequence_type() const =0;
```
jbrezmorf: rename to is_array_type

src/input/json_to_storage.hh
```diff
@@ -123,27 +150,213 @@ public:
     string str();
 
     /**
-     * Put actual address to previous_references_ set
+     * Check if index of Abstract descendant can be loaded from TYPE key (return true)
+     * or return false if automatic conversion can be performed. In other cases throw
+     * input exception.
      */
-    void put_address();
+    virtual bool has_descendent_index(bool value_at_declaration) =0;
```
jbrezmorf: OK, how we get index of descendent.

src/input/json_to_storage.cc
```diff
@@ -22,46 +22,184 @@ using namespace internal;
...
+            stream << "/" << it->second;
+        } else {
+            stream << "/" << it->first;
+        }
+    }
+}
+
+
+
+void PathBase::put_address() {
```
jbrezmorf: rename to remember_reference, improve documentation in header file

src/input/json_to_storage.cc
```diff
@@ -22,46 +22,184 @@ using namespace internal;
...
+    return ss.str();
+}
+
+
+
+/**
+ * This returns path to reference given by address in ref_address.
+ *
+ */
+PathBase * PathBase::find_ref_node(const string& ref_address)
```
jbrezmorf: TO CONTINUE HERE ...

src/input/json_to_storage.cc
```diff
@@ -22,46 +22,184 @@ using namespace internal;
...
+
+
+
+void PathBase::put_address() {
+	previous_references_.insert(str());
+}
+
+
+
+string PathBase::str() {
```
jbrezmorf: rename to 'as_string'.

src/input/json_to_storage.cc
jbrezmorf: Since we use YAML references we need this method only in JSON reader so possibly we can move this method as well as previous_references_ stuff into PathJSON.

src/input/json_to_storage.cc
```diff
@@ -85,7 +223,7 @@ bool JSONPath::get_ref_from_head(string & ref_address)
 
     const Node &ref_node = obj.begin()->second;
     if (ref_node.type() != json_spirit::str_type) {
-        THROW( ExcRefOfWrongType() << EI_ErrorAddress(*this) );
+        THROW( ExcRefOfWrongType() << EI_ErrorAddress(this) );
```
jbrezmorf: Passing pointer is not safe. We probably need to use 'as_string' method just here.

src/input/json_to_storage.cc
```diff
@@ -93,106 +231,362 @@ bool JSONPath::get_ref_from_head(string & ref_address)
...
-    string::size_type new_pos = 0;
-    string address = ref_address + '/';
-    string tmp_str;
-    bool relative_ref = false;
+bool PathJSON::get_bool_value() const {
+    if (head()->type() == json_spirit::bool_type) {
+        return head()->get_bool();
+    } else {
+        THROW( JSONToStorage::ExcInputError() << JSONToStorage::EI_Specification("The value should be 'JSON bool', but we found: ")
+                << JSONToStorage::EI_ErrorAddress(this) << JSONToStorage::EI_Format("JSON") << JSONToStorage::EI_JSON_Type( get_node_type() )
```
jbrezmorf: unsafe

src/input/json_to_storage.cc
```diff
@@ -93,106 +231,362 @@ bool JSONPath::get_ref_from_head(string & ref_address)
...
+
+
+std::int64_t PathJSON::get_int_value() const {
+    if (head()->type() == json_spirit::int_type) {
+        return head()->get_int64();
     } else {
-    	THROW( ExcReferenceNotFound() << EI_RefAddress(*this) << EI_ErrorAddress(ref_path) << EI_RefStr(ref_address)
-    	       << EI_Specification("cannot follow reference") );
+        THROW( JSONToStorage::ExcInputError() << JSONToStorage::EI_Specification("The value should be 'JSON int', but we found: ")
+                << JSONToStorage::EI_ErrorAddress(this) << JSONToStorage::EI_Format("JSON") << JSONToStorage::EI_JSON_Type( get_node_type() )
```
jbrezmorf: unsafe

src/input/json_to_storage.cc
```diff
@@ -93,106 +231,362 @@ bool JSONPath::get_ref_from_head(string & ref_address)
...
-                        << EI_Specification("index out of size of Array") );
-            }
+double PathJSON::get_double_value() const {
+    auto value_type = head()->type();
+    if (    value_type== json_spirit::real_type
+         || value_type == json_spirit::int_type) {
+        return head()->get_real();
+    } else {
+        THROW( JSONToStorage::ExcInputError() << JSONToStorage::EI_Specification("The value should be 'JSON real', but we found: ")
+                << JSONToStorage::EI_ErrorAddress(this) << JSONToStorage::EI_Format("JSON") << JSONToStorage::EI_JSON_Type( get_node_type() )
```
jbrezmorf: unsafe

src/input/json_to_storage.cc
```diff
@@ -93,106 +231,362 @@ bool JSONPath::get_ref_from_head(string & ref_address)
...
+}
+
+
+
+bool PathJSON::has_descendent_index(bool value_at_declaration) {
+	if ( this->is_map_type() ) {
+
+		PathBase *type_path = this->clone();
+		if ( !type_path->down("TYPE") ) {
+            if ( !value_at_declaration ) {
```
jbrezmorf: Wrong indent.

src/input/json_to_storage.cc
```diff
@@ -93,106 +231,362 @@ bool JSONPath::get_ref_from_head(string & ref_address)
...
+
+bool PathYAML::get_ref_from_head(string & ref_address)
+{
+	return false;
+}
+
+
+bool PathYAML::has_descendent_index(bool value_at_declaration) {
+	const Node & head_node = *( nodes_.back() );
+	if (head_node.Tag() == "!type") {
```
jbrezmorf: This still requires key 'TYPE', but we want to use tag for type specification. So the interface in PathBase needs to be different.

src/input/json_to_storage.cc
```diff
@@ -22,46 +22,184 @@ using namespace internal;
...
     }
-    return nodes_.back();
+    return (nodes_.back() != NULL);
 }
 
 
-void JSONPath::up()
+void PathJSON::up()
 {
     if (path_.size() > 1) {
```
jbrezmorf: Possible memory leak. We should delete the pointer first. However is it safe? JSONToStorage have to read the values before up() is called. It should be OK, just check it.
Same issue with PathYAML::up()

src/input/json_to_storage.cc
```diff
@@ -93,106 +231,362 @@ bool JSONPath::get_ref_from_head(string & ref_address)
...
+    const Node & head_node = *( nodes_.back() );
+    ASSERT(head_node.IsMap(), "Head node must be of type Record.\n");
+
+    if (head_node[key]) {
+    	path_.push_back( make_pair( (int)(-1), key) );
+    	nodes_.push_back( new Node( head_node[key] ) );
+    } else {
+        return false;
     }
+    return (nodes_.back() != NULL);
```
jbrezmorf: Well, how we can get 'false' here? Just if allocation fails, but when using new operetor it doesn't fail just throws.
So we should rather return true here, same for all other Path*::down(*).

src/input/json_to_storage.cc
```diff
@@ -93,106 +231,362 @@ bool JSONPath::get_ref_from_head(string & ref_address)
...
+        nodes_.pop_back();
+    }
+}
+
 
-string JSONPath::str() {
-    stringstream ss;
-    output(ss);
-    return ss.str();
+void PathYAML::go_to_root() {
```
jbrezmorf: Again memory leak. I propose to implement go_to_root in PathBase using up() and level().

src/input/json_to_storage.cc
```diff
@@ -93,106 +231,362 @@ bool JSONPath::get_ref_from_head(string & ref_address)
...
+             );
     }
-    if (relative_ref) {
-    	xprintf(Msg, "Referencing '%s' to '%s'.\n", this->str().c_str(), ref_path.str().c_str());
+	return "";
+}
+
+
+
+std::string PathJSON::get_node_type() const {
```
jbrezmorf: Use same implementation as in PathYAML, or at least make json_type_names a local variable in this method.

src/input/json_to_storage.cc
```diff
@@ -281,22 +671,22 @@ void JSONToStorage::read_stream(istream &in, const Type::TypeBase &root_type)
...
     string ref_address;
-    if (p.get_ref_from_head(ref_address)) {
+    if (p->get_ref_from_head(ref_address)) {
         // todo: mark passed references and check cyclic references
 
         // dereference and take data from there
-        JSONPath ref_path = p.find_ref_node(ref_address);
+    	PathBase * ref_path = p->find_ref_node(ref_address);
         return make_storage( ref_path, type );
     }
```
jbrezmorf: Need to delete ref_path !! Why is it necessary to convert all references (mostly safe) to pointers?

src/input/json_to_storage.cc
```diff
@@ -249,26 +647,18 @@ void JSONToStorage::read_stream(istream &in, const Type::TypeBase &root_type)
...
+	} else {
+		root_path = new PathYAML(in);
+	}
 
     root_type_ = &root_type;
     storage_ = make_storage(root_path, root_type_);
 
-    ASSERT(  storage_ != nullptr, "Internal error in JSON reader, the storage pointer is NULL after reading the stream.\n");
+    ASSERT(  storage_ != nullptr, "Internal error in %s reader, the storage pointer is NULL after reading the stream.\n",
+    		root_path->input_format_name().c_str());
```
jbrezmorf: Need to delete root_path.

src/input/json_to_storage.cc
```diff
@@ -371,15 +755,15 @@ StorageBase * JSONToStorage::make_storage(JSONPath &p, const Type::Record *recor
...
                 // key on input => check & use it
                 storage_array->new_item(it->key_index, make_storage(p, it->type_.get()) );
-                p.up();
+                p->up();
             } else {
                 // key not on input
                 if (it->default_.is_obligatory() ) {
                     THROW( ExcInputError() << EI_Specification("Missing obligatory key '"+ it->key_ +"'.")
-                            << EI_ErrorAddress(p) << EI_InputType(record->desc()) );
+                            << EI_ErrorConstAddress(p) << EI_Format( p->input_format_name() ) << EI_InputType(record->desc()) );
```
jbrezmorf: Probably better to catch the exception in read_storage and add EI_Format in order to reduce code duplicity.
That possibly may allow removing input_format_name.

src/input/json_to_storage.cc
```diff
@@ -440,12 +825,38 @@ StorageBase * JSONToStorage::record_automatic_conversion(JSONPath &p, const Type
...
+		return abstract_rec_automatic_conversion(p, abstr_rec);
+	}
+
+	unsigned int descendant_index;
+	PathBase *type_path = p->clone();
+	type_path->down("TYPE");
+    try {
+        // convert to base type to force type dispatch and reference chatching
+        const Type::TypeBase * type_of_type = &( abstr_rec->get_type_selection() );
+        descendant_index = (unsigned int)make_storage(type_path, type_of_type )->get_int();
```
jbrezmorf: Can not use this for YAML.

src/input/json_to_storage.hh
```diff
@@ -123,27 +150,213 @@ public:
...
+    virtual std::string get_string_value() const override;
+    virtual std::string get_node_type() const override;
+    virtual bool get_record_key_set(std::set<std::string> &) const override;
+    virtual int get_array_size() const override;
+    virtual bool is_map_type() const override;
+    virtual bool is_sequence_type() const override;
+    virtual PathJSON * clone() const override;
+    virtual bool has_descendent_index(bool value_at_declaration) override;
+
+    virtual inline std::string input_format_name() const override {
```
jbrezmorf: virtual methods can not be inlined.

src/input/json_to_storage.hh
```diff
@@ -123,27 +150,213 @@ public:
...
+    	return "JSON";
+    }
+
+    virtual inline std::string sequence_name() const override {
```
jbrezmorf: Why do not use get_node_type instead of these two methods?

src/input/json_to_storage.hh
```diff
@@ -123,27 +150,213 @@ public:
...
+    { return nodes_.size() - 1; }
+
+    /**
+     * Check if current head node is a JSON Object containing one key REF of type string.
+     * If yes, returns the string through reference @p ref_address.
+     */
+    bool get_ref_from_head(string & ref_address) override;
+
+    // These methods are derived from PathBase
+    virtual bool is_null_type() const override;
```
jbrezmorf: Use just override no virtual in child classes. This has sense only if you want to allow PathJSON to be a parent of some other class.

jbrezmorf:
Why we need to convert references to PathBase to pointers?
Check all new calls to have appropriate delete calls. Be extremely careful when returning pointer to make proper delete.

src/input/json_to_storage.hh
```diff
@@ -104,46 +225,140 @@ public:
...
+    bool is_record_type() const override;
+    bool is_array_type() const override;
+    PathJSON * clone() const override;
+    bool has_descendent_index(bool value_at_declaration) override;
+
+    inline std::string input_format_name() const override {
+    	return "JSON";
+    }
+
+    inline std::string sequence_name() const override {
```
jbrezmorf: virtual method can be inlined. Moreover this way we do not guarantee that string returned by this method mach type name returned by get_node_type. Solution:
1) in constructor PathYAML, PathJSON make array of type name strings with common index for:
     - NULL
     - JSON object, YAML map
     - JSON array, YAML sequence
     some positions may be empty (e.g. Scalar for JSON)
2) Implement these methods in PathBase (reading particular index from the type name array)
3) in get_node_type use case to get particular items of the array.

src/input/json_to_storage.cc
```diff
@@ -440,49 +825,45 @@ StorageBase * JSONToStorage::record_automatic_conversion(JSONPath &p, const Type
...
+	bool has_desc_index;
+	try {
+		has_desc_index = p->has_descendent_index( abstr_rec->get_selection_default().has_value_at_declaration() );
+	} catch(ExcInputError &e) {
+		e << EI_InputType(abstr_rec->desc());
+		throw;
+	}
 
-        JSONPath type_path(p);
-        if ( type_path.down("TYPE") == NULL ) {
```
jbrezmorf: replace this condition with call to get_descendant_name and test it against empty string.

src/input/json_to_storage.cc
dflanderka: We can delete node only in PathYAML::up() where copy of node is stored to nodes vector. In PathJSON is directly stored nodes from json_spirit tree and we can't delete them (it throws segfault).

jbrezmorf: That is OK since we do not allocate them.

src/input/json_to_storage.cc
```diff
@@ -440,49 +805,50 @@ StorageBase * JSONToStorage::record_automatic_conversion(JSONPath &p, const Type
...
+			if ( ! abstr_rec->get_selection_default().has_value_at_declaration() ) {
+				THROW( ExcInputError() << EI_Specification("Missing key 'TYPE' in AbstractRecord.") << EI_ErrorAddress(p.as_string()) << EI_InputType(abstr_rec->desc()) );
+			} else { // auto conversion
+				return abstract_rec_automatic_conversion(p, abstr_rec);
+			}
+		} else {
+			try {
+				unsigned int descendant_index = (unsigned int)abstr_rec->get_type_selection().name_to_int( descendant_name );
+				return make_storage(p, &( abstr_rec->get_descendant(descendant_index) ) );
+			} catch (ExcInputError & e) {
```
jbrezmorf: I do not agree with this catch. It was copied here from make_storage for Selection, but
here it would overwrite a correct information coming form make_storage called form the descendant.
I think we can just remove this catch.

src/input/json_to_storage.cc
```diff
@@ -121,78 +181,381 @@ JSONPath JSONPath::find_ref_node(const string& ref_address)
...
+	} else {
+        THROW( JSONToStorage::ExcInputError() << JSONToStorage::EI_Specification("The value should be 'YAML string', but we found: ")
+                << JSONToStorage::EI_ErrorAddress(this->as_string()) << JSONToStorage::EI_JSON_Type( get_node_type() )
+             );
+	}
+	return "";
+}
+
+
+std::string PathYAML::get_node_type() const {
```
jbrezmorf: 1) vector of type names declared in PathBase 
2) fill type names vector in PathJSON and PathYAML, use common position for null, map, sequence;
leave Scalar possition oempty for JSON and vice versa for individaul scalar types in YAML.
3) move get_map_name, get_scalar_name implementation to PathBase, just returning the item of type names vector
4) rename those methods to get_record_reader_type, get_array_reader_type
5) in PathYAML::get_node_type return appropriate items of type names vector.

jbrezmorf:
Still few things to fix, see new notes.
Further merge master, resolve conflicts

PR #411, opened 2015-08-24
Pe assembly template

Paulie14:
Trying to hide all the dim-switches into templated<dim> assembly class in darcy flow.
Any further suggestions?

### DIFF
--- a/src/flow/darcy_flow_mh.cc
+++ b/src/flow/darcy_flow_mh.cc
@@ -77,7 +77,7 @@
 #include "transport/mass_balance.hh"
 #include "input/factory.hh"
 
-
+#include "fields/vec_seq_double.hh"
 
 
 FLOW123D_FORCE_LINK_IN_CHILD(steady_MH);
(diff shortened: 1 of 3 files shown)


src/flow/darcy_flow_mh.cc
```diff
@@ -235,6 +235,59 @@ DarcyFlowMH::EqData::EqData()
 
 }
 
+template<>
+DarcyFlowMH_Steady::Assembly<1>::Assembly()
+: quad_(3),
+  fe_values_(map_, quad_, fe_rt_, 
+            update_values | update_gradients | update_JxW_values | update_quadrature_points),
+  side_quad_(1),
+  fe_p_disc_(new FE_P_disc<1,1,3>()),
```
jbrezmorf: Do you have any particular reason for this specialization?
I even wonder that it compiles since C++ do not support specialization of single method of a class template.
Possibly its just ignored, since it seems to work well without it.

jbrezmorf:
just merge into other topic branch.

PR #421, opened 2015-09-01
Js transport bc

### LINKED ISSUE #414: combined BC for TransportDG
- [x] Implement total_flux and diffusive_flux BC in TransporDG
see #367 
- [x] Allow possibly different BC types in Heat and Concentration model
- [x] Improve description of BC types in the input, it already support formulas in form (($latex formula$)), 
check e.g. html result with 'make html-doc'

jstebel:
Proposal of transport and heat boundary conditions. If OK then some tests need to be added.

Resolves #414.


src/transport/heat_model.cc
```diff
@@ -42,22 +42,79 @@ using namespace std;
 using namespace Input::Type;
 
 
-
-
-
-
-
+const std::map<unsigned int,unsigned int> HeatTransferModel::ModelEqData::bc_type_conversion = {
```
jbrezmorf: Why this conversion map?

src/transport/concentration_model.cc
```diff
@@ -42,7 +42,39 @@ using namespace std;
 using namespace Input::Type;
 
 
-
+const std::map<unsigned int,unsigned int> ConcentrationTransportModel::ModelEqData::bc_type_conversion = {
```
jbrezmorf: Why this conversion map?

src/transport/concentration_model.cc
```diff
@@ -50,12 +82,38 @@ using namespace Input::Type;
 ConcentrationTransportModel::ModelEqData::ModelEqData()
 : TransportBase::TransportEqData()
 {
+    *this+=bc_type
+            .name("bc_type")
+            .description(
+            "Boundary condition type, possible values: inflow, dirichlet, neumann, robin.")
```
jbrezmorf: Is this correct?

src/transport/heat_model.cc
```diff
@@ -300,6 +364,67 @@ void HeatTransferModel::compute_dirichlet_bc(const std::vector<arma::vec3> &poin
...
+	for (unsigned int i=0; i<np; i++)
+	{
+		bc_flux[i] = bc_flux_scalar[i];
+		bc_ad_value[i] = bc_ad_temperature[i];
+		bc_sigma[i] = bc_robin_sigma[i];
+		bc_ref_value[i] = bc_temperature[i];
+	}
+}
+
+void HeatTransferModel::get_diffusive_flux_bc_data(const std::vector<arma::vec3> &point_list,
```
jbrezmorf: Do not try to make fields of heat model scalar at this time. This makes lot of boiler plate and could be probably avoided when we use multifields (we have to do that soon, should discus it).

src/transport/advection_diffusion_model.hh
```diff
@@ -85,16 +93,80 @@ public:
...
+			arma::uvec &bc_types) = 0;
+
+	/**
 	 * Computes the Dirichlet boundary condition values.
 	 * @param point_list   Points at which to evaluate.
 	 * @param ele_acc      Element accessor.
 	 * @param bc_values    Vector of b.c. values (output).
 	 */
-	virtual void compute_dirichlet_bc(const std::vector<arma::vec3> &point_list,
+	virtual void get_dirichlet_bc_data(const std::vector<arma::vec3> &point_list,
```
jbrezmorf: Why we have virtual functions in the Model. The Idea to make TransportDG templated by the model was to aviod virtual calls.

jbrezmorf:
Conclusions:
- I rather go through the documentation together.
- Need to discuss usage of multifields and how it can help in transport.
- Have to find way how to minimize boiler plate code. Choose wisely interface of Model.

src/transport/heat_model.cc
jstebel: For the moment it seems redundant, but this allows definition of different bc types (names) for Model classes. We could have e.g. "insulated_wall" in heat transfer or "precipitation" in solute transport.

src/transport/concentration_model.cc
jstebel: See comment to heat_model.cc above.

src/transport/concentration_model.cc
jstebel: Right, I will fix this.

src/transport/advection_diffusion_model.hh
jstebel: Virtual methods are just for declaration of the common interface. TransportDG does not use virtual calls because it accesses directly the methods from Model.

jstebel:
All comments to the pull request incorporated - to be merged.

src/transport/concentration_model.hh
```diff
@@ -45,8 +45,24 @@ public:
...
-		BCField<3, FieldValue<3>::Vector> bc_conc;
+		enum Concentration_bc_types {
+			bc_inflow,
+			bc_dirichlet,
+			bc_total_flux,
+			bc_diffusive_flux
+		};
+
+		/// Convert bc_type in model to bc_type in abstract setting.
+		static const std::map<unsigned int,unsigned int> bc_type_conversion;
```
jbrezmorf: I think we can remove this map in order to don't be tempted to use it 
in get_bc_type. I prefer to use an lookup vector instead .

jbrezmorf:
Remove map and merge yourself.

PR #424, opened 2015-09-07
JHy ist formatter

x3mSpeedy:
Python update, bug fixes …
 - fix markdown bug when config call (due to markdown library version difference on test server and production server)
 - replace Exceptions with NotImplementedExceptions in same cases
 - add traceback prints when error occures
 - fix html bug when displaying one html section object

Update html-css-js library location from debug to production
 - zip ball is now downloaded directly from github server [repository](https://github.com/x3mSpeedy/Flow123d-htmldoc) using [this](https://github.com/x3mSpeedy/Flow123d-htmldoc/archive/master.zip) link so newest version is always downloaded
 - change path slightly (add /src/) ensuring github repo structure


CMake/Modules/InstallHTMLDOC.cmake
```diff
@@ -4,7 +4,7 @@
 ## HTMLDOC_INSTALL_DIR - directory where helper files will be stored such as CMakeLists.cmake file makefile
 ## HTMLDOC_SOURCE_DIR - direcotry which will contain unpacked files from server
 
-set(HTMLDOC_INSTALL_URL \"http://hybs.nti.tul.cz/test/ist.zip\")
+set(HTMLDOC_INSTALL_URL \"https://github.com/x3mSpeedy/Flow123d-htmldoc/archive/master.zip\")
```
jbrezmorf: Move this to flow.nti.tul.cz

PR #429, opened 2015-09-18
Jb switch bc

jbrezmorf:
Remove dependency of transport sources on darcy flow. 
Simplify dependencies of hc_explicit_coupling by introduction of abstract interfaces  in separate files.
Fix and simplify factories for transport and flow, unify transport constructors.

Preliminary documentation of River boundary condition.


jbrezmorf:
Won't merge.

Issue #412, opened 2015-08-25
TransportDG under TOS


jbrezmorf:
- [x] Change IST: 
	key 'secondary_equation' is Abstract with records: HeatTransfer and TransportOperatorSplitting
	TOS have key 'transport' of Abstract with records ConvectiveTransport, TranportDG
	and key 'reaction_term'
- [x] Make necessary changes in implementation that allows running TransportDG under TOS (coupled with reactions)

jbrezmorf:
merged in 4bba01d1b

PR #433, opened 2015-09-22
Js trans dg under tos

jstebel:
For review.


src/coupling/hc_explicit_sequential.cc
```diff
@@ -29,13 +29,8 @@
...
 
 #include "hc_explicit_sequential.hh"
-#include "flow/darcy_flow_interface.hh"
-#include "transport/advection_process_base.hh"
-// TODO:
-// After having general default values:
-// make TransportNoting default for AdvectionProcessBase abstract
-// use default "{}" for secondary equation.
-// Then we can remove following include.
+#include "flow/darcy_flow_mh.hh"
```
jbrezmorf: Avoid these includes. Thats why we have darcy_flow_interface.hh.

src/coupling/hc_explicit_sequential.cc
```diff
@@ -44,7 +39,8 @@
 
 
 FLOW123D_FORCE_LINK_IN_PARENT(transportOperatorSplitting);
-FLOW123D_FORCE_LINK_IN_PARENT(soluteTransport);
+FLOW123D_FORCE_LINK_IN_PARENT(soluteTransportDG);
```
jbrezmorf: These macros force  correct linking of object files. Therefore we should use rather filename-like identifiers.

src/coupling/hc_explicit_sequential.cc
```diff
@@ -117,14 +113,15 @@ HC_ExplicitSequential::HC_ExplicitSequential(Input::Record in_record)
     // TODO: optionally setup transport objects
     Iterator<AbstractRecord> it = in_record.find<AbstractRecord>("secondary_equation");
     if (it) {
-    	transport_reaction = (*it).factory< AdvectionProcessBase, Mesh &, const Input::Record>(*mesh, *it);
+    	secondary_eq = (*it).factory< AdvectionProcessBase, Mesh &, const Input::Record >(*mesh, *it);
```
jbrezmorf: In fact we should rather rename the class to CouplingTHC, and rename the keys for coupled equations, e.g to:
flow,  solute, heat. We have to postpone general coupling schemes till we solve passing the velocity through fields only.

jstebel: In my opinion, it would take little effort to enable THC simulation with coupling H->C and H->T. But I would postpone it to a separate issue.

jstebel:
Comments and questions:
* Should the TYPEs of transport under TOS be named "ConvectionTransport" and "SoluteTransport_DG" as before or use some new notation? In the description of #412 it is suggested "ConvectiveTransport" and "TransportDG". I would suggest that the names should indicate either the model complexity or the numerical method. E.g.:
  * Convection vs. ConvectionDiffusion
  * Transport_ExplicitFV vs. Transport_ImplicitDG
* Passing concentrations to reaction term: In the proposal, DG computes cell averages and after reaction::update_solution() corrects the concentration DOFs by adding the source due to reactions. This makes the DG solution more discontinuous. I have tried more sofisticated weighted correction but there was some problem with oscillations near boundary.
* Some reactions need nonnegative concentrations. This is a pain and we need to introduce some flux limiter or at least enforce nonnegative values of the solution while preserving mass balance.

jbrezmorf:
Concerning TYPE: 
Since  'multiple inheritance' of Abstract is now allowed, we should have names that characterise the equation in different contexts. E.g.
DG transport equation may be used directly in HCSequantialCoupling or as part of TransportOperatorSplitting. So I propose to use names composed from
name of the process followed by some shortcut of the numerical method, e.g. 

DarcyFlow_MH
Richards_MH
Transport_OS or only OpeartorSplitting if this would became a general purpose coupling class.
Convection_FV
ConvectionDiffusion_DG
HeatTransfer_DG

any further refinements of this concept are welcome. E.g. how reaction names fit in this scheme?

jbrezmorf:
Concerning the second problem. Need more details about its nature. Shell discuss personally at Thursday.
Third problem: agree to do some research about limiters and postprocessing to deal with negative values.

jstebel:
Remains to fix test 20, otherwise can be merged. Resolves #412

src/coupling/balance.cc
```diff
@@ -156,16 +139,35 @@ std::vector<unsigned int> Balance::add_quantities(const std::vector<string> &nam
 }
 
 
-void Balance::allocate(unsigned int n_loc_dofs,
+void Balance::allocate(const Distribution *el_ds,
```
jbrezmorf: Can we simplify the interface of this method after I finally merged Balance cleanup pull?

src/system/sys_profiler.cc
```diff
@@ -350,7 +350,43 @@ std::string common_prefix( std::string a, std::string b ) {
     return std::string( a.begin(), std::mismatch( a.begin(), a.end(), b.begin() ).first ) ;
 }
 
+void Profiler::print_timer_path (int timer_index) {
```
jbrezmorf: Is the change in profiler necessary for this pull? It should rather be part of separate pull.
However, it looks like forgotten debugging hack.

src/transport/heat_model.cc
```diff
@@ -238,9 +234,26 @@ IT::Selection HeatTransferModel::ModelEqData::get_output_selection_input_type(co
 }
 
 
-HeatTransferModel::HeatTransferModel() :
-		flux_changed(true)
-{}
+HeatTransferModel::HeatTransferModel(Mesh &mesh, const Input::Record in_rec) :
```
jbrezmorf: Why this constructor do principally more tasks then the constructor of the ConcentrationTransport model?

src/transport/transport.h
```diff
@@ -92,15 +92,22 @@ class ConvectionTransport : public TransportBase {
...
      * Constructor.
      */
-        ConvectionTransport(Mesh &init_mesh, const Input::Record in_rec);
+    ConvectionTransport(Mesh &init_mesh, const Input::Record in_rec);
 	/**
 	 * TODO: destructor
 	 */
 	virtual ~ConvectionTransport();
 
+	void initialize() override;
```
jbrezmorf: Is it really necessary to introduce other initialization method then constructor and zero_time_step?

src/transport/transport.h
```diff
@@ -108,12 +115,16 @@ class ConvectionTransport : public TransportBase {
     /**
      * 
      */
-    bool assess_time_constraint(double &time_constraint);
+    bool assess_time_constraint(double &time_constraint) override;
```
jbrezmorf: Please, ask PE for comment on this method.

jbrezmorf:
We can not change any CON file. The workflow should be:
- make chenge of the input structure in the code
- modify transformation rules to cover the change, see documentation:
http://geomop.github.io/GeoMop/aut/ModelEditor/data.yaml.html#module-data.yaml.transformator
- regenerate YAML files from CON files
- check if the result works, possibly fix transformation rules

src/transport/heat_model.cc
jstebel: HeatTransfer defines its own vector of substances. In addition, HeatTransfer is a secondary equation, while ConcentrationTransport is under operator splitting.

src/transport/transport.h
jstebel: The instance of ConcentrationTransportBase (CTB) is constructed in 3 steps:
1) TransportOperatorSplitting (TOS) calls the constructor
2) TOS initializes several members of CTB from its input record
3) CTB::initialize() finalizes the construction.
If the constructor takes an additional argument, the input record of TOS, then it may be possible to eliminate initialize().

src/coupling/balance.cc
jstebel: OK

src/system/sys_profiler.cc
jstebel: OK, reverted.

jstebel:
I have tried to fix all commented issues except for changes in CON files. This will be done in a separate branch and pull.

PR #434, opened 2015-09-24
Df multiple inheritance

jbrezmorf:
Implements multiple inheritance of Records (from Abstracts).


src/input/accessors.cc
```diff
@@ -142,8 +142,8 @@ std::string Address::make_full_address() const {
         } else
 		if (typeid(*input_type) == typeid(Type::AbstractRecord)) {
 			const Type::AbstractRecord * a_rec = static_cast<const Type::AbstractRecord *>(input_type);
-			const StorageInt * storage_type = static_cast<const StorageInt *>(storage->get_item(0));
-			input_type = & a_rec->get_descendant((unsigned int)storage_type->get_int());
+			const StorageString * storage_type = static_cast<const StorageString *>(storage->get_item(0));
```
jbrezmorf: Try to call type() method here.

src/input/storage_transpose.cc
```diff
@@ -112,7 +112,8 @@ StorageBase * StorageTranspose::modify_storage(const Type::TypeBase *target_type
 	ASSERT( target_arec != NULL, "Incompatible type of target type. Must be AbstractRecord!\n");
 	ASSERT( typeid(*source_storage) == typeid(StorageArray), "Incompatible type of storage. Must be Array!\n");
 
-	unsigned int descendant_index = (unsigned int)(source_storage->get_item(0)->get_int());
+	string descendant_name = source_storage->get_item(0)->get_string();
```
jbrezmorf: Try to use directly get_descendant(string) 
possibly we can remove Record::get_descendant(int)

jbrezmorf:
Remove commits with massive changes in OutputType

src/main.cc
```diff
@@ -209,23 +209,7 @@ void Application::parse_cmd_line(const int argc, char ** argv) {
     	exit( exit_output );
     }
 
-    // if there is "full_doc" option
```
jbrezmorf: Just comment out. We may support that in future using Python formaters.

unit_tests/input/reader_to_storage_test.cpp
```diff
@@ -678,6 +678,59 @@ TEST_F(InputReaderToStorageTest, AbstractRec) {
...
+			.derive_from(a_rec2)
+			.declare_key("c_val", Type::Double(), Type::Default::obligatory(), "")
+			.close();
+
+	Type::Record root = Type::Record("problem", "Root record")
+			.declare_key("primary", a_rec1, Type::Default::obligatory(), "")
+			.declare_key("secondary", a_rec2, Type::Default::obligatory(), "")
+			.close();
+
+    {   // Try correct type
```
jbrezmorf: Check that we have test of errors in type name on input.

src/input/accessors.cc
dflanderka: type() is method of accessor Input::AbstractRecord, but in this case we have Input::Type::AbstractRecord

jbrezmorf: You are right. The address concept is completely rotten anyway we will work on it later.

jbrezmorf:
Merged in branch DF_multiple_inheritance_merge.

PR #437, opened 2015-09-30
Df ist output

dflanderka:


src/input/type_output.cc
```diff
@@ -3,6 +3,8 @@
  */
 
 #include "input/type_output.hh"
+#include "input/type_repository.hh"
+#include "rev_num.h"
```
jbrezmorf: This include is forbidden for every  source else then main.cc. 
All information have to be passed through appropriate method of OutputJSON.

src/input/type_output.hh
```diff
@@ -11,6 +11,7 @@
 
 #include "input/type_base.hh"
 #include "input/type_record.hh"
+#include "input/type_generic.hh"
 #include <boost/regex.hpp>
```
jbrezmorf: Can we remove this include?

unit_tests/input/type_output_test.cpp
```diff
@@ -92,24 +92,10 @@ TEST(OutputTypeTypeBase, record_output_test) {
...
-    cout << OutputLatex(&main) << endl;
+    cout << OutputText(&main) << endl << endl;
 
     cout << "## " << "OutputJSONMachine printout" << endl;
-
-    cout << OutputJSONMachine(&main) << endl;
+    cout << OutputJSONMachine() << endl;
 }
 
 /*TEST(OutputTypeAbstractRecord, abstract_record_test) {
```
jbrezmorf: Why these tests are commented. Try to uncomment and possibly fix them.

src/input/type_output.hh
dflanderka: No, we use boost::regex in inner class OutputJSONMachine::RewriteRule

jbrezmorf: RewriteRule is used only in escape_description and is mere pair. So, we can remove class RewriteRule and use 
static vector of pairs in the method OutputJSONMachine::escape_description.

src/main.cc
```diff
@@ -112,23 +112,32 @@ void Application::split_path(const string& path, string& directory, string& file
     }
 }
 
+
+std::map<string, string> Application::get_rev_num_data() {
```
jbrezmorf: Why using map where we can use a struct?
Declare appropriate struct in type_output.hh.

Issue #339, opened 2015-03-26
Jump times


jbrezmorf:
Issue: We can not use data from the actual time step in the explicit transport
  and in explicit solvers in general because of the CFL condition.
So we turn back to classical explicit scheme.

The problem is how to achieve same behaviour of implici and explicit schemes at times with data discontinuity. The problem is mainly on output where one may want to see the response of the solution at both time sides in particular if this involves jump in the solution (usualy caused by jump in the data connecting conservative quantity and principal variable.

One possiblility is to compute solution (and possibly perform output) at both sides of the jump time. This means to compute the solution at that time with left limit data, then apply right limit data to conserved quantity to get right limit solution. 

TODO:
- [x] change convective transport into conservative form
-  ~~introduce field flag for the data in the time term~~
- ~~implicitly treat all changes in fields as continuous, but adapt the time step to them~~
- ~~allow to specify some field changes as continuous and do not adapt the time step to them~~
- ~~Discuss previous with users.~~
- ~~introduce an input key in the field descriptor to mark substantial jump in the data~~
- ~~apply proposed scheme if any field of the time term have a jump time, 
  propose general mechanism  applicable to any implicit or explicit method.~~
- ~~allow specify precise output of the jump times at output~~

jbrezmorf:
Replaced by #694

Issue #338, opened 2015-03-23
Transport CFL condition for sourcs


jbrezmorf:
- Implement combined Neuman and Newton BC in ConvectiveTransport, see #367
- Constrain time step in convective transport by CFL condition for sources.
- Do we need some form of CFL condition in reactions and/or dual porosity?
- Add descriptions to timestep constrains and report description of active constrains in the log.
- Document all possible CFL conditions that influence the timestep.

Paulie14:
Notes:
Divide CFL computation in explicit transport from assembly.
Then suggest passing CFL condition among equations.

Paulie14:
prerequisite: 
implement source term in the form "s(cs - c[t])" not with difference"s(cs-c[t-1])" 
This way it contributes into diagonal of the trasnport matrix (saved in separate vector)
this is done by commit ffedcd03f3563bd2df298a3eab6c85d865b5b790

Paulie14:
How to compute CFL condition?
condition is in the form timestep < 1 / (matrix_part + source_part),
where matrix_part is due to flow and is assembled in matrix assembly,
on the other hand source_part is assembled separately with source term assembly (because matrix assembly is independent of substances data)

To compute the condition correctly, I would need to save matrix_part in all elements in a parallel vector and source_part in another one. Then add the vectors and find its maximum.
Is this the correct way?

jbrezmorf:
Yep. Petsc matrix have function MatGetDiagonal to this end. Soon we should also move to deal.ii linear algebra interface.

jbrezmorf:
Is this completed or something remain to finish?

Paulie14:
- I didn't do anything with BC
- constraint by sources is done
- I didn't do anything about the constraints from reactions yet
- description of constraints are now being discussed in pull request
- I didn't write about CFL in documentation yet

jbrezmorf:
Time step constraint messages implemented in pull: #493
Reactions postponed to issue #552.
Documentation postponed to issue #553

PR #438, opened 2015-09-30
Pe cfl

Paulie14:
- implement sources correctly by the explicit scheme - addition into transport matrix diagonal
- divide assembly and update solution in convection transport
- cfl condition is now limited by sources
- move next_time() at the beginning of update_solution in convection
- add test in test16 for sources cfl condition
- correct test20 with sources, where sources cfl condition was not satisfied

- partially solves #338 and #339


src/transport/transport.cc
```diff
@@ -168,30 +169,41 @@ ConvectionTransport::~ConvectionTransport()
...
     //Destroy mpi vectors at first
-    VecDestroy(&v_sources_corr);
     MatDestroy(&tm);
-    
-    VecDestroy(vconc);
-    VecDestroy(bcvcorr);
-    VecDestroy(vpconc);
-    VecDestroy(vcumulative_corr);
-    
+    VecDestroy(&vcfl_flow_);
```
jbrezmorf: Try to use VectorSeqDouble in fields/vec_seq_double.hh for all internal vectors namely if both PETSc vector and raw ptr are necessary.

src/transport/transport.h
```diff
@@ -119,6 +119,10 @@ class ConvectionTransport : public TransportBase {
 	 * Initialize solution at zero time.
 	 */
     void zero_time_step() override;
+    /**
+     * 
+     */
+    bool assess_time_constraint(double &time_constraint);
```
jbrezmorf: Dosumentation. 'Assess' have misleading meaning in this context I propose to use 'evaluate' instead.

src/transport/transport.h
```diff
@@ -224,33 +231,37 @@ class ConvectionTransport : public TransportBase {
...
-           **sources_conc,
-           **sources_sigma;
 
     TimeMark::Type target_mark_type;    ///< TimeMark type for time marks denoting end of every time interval where transport matrix remains constant.
-    double cfl_max_step;
-            // only local part
-
+    double cfl_max_step;    ///< Time step constraint coming from CFL condition.
+    
+    Vec vcfl_flow_,     ///< Parallel vector for flow contribution to CFL condition.
```
jbrezmorf: Can we create these CFL related vectors temporary in asses_time_constraint?
Why there is recalculation of cfl_flow_ in convection_matrix_mpi?

src/transport/transport.cc
```diff
@@ -466,87 +480,156 @@ void ConvectionTransport::zero_time_step()
...
+    data_.set_time(time_->step()); // set to the last computed time
+    
+    START_TIMER("data reinit");
+    
+    bool cfl_changed = false;
+    
+    // if FLOW or DATA changed ---------------------> recompute transport matrix
+    if (mh_dh->time_changed() > transport_matrix_time  || data_.porosity.changed())
+    {
         create_transport_matrix_mpi();
```
jbrezmorf: This means that we do matrix assembly two times. 
What about splitting create_transport matrix_mpi into smaller functions (possibly inlined) for different tasks on single element. Then use only necessary parts to get CFL condition (matrix diagonal).

src/transport/transport.cc
```diff
@@ -466,87 +480,156 @@ void ConvectionTransport::zero_time_step()
...
+        for (unsigned int sbi=0; sbi<n_subst_; sbi++)
+        {
+            VecScale(v_sources_corr[sbi], dt_scaled);
+            VecScale(v_tm_diag[sbi], dt_scaled);
+        }
+    }
+    
+    // if DATA or TIME STEP changed -----------------------> rescale transport matrix
+    scaled = false; //reset; flag to avoid rescaling newly created tm term
+    if ( !is_convection_matrix_scaled ) { // scale fresh convection term matrix
```
jbrezmorf: Is there still reason to not perform scaling at matrix assembly? Since the reason was that we did not know time step at assembly, but now we should know it. Same for BC and sources above.

src/transport/transport.cc
```diff
@@ -738,58 +814,21 @@ int *ConvectionTransport::get_row_4_el(){
...
-		{
-			if (conc[loc_el] < sources_conc[sbi][loc_el])
-				conc_diff[loc_el] = sources_conc[sbi][loc_el] - conc[loc_el];
-			else
-				conc_diff[loc_el] = 0;
-		}
-
 		balance_->calculate_mass(sbi, vconc[sbi]);
-		balance_->calculate_source(sbi, vconc_diff);
+		balance_->calculate_source(sbi, vconc[sbi]);
```
jbrezmorf: Are the balance matrices for source term set?

jbrezmorf:
Since JS want to continue on this work:
1) make a branch from current master to possibly repeat this diff
2) merge this branch into master manually in your repository and push the master (should leave this pull request open)
3) continue on proposed improvements if they are possible, indeed.

src/transport/transport.h
Paulie14: This is due to the fact, that cfl is computed in two different parts of assembly - 'flow' and sources.
If e.g. only sources change, we update only vcfl_sources and use vcfl_flow from previous time step. If these vectors are only temporal, then I will need to recompute cfl from flow and sources every time any data or flow is changed. 

Cfl condition is evaluated while assembling matrix and sources. If we separate it, we  will do lots of computation again, I think.

Current approach is:
- (if data, or flow changed) assembly matrix and sources while the CFL is evaluated
- time step is selected
- matrix is scaled

From your other comments, I understand following approach:
- splitting create_transport matrix_mpi into smaller functions
- compute CFL independently, ahead of assembly (-> some data and fluxes would be evaluated twice)
- then use only temporal vectors for cfl computation (-> save memory, possibly increase time by recomputation of cfl terms)
- then in TOS we would evaluate CFL (without assembly), then choose timestep and then assembly matrix
- then it would be possible to assemble and scale the matrix at once

src/transport/transport.cc
Paulie14: We do not do assembly twice. See comment above.

src/transport/transport.cc
Paulie14: The sequence is: assembly (+ cfl evaluation) -> select time step (in future possibly changed by reactions) -> scale matrix.
See above.

src/transport/transport.cc
Paulie14: I believe so. It is set in compute_concentration_sources(), transport.cc: 386.

PR #446, opened 2015-10-14
Pe time constraints

Paulie14:
Suggestion for managing and reporting time constraints of equations.
Also including some minor improvements in explicit transport suggested in PE_cfl pull request.


jbrezmorf:
Why we need to save all constraints? It should be enough to save name of the active lower and upper constraint.

Paulie14:
Because:
- the constraint list contains now also the permanent min and max constraints
- when defining new constraint, we set the message; when we repeatedly set the value of constraint, the message is already there... (the constraint can be set from different places and the message would be badly managable)

jbrezmorf:
- permanent constraint may be set just after setting new time
- Is the message helpful if it is not connected to particular setting of the constraint? Two distinguish calls of set_constrains should be two distinguished 
constraints with two distinguished messages.

PR #449, opened 2015-10-20
Df general default values

jbrezmorf:


src/input/reader_to_storage.cc
```diff
@@ -472,80 +477,12 @@ StorageBase * ReaderToStorage::make_storage(PathBase &p, const Type::String *str
...
     try {
-    	/*
-    	// Possible simplification of this method (need default strings to be valid JSON)
-    	ReaderToStorage  tmp_storage(dflt_str, *type);
+    	// default strings must be valid JSON
+    	//ReaderToStorage  tmp_storage(dflt_str, *type, reader_file_format_);
+    	istringstream is(dflt_str);
+    	ReaderToStorage tmp_storage;
+    	tmp_storage.read_stream(is, *type, reader_file_format_);
     	return tmp_storage.storage_;
```
jbrezmorf: Use always JSON parser for default strings.

src/input/type_record.cc
```diff
@@ -214,17 +232,12 @@ bool Record::is_closed() const {
 
 bool Record::check_key_default_value(const Default &dflt, const TypeBase &type, const string & k_name) const
```
jbrezmorf: Is this method still necessary? Possibly remove.

src/input/type_record.cc
```diff
@@ -62,6 +63,23 @@ TypeBase::TypeHash Default::content_hash() const
...
+		reader.read_stream(is, type, FileFormat::format_JSON);
+		return true;
+	} catch ( Input::ReaderToStorage::ExcNotJSONFormat &e ) {
+		THROW( ExcWrongDefault() << EI_DefaultStr( value_ ) << EI_TypeName(type.type_name()));
+	} catch ( Input::ReaderToStorage::ExcInputError &e ) {
+		THROW( ExcWrongDefault() << EI_DefaultStr( value_ ) << EI_TypeName(type.type_name()));
+	}
+}
+
+
```
jbrezmorf: Add a pointer to storage into Default class.
Add method get_storage() (automatically calls check_validity if the storage is NULL)
Use the storage in ReaderToStorage and in Abstract::get_default_descendant.

src/transport/transport_dg.cc
```diff
@@ -91,12 +91,12 @@ const Record & TransportDG<Model>::get_input_type() {
...
 		.declare_key("input_fields", Array(TransportDG<Model>::EqData().make_field_descriptor_type(std::string(Model::ModelEqData::name()) + "_DG")), IT::Default::obligatory(), "")
-		.declare_key("dg_variant", TransportDG<Model>::get_dg_variant_selection_input_type(), Default("non-symmetric"),
+		.declare_key("dg_variant", TransportDG<Model>::get_dg_variant_selection_input_type(), Default("\"non-symmetric\""),
 				"Variant of interior penalty discontinuous Galerkin method.")
 		.declare_key("dg_order", Integer(0,3), Default("1"),
 				"Polynomial order for finite element in DG method (order 0 is suitable if there is no diffusion/dispersion).")
 		.declare_key("output_fields", Array(EqData::get_output_selection()),
-				Default(Model::ModelEqData::default_output_field()),
+				Default("\""+Model::ModelEqData::default_output_field()+"\""),
 				"List of fields to write to output file.")
```
jbrezmorf: Move \" into model method.

jbrezmorf:
Add unit test for JSON parser JSON_spirit (like YAML_lib.cpp) test how we can read default values. Possibly wrapped in brackets: [ <value> ].

src/input/path_json.cc
```diff
@@ -25,8 +25,11 @@ PathJSON::PathJSON(istream &in)
 {
     io::filtering_istream filter_in;
 
-    filter_in.push(uncommenting_filter());
-    filter_in.push(in);
+	std::stringstream wrapped_stream;
+	wrapped_stream << "[\n" << in.rdbuf() << "\n]";
+
+	filter_in.push(uncommenting_filter());
+    filter_in.push(wrapped_stream);
```
jbrezmorf: This wrapping should be made only when using JSON reader for reading default values. Using it here for any input stream breaks correctness of reported error paths for the user provided input file.

src/input/reader_to_storage.cc
```diff
@@ -58,6 +62,9 @@ ReaderToStorage::ReaderToStorage(const FilePath &in_file, const Type::TypeBase &
 ReaderToStorage::ReaderToStorage( const string &str, const Type::TypeBase &root_type, FileFormat format)
 : ReaderToStorage()
 {
+	// finish all lazy input types
+    Input::Type::TypeBase::lazy_finish();
```
jbrezmorf: Move into try block.

src/input/type_record.cc
```diff
@@ -247,7 +256,12 @@ bool Record::finish(bool is_generic)
 
             // we check once more even keys that was already checked, otherwise we have to store
             // result of validity check in every key
```
jbrezmorf: Is the comment still relevant?

src/input/type_record.cc
```diff
@@ -415,7 +416,12 @@ void Record::RecordData::declare_key(const string &key,
                          const Default &default_value, const string &description)
 {
     ASSERT(!closed_, "Can not add key '%s' into closed record '%s'.\n", key.c_str(), type_name_.c_str());
-
+    try {
```
jbrezmorf: Do we need to test validity of default value both here and in finish?

Issue #386, opened 2015-06-24
Review Mesh setting


jbrezmorf:
- [ ] just one array "regions" Array of Region
- [ ] key "print_regions" - print table of all bulk and boundary regions after all operations on stdout,
      columns: name, specification either list of subregions or label and ID in the mesh

- [ ] Region is IT::abstract with childs:
  - From_ID { name, id <obligatory>}
  - From_Label ```{ name, mesh_label <obligatory>}``` mesh_label is e.g. physical volume name in GMSH format, Regions with name=mesh_label are created automatically.
  - From_Elements ```{ name, id <optional>, element_list = list of element IDs}```
  - Union { name,  region_ids, regions }
  - Difference {name,  regions=[Array[2,2] }
  - Intersection { name, regions }
- [x] Need to move reading of region table after setup_topology in init_from_input. Should not be problem in principle, but some technicalities may appear.
- [ ] Improve documentation.
- [x] Add auto conversion of the mesh from its key "file".
- [ ] Test and possibly fix transformation file: ```src/python/GeoMop/ModelEditor/resources/transformation/main.json```, transformation of tests is done by scripts bin/test/transform_*

depends on #456

PR #453, opened 2015-10-30
Df generalized region

dflanderka:


src/fields/field.impl.hh
```diff
@@ -328,12 +328,17 @@ void Field<spacedim,Value>::update_history(const TimeStep &time) {
 				}
 
 			} else if (shared_->list_it_->opt_val("region", domain_name)) {
+				domain = mesh()->region_db().get_region_set(domain_name);
```
jbrezmorf: Should remove 'r_set' key.

src/mesh/elements.h
```diff
@@ -126,6 +126,7 @@ class Element
     unsigned int dim_;
 
     friend class GmshMeshReader;
+    friend class Mesh;
```
jbrezmorf: Why?

src/mesh/mesh.cc
```diff
@@ -219,15 +221,18 @@ void Mesh::init_from_input() {
     Input::Array region_list;
     RegionDB::MapElementIDToRegionID el_to_reg_map;
 
-    // create regions from our input
-    if (in_record_.opt_val("regions", region_list)) {
-        region_db_.read_regions_from_input(region_list, el_to_reg_map);
-    }
     // read raw mesh, add regions from GMSH file
     GmshMeshReader reader( in_record_.val<FilePath>("mesh_file") );
     reader.read_mesh(this, &el_to_reg_map);
```
jbrezmorf: So, now we can remove el_to_reg_map parameter and appropriate code form read_mesh.

src/mesh/region.cc
```diff
@@ -146,6 +146,12 @@ Region RegionDB::add_region(unsigned int id, const std::string &label) {
         THROW(ExcNonuniqueLabel() << EI_Label(label) << EI_ID(id) << EI_IDOfOtherLabel(it_label->get_id()) );
     }
 
+    OnlyIDIter it_only_id = region_set_.get<OnlyID>().find(id);
```
jbrezmorf: Why we need this if regions for region sets are added in close()?
However in order to simplify operations for creating region sets (union, etc.) we should
create region sets for regions just after reading the mesh.

Possibly we can just rename this method to rename_region, corresponding to the operation to name the region.

src/mesh/region.cc
```diff
@@ -234,14 +240,19 @@ void RegionDB::close() {
...
    for(unsigned int i=0; i< size(); i++) {
        Region reg(i, *this);
+   	   RegionSet region_set;
+   	   region_set.push_back( reg );
+
 
        if (reg.is_boundary() && (reg.boundary_idx() < boundary_size()) ) {
            add_to_set("BOUNDARY", reg );
            add_to_set("ALL", reg);
+       	   add_set(reg.label(), region_set);
```
jbrezmorf: Possibly we need to add regions into sets earlier.

src/mesh/region.cc
```diff
@@ -535,6 +546,22 @@ Region RegionDB::replace_region_dim(DimIDIter it_undef_dim, unsigned int dim, bo
...
+	unsigned int index = it_only_id->index;
+
+	RegionItem item(index, it_only_id->get_id(), label, it_only_id->dim());
+	region_set_.replace(
+            region_set_.get<Index>().find(index),
+            item);
+
+	Region r_id=Region(index, *this);
+    // check boundary
+    if ( r_id.is_boundary() != boundary )
```
jbrezmorf: Is it necessary to check the boundary at this level?

jbrezmorf:
- This pull contains also changes in initialization of region DB which should be rather part of other pull for the mesh input changes.
- Missing remove of r_set key.
- Missing appropriate transformation file and changes in yaml files.

src/mesh/mesh.cc
dflanderka: We can't remove this code, it relates with key 'element_list', that is not removed or changed (maybe some changes will be in next issue #386). Marked 'if (in_record_.opt_val("regions" ...' is not removed, only moved after call setup_topology.

src/mesh/region.cc
dflanderka: region_set_ is RegionTable (database of all region) - variable name is misleading.
I propose to rename region_set_ (e.g. to region_table_).

Method is renamed.

src/mesh/region.cc
dflanderka: Problem with earlier added regions is with regions renamed by user. In this case we would must rename appropriate set too.

src/mesh/mesh.cc
```diff
@@ -238,6 +243,20 @@ void Mesh::init_from_input() {
 
 
 
+void Mesh::modify_element_ids(const RegionDB::MapElementIDToRegionID *map) {
+	if (map) {
+		FOR_ELEMENTS(this, ele) {
```
jbrezmorf: Try to use element.find_id(id) to loop only over the map.

PR #459, opened 2015-11-04
Fix make dependencies

jbrezmorf:
- fix wrong dependencies in the main makefile


jbrezmorf:
Part merged with JB_make_dependencies branch. Won't make big changes since the thing is more complex then it appears.

PR #460, opened 2015-11-06
Js transformation con yaml

### LINKED ISSUE #456: Prepare tests for creating transformation rules
- [x] Add transformation script into repository.
- [x] Create (empty) main file with transformation rules.
- [x] into ```bin/``` add simple python script that use the main transformation file.
- [x] Make a script that apply transformation to all test con files. 
- [x] Apply this script with empty transformation rules. Fix some CON files with complex references.
- [x] change the test scripts and change ref_output subdirs to not contain suffix.
- [x] Make all tests run on YAML files.

jstebel:
Resolves #456 (almost).
The GeoMop files are copied statically. If submodule is preferred, I can switch to it after one small fix is done in GeoMop repository.


jbrezmorf:
Submodule not necessary. Some issue in test 10. Otherwise very good. Thanks.

jstebel:
I hope the problem with test 10 was exceptional. I cannot reproduce it.

PR #475, opened 2015-12-02
Df array transposition

dflanderka:


src/input/reader_to_storage.cc
```diff
@@ -347,17 +347,73 @@ StorageBase * ReaderToStorage::make_storage(PathBase &p, const Type::Array *arra
...
+				StorageArray *storage_array = new StorageArray(1);
+				const Type::TypeBase &sub_type = array->get_sub_type();
+				storage_array->new_item(0, make_storage(p, &sub_type) );
+
+				return storage_array;
+			} else {
+				THROW( ExcInputError() << EI_Specification("Automatic conversion to array not allowed. The value should be '" + p.get_node_type(ValueTypes::array_type) + "', but we found: ")
+						<< EI_ErrorAddress(p.as_string()) << EI_JSON_Type( p.get_node_type(p.get_node_type_index()) ) << EI_InputType(array->desc()) );
+			}
         }
```
jbrezmorf: add else here to improve readibility

src/input/reader_to_storage.cc
```diff
@@ -347,17 +347,73 @@ StorageBase * ReaderToStorage::make_storage(PathBase &p, const Type::Array *arra
...
+
+    	// set variables managed transposition
+        try_transpose_read_ = true;
+    	transpose_index_ = 0;
+    	transpose_array_sizes_.clear();
+
+    	const Type::TypeBase &sub_type = array->get_sub_type();
+    	StorageBase *tmp_storage = make_storage(p, &sub_type);
```
jbrezmorf: Here we should catch thrown exceptions, form their messages and throw new exception with message:
Failed transpose auto-conversion of the intput EI_InputType to the type EI_RequiredType. 
Address: EI_Address
Transpose filed due to: <original message>
Address: EI_Address

src/input/reader_to_storage.cc
```diff
@@ -347,17 +347,73 @@ StorageBase * ReaderToStorage::make_storage(PathBase &p, const Type::Array *arra
...
+    	if (try_transpose_read_) {
+			// try automatic conversion to array with one element
+			if ( array->match_size( 1 ) ) {
+				StorageArray *storage_array = new StorageArray(1);
+				const Type::TypeBase &sub_type = array->get_sub_type();
+				storage_array->new_item(0, make_storage(p, &sub_type) );
+
+				return storage_array;
+			} else {
+				THROW( ExcInputError() << EI_Specification("Automatic conversion to array not allowed. The value should be '" + p.get_node_type(ValueTypes::array_type) + "', but we found: ")
```
jbrezmorf: Adapt the message to the new context, e.g.
During transpose auto-conversion, the conversion to the single element array not allowed. Require type:
... \nFound on input: ...\n

src/input/reader_to_storage.cc
```diff
@@ -347,17 +347,73 @@ StorageBase * ReaderToStorage::make_storage(PathBase &p, const Type::Array *arra
...
+        try_transpose_read_ = true;
+    	transpose_index_ = 0;
+    	transpose_array_sizes_.clear();
+
+    	const Type::TypeBase &sub_type = array->get_sub_type();
+    	StorageBase *tmp_storage = make_storage(p, &sub_type);
+
+    	// automatic conversion to array with one element
+    	if (transpose_array_sizes_.size() == 0) {
+			if ( array->match_size( 1 ) ) {
```
jbrezmorf: Make a method to this to avoid code duplicity.

src/input/reader_to_storage.cc
```diff
@@ -347,17 +347,73 @@ StorageBase * ReaderToStorage::make_storage(PathBase &p, const Type::Array *arra
...
+				StorageArray *storage_array = new StorageArray(1);
+				storage_array->new_item(0, tmp_storage);
+				try_transpose_read_ = false;
+
+				return storage_array;
+			} else {
+				THROW( ExcInputError() << EI_Specification("Automatic conversion to array not allowed. The value should be '" + p.get_node_type(ValueTypes::array_type) + "', but we found: ")
+						<< EI_ErrorAddress(p.as_string()) << EI_JSON_Type( p.get_node_type(p.get_node_type_index()) ) << EI_InputType(array->desc()) );
+			}
+    	}
```
jbrezmorf: Add else.

src/input/reader_to_storage.cc
```diff
@@ -347,17 +347,73 @@ StorageBase * ReaderToStorage::make_storage(PathBase &p, const Type::Array *arra
...
+
+    	// check sizes of arrays stored in transpose_array_sizes_
+    	transpose_array_sizes_.erase( unique( transpose_array_sizes_.begin(), transpose_array_sizes_.end() ),
+    								  transpose_array_sizes_.end() );
+    	if (transpose_array_sizes_.size() == 1) {
+    		unsigned int sizes = transpose_array_sizes_[0]; // sizes of transposed
+
+    		// array size out of bounds
+    		if ( !array->match_size( sizes ) ) {
+    			THROW( ExcInputError() << EI_Specification("Do not fit into size limits of the Array.")
```
jbrezmorf: Improve the message:
"Result of transpose auto-conversion do not fit the size of the Array:\n
EI_InputType\n
Input size: EI_InputArraySize"

src/input/reader_to_storage.cc
jbrezmorf: Rename tmp_storage to first_item_storage

src/input/reader_to_storage.cc
```diff
@@ -347,17 +347,73 @@ StorageBase * ReaderToStorage::make_storage(PathBase &p, const Type::Array *arra
...
+    					<< EI_ErrorAddress(p.as_string()) << EI_InputType(array->desc()) );
+    		}
+
+			// create storage of array
+    		StorageArray *storage_array = new StorageArray(sizes);
+			storage_array->new_item(0, tmp_storage);
+			if (sizes>1) {
+				++transpose_index_;
+				while (transpose_index_ < sizes) {
+					storage_array->new_item(transpose_index_, make_storage(p, &sub_type));
```
jbrezmorf: Same as with the first call, add catch.

src/input/reader_to_storage.cc
```diff
@@ -347,17 +347,73 @@ StorageBase * ReaderToStorage::make_storage(PathBase &p, const Type::Array *arra
...
+					++transpose_index_;
+				}
+			}
+
+			try_transpose_read_ = false;
+			return storage_array;
+    	} else {
+			THROW( ExcInputError() << EI_Specification("Unequal sizes of sub-arrays during transposition of '" + p.get_node_type(ValueTypes::array_type) + "'")
```
jbrezmorf: replace transposition by 'transpose auto-conversion'

src/input/reader_to_storage.cc
```diff
@@ -373,6 +429,9 @@ StorageBase * ReaderToStorage::make_storage(PathBase &p, const Type::Selection *
 		int value = selection->name_to_int( item_name  );
 		return new StorageInt( value );
 	} catch (ExcInputError & e) {
+		if (try_transpose_read_) {
```
jbrezmorf: Exceptions should be used only for errors. So we have to check for the array on input just before we call get_string_value().

src/input/reader_to_storage.cc
```diff
@@ -502,6 +573,31 @@ StorageBase * ReaderToStorage::make_storage_from_default(const string &dflt_str,
...
+		THROW( ExcInputError() << EI_Specification("Empty array during transposition.")
+			<< EI_ErrorAddress(p.as_string()) << EI_InputType(type->desc()) );
+	} else if ( arr_size > 0 ) {
+		if (transpose_index_ == 0) transpose_array_sizes_.push_back( arr_size );
+		p.down(transpose_index_);
+		StorageBase *storage = make_storage(p, type);
+		p.up();
+		return storage;
+	} else { // arr_size == -1
+		 THROW( ExcInputError()
```
jbrezmorf: If this method is used only for p.is_array_type() true, this case should not happen. Thus we should have an ASSERT here.

unit_tests/input/reader_to_storage_test.cpp
```diff
@@ -377,7 +378,7 @@ TEST_F(InputReaderToStorageTest, Array) {
         EXPECT_EQ(3.2, storage_->get_item(0)->get_double() );
 
         stringstream ss1("{ key=3.2}");
-        EXPECT_THROW_WHAT( {read_stream(ss1, darr_type);}, ExcInputError , "The value should be 'JSON real', but we found:.* 'JSON object'");
+        EXPECT_THROW_WHAT( {read_stream(ss1, darr_type);}, ExcInputError , "The value should be 'JSON array', but we found:.* 'JSON object'");
```
jbrezmorf: The new message is invalid. Why it reports required type 'JSON array'?

src/input/reader_to_storage.cc
```diff
@@ -502,6 +573,31 @@ StorageBase * ReaderToStorage::make_storage_from_default(const string &dflt_str,
 
 
 
+StorageBase * ReaderToStorage::make_transposed_storage(PathBase &p, const Type::TypeBase *type) {
+	ASSERT(try_transpose_read_, "Unset flag try_transpose_read_!\n");
+
+	int arr_size = p.get_array_size();
+	if ( arr_size == 0 ) {
+		THROW( ExcInputError() << EI_Specification("Empty array during transposition.")
```
jbrezmorf: use 'transpose auto-conversion'

PR #477, opened 2015-12-03
JHY - edit transform_profiler_data, now limited to MPI_RANK=0 only

x3mSpeedy:


jbrezmorf:
Is this fix of the issue with test 20? Does it really fix the problem? I wonder since the issue was that JSON file was not created at all.

Anyway this fix real bug, but to prevent other possible issues we should save MPI communicator used during output and use the same in the transform method. Otherwise different communicators may be passed resulting in possible error.

x3mSpeedy:
When calling output, only rank 0 writes a file, since other processors skip file export they can arrive (if file exporting takes long) earlier to python execution part, thus invoking python when there is no such file yet (or is empty).

Regarding to MPI communicator, in ```main.cc``` we are passing the same communicator, or can the variable **```PETSC_COMM_WORLD```** change between calls?
```cc
Profiler::instance()->output (PETSC_COMM_WORLD);
Profiler::instance()->transform_profiler_data(".txt", "SimpleTableFormatter", PETSC_COMM_WORLD);
```

jbrezmorf:
OK. I understand the reason and its fix. 
Concerning the MPI_Communicator. Of course in this particular case it works, however the profiler aims to be a separate library, where it is possible source of errors. Moreover it is just bad design to force user pass again something that has to be equal to what was already passed before.

Other small issue. I realized that the function have badly documented parameters.

jbrezmorf:
To finish this pull, we need rebase against master.

Issue #323, opened 2015-03-11
Replace vector fields


jbrezmorf:
- [x]  Decide form of input, major changes possible. 
	A multifield is just an array of fields. Its description should specify 
	how the number of subfields is determined from the other input data.
	This array allows "transpose" auto conversion (proposed as general mechanism). 

- [x] Change the way Fields are read in, A FieldSet  have method read_input, which pass through the list of Filed descriptors and for every fields it calls virtual method add_change(time, RegionSet, Accessor) and let the Field (or MultiField) store the list of changes. This requires an abstract input accessor.  Duplicate of  #275.

- [x] Implement transposition as general auto conversion method extending current autoconversion of arrays.

in make_storage(Array) if there is non-array on input:
1. set ReaderToStorage::try_transpose_read=1, transpose_index=0, call read of value with array subtype
2. for ReaderToStorage::try_transpose_read=1, check in other make_storage methods that there is Array on input, if so ..
3. ... check that array is non-empty ... else InputException empty array at address ...
4. ... read element at index given by transpose_index using same Input::Type, ASSERT that we ar in array
5. ... for transpose_index==0 add size of array into ReaderToStorage::transpose_array_sizes which is of type vector<unsigned int>
6. ... back in make_storage(Array) check sizes in transpose_array_sizes, and make other calls of reading the subtree for other values of transpose_index.
	
- [x] Use MultiField instead of (variable length) vector valued Fields in all equations. 
	Implement value and value_list in MultiField (can be used in ReactionTerm).
- [x]  Test output of all bulk input fields.

This issue finish effort partly done in #225 and in #172 .
Depends on new input.

jbrezmorf:
EqData to check for usage of MultiFields:
transport_operator_splitting.hh: 76 - TransportBase::TransportEqData
transport.hh:82 - ConvectionTransport::EqData
transport_dg.hh:139
heat_model.hh:45 - these probably should be scalar have to consult with JS
concentration_model.hh:45

jbrezmorf:
Use same set_input_list for field as for multifield, store input_list for multifield and pass it to the fields
in setup_components.

This solve problem with assert in MultiField<>::set_input_list and move subfield construction into set_time.

jbrezmorf:
Try to make dedicate constructor of Field for setup_components, the index and possibly other data may be the parameters.

jbrezmorf:
Can we move FieldCommon<>::set_components into MultiField?

[...]

PR #479, opened 2015-12-08
Df vector fields

jbrezmorf:


src/fields/field.hh
```diff
@@ -98,6 +98,11 @@ public:
     	 * @p FieldBase<...>::function_factory.
     	 */
     	virtual FieldBasePtr create_field(Input::Record rec, const FieldCommon &field);
+
+    	/**
+    	 * Check if Input::Record accessor contains data of field given by input_name
```
jbrezmorf: Add comment: Returns true when ever the method create_field returns non-null pointer, otherwise returns false. 
The difference is that the field is actually not created.

src/fields/multi_field.impl.hh
```diff
@@ -149,14 +149,24 @@ void MultiField<spacedim, Value>::set_up_components() {
...
     	}
-
-    	sub_fields_[i_comp].set_input_list(this->shared_->input_list_);
     }
 }
 
 
 
 template<int spacedim, class Value>
+void MultiField<spacedim,Value>::set_input_list(const Input::Array &list) {
```
jbrezmorf: I think that this will not work. Do we have some unit tests for multifields?
Anyway we can postpone it when we will deal with necessary changes in the MultiFieldFactory.

dflanderka: We have very simple unit test for multifields. It will be completly redesigned in #323.

PR #486, opened 2015-12-16
Js fix con files

jstebel:
Follows the branch JS_trans_dg_under_tos and uses the conserved CON files from master.

Transformation rules written to main.json file under GeoMop subdirectory (maybe the file has to be renamed to avoid accidental rewrite when updating the directory).


src/coupling/balance.cc
```diff
@@ -160,10 +160,27 @@ void Balance::allocate(unsigned int n_loc_dofs,
...
+	const int n_bulk_regs_per_dof = min(10, (int)mesh_->region_db().bulk_size());
 	const unsigned int n_quant = quantities_.size();
-	const unsigned int n_bdr_reg = regions_.boundary_size();
-	const unsigned int n_blk_reg = regions_.bulk_size();
+	const unsigned int n_bdr_reg = mesh_->region_db().boundary_size();
+	const unsigned int n_blk_reg = mesh_->region_db().bulk_size();
+
+
+	// construct vector of regions of boundary edges
+    for (unsigned int loc_el = 0; loc_el < mesh_->get_el_ds()->lsize(); loc_el++)
```
jbrezmorf: Could this loop be removed? It is missing in master and it is not moved neither.

src/coupling/hc_explicit_sequential.cc
```diff
@@ -18,12 +18,7 @@
 
 #include "hc_explicit_sequential.hh"
 #include "flow/darcy_flow_interface.hh"
-#include "transport/advection_process_base.hh"
-// TODO:
```
jbrezmorf: Recover or apply this comment.

src/transport/advection_process_base.hh
```diff
@@ -35,18 +39,18 @@ public:
...
-
     /// Common specification of the input record for secondary equations.
     static Input::Type::Abstract & get_input_type() {
         return Input::Type::Abstract("Transport",
-                "Secondary equation for transport of substances.")
+                "Secondary equation for transport of substances or heat transfer.")
                 .close();
     }
 
+protected:
```
jbrezmorf: Could we move this into EquationBase.

src/transport/concentration_model.cc
```diff
@@ -123,26 +123,12 @@ ConcentrationTransportModel::ModelEqData::ModelEqData()
 
 
 
-UnitSI ConcentrationTransportModel::balance_units()
```
jbrezmorf: How is this replaced? For the heat model it is moved into constructor. Maybe similar setting is missing for the concentration model.

src/transport/transport.cc
```diff
@@ -45,17 +45,38 @@
...
+const int ConvectionTransport::registrar =
+		Input::register_class< ConvectionTransport, Mesh &, const Input::Record >("Convection_FV") +
+		ConvectionTransport::get_input_type().size();
+
+const IT::Record &ConvectionTransport::get_input_type()
+{
+	return IT::Record("Convection_FV", "Explicit in time finite volume method for solute transport.")
+			.derive_from(ConcentrationTransportBase::get_input_type())
+			.declare_key("input_fields", IT::Array(
+					IT::Record("TransportOperatorSplitting_Data", FieldCommon::field_descriptor_record_decsription("TransportOperatorSplitting_Data") )
```
jbrezmorf: Since we do add any particular keys as in the DarcyFlow, the EqData().make_field_descriptor_type("Transport_OS")
may be used directly.

src/transport/transport.h
```diff
@@ -272,9 +303,22 @@ class ConvectionTransport : public TransportBase {
...
+     * TODO: introduce FieldDiscrete -containing true DOFHandler and data vector and pass such object together with other
+     * data. Possibly make more general set_data method, allowing setting data given by name. needs support from EqDataBase.
+     */
+    const MH_DofHandler *mh_dh;
+
 	/// List of indices used to call balance methods for a set of quantities.
 	vector<unsigned int> subst_idx;
 
-            friend class TransportOperatorSplitting;
+    /// (new) object for calculation and writing the mass balance to file.
```
jbrezmorf: Isn't the balance ptr derived form AdvectionBase?

src/coupling/balance.cc
jstebel: Yes, it is a result of wrong merge.

src/transport/concentration_model.cc
jstebel: TransportDG<ConcentrationTransportModel> obtains the instance of Balance from TransportOperatorSplitting, so here it does not need to define its units.

src/transport/transport.cc
jstebel: OK. The argument of make_field_descriptor_type() will be "Convection_FV".

src/transport/advection_process_base.hh
jstebel: OK

Issue #225, opened 2014-10-14
MultiField - replace variable vector valued fields


jbrezmorf:
- [ ] implement value and value_list functions, mark them in documentation as OBSOLETE and possibly slow; one should rather use those methods for individual components of the MultiField
- [x] replace Field::read_field_descriptor_hook with a shared pointer to a FieldBaseFactory class.
  This is defined in the Field<> class and has single virtual method 
  create_field(Input::Record, FieldCommon).
  Default implementation would be as current static method Field<>::read_field_descriptor.
  Other implementations can override this method to create the fields in particular way, namely:

  1. backward compatibility in old_bcd.hh implements several hooks, 
      these can be replaced with single descendent of FieldBaseFactory 
      template with particular field passed at construction.
  
  1. similarly for DarcyFlow::bc_piezo_hook, where we must relay on 
      global variable for gravity field vector, introducing FieldBaseFactory
  
  1. We have to reimplement FieldBaseFactory for multifields to read values 
      from the common key and possibly add missing values into array under the 
      components key in the input Record.

jbrezmorf:
Implement FieldFactory for Multifield:

similar to Field<>::FactoryBase 
but instead of direct call to FieldBaseType::function_factory:
* check if the key "common" is non-empty, if yes and key "components" is empty convert it to the components as follows:

  
* call FieldBaseType::function_factory for every item in the key "components"

jbrezmorf:
FieldElementwise:
see also #255 
* add FieldCommon::set_component_index(unsigned int i) 
to set FieldCommon::component_index_
this specify if the field is part of a multifield and which component it is
use some large value for undefined component (the field is not part of multifield)

* use component_index_ when FieldElementwise ask FieldElementwiseCache for its data.

jbrezmorf:
- add FieldCommon::is_multifield
- use it in FieldSet or other places where we have to deal with different interface of these two clases
  (all these places should be marked with TODO, since our goal is to make these two classes to behave similarly)

- change FieldSet::set_n_components to 
  FieldSet::set_components(const vector<string> &names)
- make same method in FieidCommon, move vector with names from MultiField into FieldCommon
  this method replace set_n_components, change implementation of the method n_comp
- remove MultiField::init; move code into set_time when called first time
- modify implementation of MultiField<...>::set_time to:
   for first time:
	   - check that we have non  empty vector of components
	   - create SubFields (resize its vector)
	   - call set mesh etc. .... for SubFields  
[...]

PR #488, opened 2016-01-03
Df use multifield

jbrezmorf:
Usage of multifields.
- finish multifields implementation
- apply in Flow123d, namely transport equations.


src/fields/field.impl.hh
```diff
@@ -313,21 +326,21 @@ void Field<spacedim,Value>::update_history(const TimeStep &time) {
     // read input up to given time
 	double input_time;
     if (shared_->input_list_.size() != 0) {
-        while( shared_->list_it_ != shared_->input_list_.end()
-        	   && time.ge( input_time = shared_->list_it_->val<double>("time") ) ) {
+        while( shared_->list_idx_ < shared_->input_list_.size()
+        	   && time.ge( input_time = shared_->input_list_[shared_->list_idx_].val<double>("time") ) ) {
```
jbrezmorf: Make reference for shared_->input_list[shared_->list_idx_]

src/fields/field_common.cc
```diff
@@ -67,68 +67,18 @@ const std::string FieldCommon::field_descriptor_record_decsription(const string&
...
 
     // pass through field descriptors containing key matching field name.
     double time;
-    for( auto it = shared_->input_list_.begin<Input::Record>();
-         it != shared_->input_list_.end(); ++it)
-        if (it->find<Input::AbstractRecord>(input_name())) {
-            time = it->val<double>("time"); // default time=0
-            TimeGovernor::marks().add( TimeMark(time, mark_type | TimeGovernor::marks().type_input() ));
-        }
+    for( auto it = shared_->input_list_.begin();
```
jbrezmorf: Just try range loop: 
	for( auto &item : shared->input_list_) 
		...

src/fields/multi_field.impl.hh
```diff
@@ -150,38 +142,72 @@ void MultiField<spacedim, Value>::set_up_components() {
     		sub_fields_[i_comp].shared_->input_name_ = name();
     	}
```
jbrezmorf: Missing setup of the sub_field nema and input name if the comp_names_ is present.

src/fields/multi_field.impl.hh
```diff
@@ -150,38 +142,72 @@ void MultiField<spacedim, Value>::set_up_components() {
...
-			field_algo_base->set_component_idx(index_);
-			return field_algo_base;
+		unsigned int position = 0;
+		for (auto it = multifield_arr.begin<Input::AbstractRecord>(); it != multifield_arr.end(); ++it, ++position)
+		{
+			if (index_ == position) {
+				typename Field<spacedim,Value>::FieldBasePtr field_algo_base = Field<spacedim,Value>::FieldBaseType::function_factory( (*it), field.n_comp() );
+				field_algo_base->set_component_idx(index_);
+				return field_algo_base;
+			}
```
jbrezmorf: Yet missing check for the case multifield_arr.size() != n_components and multifield_arr.size() == 1.
Moreover we need similar check in setup_components but for the whole full_input_list.

src/fields/field.impl.hh
```diff
@@ -60,6 +60,19 @@ Field<spacedim,Value>::Field(const string &name, bool bc)
 
 
 template<int spacedim, class Value>
+Field<spacedim,Value>::Field(unsigned int component_index)
+: data_(std::make_shared<SharedData>())
```
jbrezmorf: The name and input_name should also be passed as parameters of this constructor. This way we
1. get rid of setting input_name explicitly in setup_components
2. better document that input_name is set to the name of the multifield.

unit_tests/coupling/eq_data_test.cpp
```diff
@@ -186,9 +186,14 @@ class SomeEquation : public testing::Test, SomeEquationBase {
          */
 
         DBGMSG("init\n");
+
+        static std::vector<Input::Array> inputs;
```
jbrezmorf: Better documnet why we need this hack.

unit_tests/fields/multi_field_test.cpp
```diff
@@ -22,15 +22,14 @@ FLOW123D_FORCE_LINK_IN_PARENT(field_constant)
...
     },
-    transposed={ 
-        TYPE="MultiField", 
-        component_names=[ "field_1", "field_2", "field_3" ],
-        components=[ { TYPE="FieldConstant", value=1}, { TYPE="FieldConstant", value=2}, { TYPE="FieldConstant", value=3} ]
-    }
+    transposed=[ 
+        { TYPE="FieldConstant", value=1},
+        { TYPE="FieldConstant", value=2},
+        { TYPE="FieldConstant", value=3}
```
jbrezmorf: Test also multifield with components given by different algorithms.
Do we have test of multifield with components of type FieldElementwise? There are some tricks in GMSH reader that needs testing.

src/fields/field.impl.hh
```diff
@@ -313,21 +328,22 @@ void Field<spacedim,Value>::update_history(const TimeStep &time) {
     // read input up to given time
 	double input_time;
     if (shared_->input_list_.size() != 0) {
-        while( shared_->list_idx_ < shared_->input_list_.size()
+        while( shared_->list_idx_ < (int)shared_->input_list_.size()
```
jbrezmorf: Could shared_->list_idx_ be unsigned? It could be initialized to input_list_.size() ==0.

src/fields/field_common.cc
```diff
@@ -38,7 +38,8 @@ FieldCommon::FieldCommon()
 
 
 FieldCommon::FieldCommon(const FieldCommon & other)
-: shared_(other.shared_),
+: name_(other.name_),
```
jbrezmorf: I think that omitting copy of the name_ is intentional (unfortunately not documented). 
E.g. it is used in Field<..>::copy_from. Why you need it with the name_ copy?
To add name_ copy, we must first somehow find all places where this constructor is used.
Definitely, we have to fix its documentation and explain why we do not copy name_.

src/fields/multi_field.impl.hh
```diff
@@ -130,25 +123,29 @@ bool MultiField<spacedim, Value>::is_constant(Region reg) {
...
+    sub_fields_.reserve( comp_size );
+    for(unsigned int i_comp=0; i_comp < comp_size; i_comp++)
     {
+    	if (this->shared_->comp_names_[i_comp].length() == 0)
+    		full_name = name();
+    	else {
+    		full_name = this->shared_->comp_names_[i_comp] + "_" + name();
+    	}
+
+    	sub_fields_.push_back( SubFieldType(i_comp, name(), full_name) );
```
jbrezmorf: I see, here the copy constructor for Field<>should copy name_as well. So we have to check all places where this constructor is used.

src/fields/multi_field.impl.hh
```diff
@@ -158,37 +155,66 @@ template<int spacedim, class Value>
...
-	}
+    // Check sizes of Arrays defined MultiField in field descriptors
+    for (Input::Iterator<Input::Record> it = list.begin<Input::Record>();
+					it != list.end();
+					++it) {
+    	Input::Array mf_array;
+    	if ( it->opt_val(this->input_name(), mf_array) ) {
+    		unsigned int comp_size = this->shared_->comp_names_.size();
+    		if (mf_array.size() != 1 && mf_array.size() != comp_size)
+    			THROW( Input::ExcInputMessage() << EI_Message("Invalid size of Array defined for MultiField " + this->input_name()
```
jbrezmorf: Define new type of the input exception.
The message should contain actual and expected size, e.g.
'Invalid size XX of the MultiField <name>, expected size: YY. '
No need to add address explicitly into the message, see INPUT_EXCEPTION usage.

src/reaction/sorption_base.cc
```diff
@@ -119,14 +119,16 @@ SorptionBase::~SorptionBase(void)
   if (data_ != nullptr) delete data_;
 
   VecScatterDestroy(&(vconc_out_scatter));
-  VecDestroy(vconc_solid);
-
-  for (unsigned int sbi = 0; sbi < substances_.size(); sbi++)
-  {
-    //no mpi vectors
-    xfree(conc_solid[sbi]);
+  if (vconc_solid != NULL) {
```
jbrezmorf: Is it possible to replace vconc_solid + conc_solid by VeSeqDouble ? If yes, just add a TODO to the declaration of vconc_solid.

unit_tests/fields/multi_field_test.cpp
jbrezmorf: If we postpone creation of appropriate tests, make an issue for it.

dflanderka: Created #504

jbrezmorf:
preparedfor final step to implement #225

PR #500, opened 2016-01-12
Js change flux sign

### LINKED ISSUE #492: Unify sign of boundary flux with sources
Change sign of bc fluxes in darcy flow and in transport dg. Positive flux will be inflow into the domain. 
This is convention in hydrology software like Modflow. 
To this end we need additional functionality of the transformation script to scale given values.

Similarly we need to change signe in the balances.

jstebel:
Almost resolves #492. It remains only to add transformation rules for change of sign in bc_flux.


jstebel:
Done.

src/flow/darcy_flow_mh.cc
```diff
@@ -588,7 +588,9 @@ void DarcyFlowMH_Steady::assembly_steady_mh_matrix()
 
                 if (balance_ != nullptr)
                 {
-                    balance_->add_flux_matrix_values(water_balance_idx_, loc_b, {side_row}, {1});
+		    // We sum negative side fluxes which are oriented outwards, since the
+		    // fluxes in input and output are interpreted as incoming.
+                    balance_->add_flux_matrix_values(water_balance_idx_, loc_b, {side_row}, {-1});
```
jbrezmorf: I assume that we need change only on input and on output, this is somehow in the middle so I'm confused.

src/transport/transport.cc
```diff
@@ -368,8 +368,8 @@ void ConvectionTransport::set_boundary_conditions()
                         {
                         	for (unsigned int sbi=0; sbi<n_substances(); sbi++)
                         	{
-                        		balance_->add_flux_matrix_values(subst_idx[sbi], loc_b, {row_4_el[el_4_loc[loc_el]]}, {0.});
-                        		balance_->add_flux_vec_value(subst_idx[sbi], loc_b, flux*value[sbi]);
+//                         		balance_->add_flux_matrix_values(subst_idx[sbi], loc_b, {row_4_el[el_4_loc[loc_el]]}, {0.});
```
jbrezmorf: Same as for Darcy flow, it would be better to change sign just after reading the field in the model so that all internal code use PDE convention.

jstebel:
Internally fluxes reverted to outward. Corrected balance outputs, updated reference manual.

jbrezmorf:
close #492

PR #501, opened 2016-01-12
Js flow total flux

### LINKED ISSUE #413: Flux BC in DarcyFlow
- [ ] Implement combined BC in DarcyFlow, new BC have name 'TotalFlux'

see #367

jstebel:
Resolves #413.


src/flow/darcy_flow_mh.cc
```diff
@@ -175,16 +175,16 @@ DarcyFlowMH::EqData::EqData()
         bc_type.input_selection( &get_bc_type_selection() );
         bc_type.units( UnitSI::dimensionless() );
 
-    ADD_FIELD(bc_pressure,"Dirichlet BC condition value for pressure.");
-    	bc_pressure.disable_where(bc_type, {none, neumann} );
+    ADD_FIELD(bc_pressure,"Pressure value for Dirichlet or Robin/total flux boundary condition.", "0.0");
```
jbrezmorf: Improve description here and two fields later on. 
1) Just mention meaning of the field and in which boundary conditions it is used, using their names
in the input.
2) Describe total_flux using formula

jbrezmorf:
Thanks for good description of BC.

Issue #354, opened 2015-04-16
Embedded Python infrastructure


jbrezmorf:
Related to #336 and #98.

In order to implement parts of flow funcionality in Python, we need:
1. Source directory where to place our python modules (proposed: src/python and its subdirectories)
2. Source directory to place scripts that could be used by users (proposed: bin)
3. Source directory to place advanced/auxiliary scripts (bin/py)
4. Copy modules into 'build_tree/libs'
5. Modify PythonLoader constructor to set access to both system modules (either in system prefix or in our copy of system modules according to
    the COPY_PYTHON variable in the config.cmake) and to our modules. Current implementation is messed with various experiments to overcome 
    a bug in particular CYGWIN version. Is possible that Py_SetProgramName is not necessary. Question how to augment python module search path 
    through the C API.
6. Make some checks in modules of correct python version. Since when modules are used through scripts on the user system they can run under system python,
while when they are used through C API within flow123d they run under our embedded python. Could there be a confict of generated pyc files? 
Are the pyc files even generated by the embedded python? 

7. Test the infrastructure on Profiler formaters.

x3mSpeedy:
- [x]  ad 1) in ```src/python``` is hierarchical structure of python modules
- [ ]   ad 2) currently no script are in ```bin```
- [x]  ad 3) in ```bin/python``` are python scripts which requires python modules located in ```src/python```, path is automatically fixed so **scripts can see modules**
- [x]  ad 4) cmake will copy (if flag COPY_PYTHON) contents of ```src/python``` into ```build_tree/lib/flow123d```. This path is appended so using embedded python in c++ will see modules.
- [x]  ad 5) in cmake is defined constant using ```flow_define``` named ```PYTHON_EXTRA_MODULES_PATH```. Value differ depending on COPY_PYTHON flag. It points to either ```src/python``` or ```build_tree/lib/flow123d```
- [x]  ad 6) all script files includes pathfix python file which appends to module search file ```src/python``` folder.
All script and modules includes python file which checks running version. If version is 3, execution is terminated with error.

After some testing: 
 - pyc files are generated even when embedded python is used. Pyc files behavior is following:

>.pyc files generated whenever the corresponding code elements are imported, and updated if the corresponding code files have been updated.
> The .pyc's timestamp **must be older** than the corresponding .py's timestamp to trigger a recompilation

- There should not be conflict since after copying all timestamps will be changed so pyc will be refreshed.

jbrezmorf:
[...]

Issue #184, opened 2014-05-19
Continuous test of install packages


jbrezmorf:
- create linux and windows integration test nodes without any libraries and development tools
- build and publish install packages
- test every build on integration test nodes

jbrezmorf:
Principally done in pull #506. Remains:
- create Linux package test node - issue #353 
- deal with inconsistent treatment of DLL under cygwin, Python seem's to fail loading its own libraries. The library is found, since when the file is replaced by the fake one it is loaded. The same DLL loads fine when python is started under cygwin shell, but fails to load when python is called from the windows cmd line. Separate issue #556

Issue #183, opened 2014-05-18
self-conteined package with shared libraries


jbrezmorf:
Make a self-contined package for Linux is problematic. We can not simply make everything static 
1) some libraries (python) are only shared in standard distributions
2) system (gcc) libraries are shared

Current state:
We detect shared libraries of the linked binary using 'ldd' command. "system libraries" determined by experimenting are put to lib/system, while other libraries are put into lib directory of the package. Moreover, we add "$ORIGIN/../lib" into RPATH of the flow123d binary, so the loader can find non-system libraries placed relative to the binary. If the target system has libgfortran everything seems to run smoothly as system libraries from target system are used. If libgfortran is not available we must switch to libraries placed into lib/system including loader "ld-linux-*.so". Then we can run binary as follows:

```LD_LIBRARY_PATH=`pwd`/lib/system lib/system/ld-linux-x86-64.so bin/flow123d ```

 TODO:
1) reconsider static linking, it could be possible with our own python build and
without gfortran (NO OPTION)
2) in order to robustly support dynamic linking we should have wrapper script of binary to run provided loader as mentioned above
3) make our own build of python as a constant and controled package containing what we may need e.g. NumPy etc. (NO OPTION)

jbrezmorf:
Ask a unix specialist about best practice.

x3mSpeedy:
## Research related to this problem
### Testing package downloaded from flow.nti.tul.cz
Downloaded [this](http://flow.nti.tul.cz/packages/0.0.master/flow123d_0.0.master_linux_x86_64.tar.gz) package
<hr />

1. Standard execution: ```bin/flow123d ```

  According to [this discussion](http://stackoverflow.com/questions/7251573/gcc-relocation-error-at-runtime) and [this page](https://docs.oracle.com/cd/E19120-01/open.solaris/819-0690/chapter3-5/index.html) problem lies with linker **```ld-linux-x86-64.so```**. Error is complaining on shared library **libc.so.6**
  ```
  bin/flow123d: relocation error: /home/jan-hybs/flow/bin/../lib/libpthread.so.0: symbol __getrlimit, version GLIBC_PRIVATE not defined in file libc.so.6 with link time reference
  ```
  <hr />
2. Wrapper execution: ```./wrapper.sh```

  content of ```wrapper.sh```:
  ```bash
  #!/bin/bash
  export LD_LIBRARY_PATH=`pwd`/lib/system
[...]

PR #503, opened 2016-01-19
JHy embedded python

x3mSpeedy:
Pull fixing #183 (self-contained package) and #354 (embedded Python infra). After merge we can start working on issue #184 (continuous package test).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -553,7 +553,7 @@ endif()
 #
 # optionaly we use Python interface
 set(FLOW_WRAPPER "#!/bin/bash\n")
-set(FLOW_WRAPPER "${FLOW_WRAPPER}\n")
+set(FLOW_WRAPPER "${FLOW_WRAPPER}DIR=\"\$(dirname \$(dirname \$0))\"\n")
 
 if (USE_PYTHON)
     message(STATUS "\n ======================================================= ")
@@ -614,7 +614,7 @@ if (USE_PYTHON)
                 execute_process(
                   COMMAND rsync -aR ${PY_PREFIX}/.${PY_ALT_PATH} ${TARGET_PREFIX}/
                  )
-                 set(FLOW_WRAPPER "${FLOW_WRAPPER}PYTHONPATH=\${PYTHONPATH}:.${PY_ALT_PATH}\n")
+                 set(FLOW_WRAPPER "${FLOW_WRAPPER}PYTHONPATH=\${PYTHONPATH}:\${DIR}${PY_ALT_PATH}\n")
             endforeach(PY_PATH)
             execute_process(
               COMMAND chmod -R 764 ${TARGET_PREFIX}/
@@ -635,7 +635,7 @@ if (USE_PYTHON)
                 )
                 
             list(APPEND PYTHON_WRAPPER_PATHS "./lib/flow123d")
-            set(FLOW_WRAPPER "${FLOW_WRAPPER}PYTHONPATH=\${PYTHONPATH}:./lib/flow123d\n")
+            set(FLOW_WRAPPER "${FLOW_WRAPPER}PYTHONPATH=\${PYTHONPATH}:\${DIR}/lib/flow123d\n")
             # copy python include (h files for embedded python)
             set(PYTHON_COPY_INCLUDE_TARGET include/${PYTHON_SUBDIR})
             execute_process(          
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -604,9 +604,10 @@ if (USE_PYTHON)
             foreach(PY_PATH ${PY_PATHS})
                 string(REPLACE "${PY_PREFIX}" "" PY_ALT_PATH "${PY_PATH}")
                 list(APPEND PYTHON_WRAPPER_PATHS "${PY_ALT_PATH}")
+                message(STATUS "rsync -rR ${PY_PREFIX}/.${PY_ALT_PATH} ${TARGET_PREFIX}/")
                 execute_process(
-                  COMMAND cp -rn ${PY_PATH} ${TARGET_PREFIX}/${PY_ALT_PATH}
-                  )
+                  COMMAND rsync -rR ${PY_PREFIX}/.${PY_ALT_PATH} ${TARGET_PREFIX}/
+                 )
             endforeach(PY_PATH)
             
             message(STATUS "Python prefix: ${PY_PREFIX}")
@@ -616,6 +617,9 @@ if (USE_PYTHON)
             flow_define(PYTHON_PATHS  ${PYTHON_WRAPPER_PATHS})
             
             # copy flow123d python scripts and modules
+            execute_process(
+                COMMAND ${TARGET_PREFIX}/lib/flow123d 
+                )
             execute_process(
                 COMMAND cp -rT ${CMAKE_SOURCE_DIR}/src/python ${TARGET_PREFIX}/lib/flow123d 
                 )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -552,6 +552,9 @@ endif()
 # PYTHON
 #
 # optionaly we use Python interface
+set(FLOW_WRAPPER "#!/bin/bash\n")
+set(FLOW_WRAPPER "${FLOW_WRAPPER}\n")
+
 if (USE_PYTHON)
     message(STATUS "\n ======================================================= ")
     find_package(PythonLibs 2.7)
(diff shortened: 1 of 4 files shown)


bin/copy_shared_libs.sh
```diff
@@ -29,18 +30,31 @@ then
 else  
   if [ -x  ./flow123d ]
   then
+    # detect shared libs
     loader=`ldd flow123d | grep "/ld" | sed 's/^\t*\(\/.*\) (.*)$/\1/'`
-    libs_non_system=`ldd flow123d | grep " => /" | grep -v 'libc\.\|libdl\.\|libgcc_s\.\|libgfortran\.\|libm\.\|libstdc++\.\|librt\.' | sed 's/^.* => \(\/.*\) (.*)$/\1/'`
-    libs_system=`ldd flow123d | grep " => /" | grep 'libc\.\|libdl\.\|libgcc_s\.\|libgfortran\.\|libm\.\|libstdc++\.\|librt\.' | sed 's/^.* => \(\/.*\) (.*)$/\1/'`
-    libs_system="${libs_system} ${loader}"
+    libs=`ldd flow123d | grep -n "^/" | sed 's/^.* => \(\/.*\) (.*)$/\1/'`
+    libs="${libs}"
```
jbrezmorf: Reason for this line? Comment or delete.

x3mSpeedy: Will delete this *very useful* line. Its a leftover from joining ```${libs} {loader}``` together.

CMakeLists.txt
```diff
@@ -572,47 +572,78 @@ if (USE_PYTHON)
...
             
-            # copy python files to build tree 
-            # first get prefix (we support just one prefix)
+            # get all subfolders from paths so we can copy them during install
+            # some are located in lib some in lib64 some in local
             execute_process(
-                COMMAND python -c "import sys; print sys.prefix"
-                OUTPUT_VARIABLE PY_PREFIX
-                OUTPUT_STRIP_TRAILING_WHITESPACE)        
+              COMMAND python -c "import sys, re; print ';'.join(set([[p for p in \\
```
jbrezmorf: This code is quite complex and repeats in the CPack configuration so a script would be better solution.

CMakeLists.txt
```diff
@@ -572,47 +572,78 @@ if (USE_PYTHON)
...
-            set(PYTHON_COPY_LIB_TARGET  lib/${PYTHON_SUBDIR_MINOR})
-            execute_process(
-                COMMAND mkdir -p ${TARGET_PREFIX}/lib 
-                )             
-            # recursive copy, not overwrite existing files
-            # we copy all and exclude some paths and files during install    
+            # copy flow123d python scripts and modules
             execute_process(
-                COMMAND cp -rn ${PY_PREFIX}/lib/${PYTHON_SUBDIR_MINOR} ${TARGET_PREFIX}/lib 
+                COMMAND ${TARGET_PREFIX}/lib/flow123d
```
jbrezmorf: Why to call flow123d?

CMakeLists.txt
```diff
@@ -844,6 +872,8 @@ if( PETSC_FOUND )
   
   
   ### add target for flow123d binary
+  #
+  # set(CMAKE_EXECUTABLE_FORMAT "ELF")
```
jbrezmorf: Remove?

CMakeLists.txt
```diff
@@ -911,21 +941,24 @@ else()
...
 if (PYTHON_COPY)
-    install(DIRECTORY build_tree/${PYTHON_COPY_LIB_TARGET} 
-            DESTINATION lib
-            PATTERN *.dll EXCLUDE
-            PATTERN *.a EXCLUDE
-            PATTERN *.so EXCLUDE
-            PATTERN */dist-packages EXCLUDE
-            PATTERN *.py EXCLUDE)
+    # install all sub folders of Python
+    foreach(PY_SUB_PATH ${PY_SUB_PATHS})
```
jbrezmorf: Could we install whole prefix in one step?

CMakeLists.txt
```diff
@@ -911,21 +941,24 @@ else()
...
+    endforeach(PY_PATH)
+    # copy Python's include
     install(DIRECTORY build_tree/${PYTHON_COPY_INCLUDE_TARGET} 
             DESTINATION include)
 endif()
 
-install(TARGETS    bin/flow123d  
-  RUNTIME DESTINATION bin)
+
+install(PROGRAMS    build_tree/bin/flow123d             RENAME "flow123d.bin"
```
jbrezmorf: We should make this only for Linux platform until we deal with the windows as well.

bin/copy_shared_libs.sh
```diff
@@ -29,18 +30,31 @@ then
...
+    # manually copy and rename file to bin dir so cmake can make it executable
+    cp ${loader} ./ld-linux-loader.so
+    chmod +x "./ld-linux-loader.so"
+    
+    # now we alter wrapper file located in bin/flow123d_wrapper.sh
+    wrapper="flow123d_wrapper.sh"
+    echo "Generating wrapper file '$wrapper'"
+    echo "#!/bin/bash" > $wrapper
+    echo "DIR=\"\$( cd \"\$( dirname \"\${BASH_SOURCE[0]}\" )\" && pwd )\"" >> $wrapper
+    python -c "import sys; print '\n'.join(['PYTHONPATH=\${PYTHONPATH}:\${DIR}/..'+p.replace(sys.prefix, '') for p in sys.path if p])" >> $wrapper
```
jbrezmorf: Use dedicated script.

jbrezmorf: Have to check quoting in the wrapper to deal correctly with spaces in the path.

CMakeLists.txt
```diff
@@ -572,47 +572,78 @@ if (USE_PYTHON)
...
+              OUTPUT_VARIABLE PY_PATHS
+              OUTPUT_STRIP_TRAILING_WHITESPACE)
 
-            string(REGEX MATCH "python[^/]*$" PYTHON_SUBDIR ${PYTHON_INCLUDE_DIRS})
-            string(REGEX MATCH "python[0-9]\\.[0-9]" PYTHON_SUBDIR_MINOR ${PYTHON_SUBDIR})
             
-            # copy python files to build tree 
-            # first get prefix (we support just one prefix)
+            # get all subfolders from paths so we can copy them during install
+            # some are located in lib some in lib64 some in local
```
jbrezmorf: Better comment, give an example. If the code does what I think the comment suggest a different function. However the code is to cryptic.

CMakeLists.txt
```diff
@@ -572,47 +572,78 @@ if (USE_PYTHON)
...
+                COMMAND ${TARGET_PREFIX}/lib/flow123d 
                 )
-            # copy flow123d python scripts and modules
             execute_process(
                 COMMAND cp -rT ${CMAKE_SOURCE_DIR}/src/python ${TARGET_PREFIX}/lib/flow123d 
                 )
-
-            # copy python include            
+                
+            list(APPEND PYTHON_WRAPPER_PATHS "./lib/flow123d")
```
jbrezmorf: This is meaning less since PYTHON_WRAPPER_PATHS is not used further on. Anyway I do not understand why to add flow123d directory at all. The these directories are added into PYTHON_PATH in PythonLoader and should not be used before.

bin/copy_shared_libs.sh
```diff
@@ -29,18 +30,31 @@ then
...
+    
+    # manually copy and rename file to bin dir so cmake can make it executable
+    cp ${loader} ./ld-linux-loader.so
+    chmod +x "./ld-linux-loader.so"
+    
+    # now we alter wrapper file located in bin/flow123d_wrapper.sh
+    wrapper="flow123d_wrapper.sh"
+    echo "Generating wrapper file '$wrapper'"
+    echo "#!/bin/bash" > $wrapper
+    echo "DIR=\"\$( cd \"\$( dirname \"\${BASH_SOURCE[0]}\" )\" && pwd )\"" >> $wrapper
```
jbrezmorf: 1. Rather use here document for the script in order to avoid some escaping and improve readability. 
2. I do not understand meaning of this line, the part before pwd.

CMakeLists.txt
x3mSpeedy: Python's sys.path contains several paths we must copy in order to assure Python's correct functionality. We decided to remove part ```/usr``` from paths to make them shorter. sys.path: 
```
/usr/lib64/python27.zip
/usr/lib64/python2.7
/usr/lib64/python2.7/plat-linux2
/usr/lib64/python2.7/lib-tk
/usr/lib64/python2.7/lib-old
/usr/lib64/python2.7/lib-dynload
/usr/lib64/python2.7/site-packages
/usr/lib/python2.7/site-packages
```

if we remove prefix we get:
```
/lib64/python27.zip
/lib64/python2.7
/lib64/python2.7/plat-linux2
/lib64/python2.7/lib-tk
/lib64/python2.7/lib-old
/lib64/python2.7/lib-dynload
/lib64/python2.7/site-packages
/lib/python2.7/site-packages
```
thus we want to copy folders ```/lib64``` and ```/lib```. Variable ```PY_SUB_PATHS``` contains these values and it comes from Python command called in PYTHON_COPY phase

CMakeLists.txt
x3mSpeedy: There is/was [bug](https://cmake.org/Bug/view.php?id=13934) in cmake and I was not able to **cpack** any tarball. I found workaround but I don't know nor I'm able to find what does command do:
```cmake
set(CMAKE_EXECUTABLE_FORMAT "ELF")
```
But it helped in my case. Should I uncomment or delete it @jbrezmorf ?

CMakeLists.txt
```diff
@@ -576,43 +576,76 @@ if (USE_PYTHON)
...
+            # Python's sys.path contains several paths we must copy in order to assure Python's correct functionality.
+            # We decided to remove part /usr from paths to make them shorter. sys.path:
+            #    /usr/lib64/python2.7
+            #    /usr/lib64/python2.7/site-packages
+            #    /usr/lib/python2.7/site-packages
+            # if we remove prefix we get:
+            #    /lib64/python2.7
+            #    /lib64/python2.7/site-packages
+            #    /lib/python2.7/site-packages
+            # we store first part of each stripped sys.path, remove duplicate and join them with compilation
```
jbrezmorf: It's still unclear what the 'first part' is. You mean: "The script::
1. takes all paths from PYTHON_PATH
2. for each extract the name of the first subdirectory under prefix
3. make list unique 

so we get something like:
"./lib;./lib64;./local"

If this is the function of the script, the name of the output variable should be rather PY_PREFIX_SUBDIR, since PY_SUB_PATHS  evokes something deeper then PY_PATHS, i.e. quite opposite thing.

CMakeLists.txt
jbrezmorf: IF it is at least hopefully working workaround for known bug in cmake, we should check cmake version and use command only for particular version and provide extensive comment including url of appropriate bug report.

CMakeLists.txt
jbrezmorf: I see. The problem was probably misunderstanding meaning of PY_SUB_PATHS, see comment before. Just to make it more clear we should rename TARGET_PREFIX e.g. to PY_BUILD_PREFIX and us it here as parameter of DIRECTORY. instead of 'build_tree'.

PR #505, opened 2016-01-22
Df review mesh

dflanderka:


src/mesh/mesh.h
```diff
@@ -193,7 +194,7 @@ class Mesh {
     /**
      *
      */
-    void read_gmsh_from_stream(istream &in);
+    void read_gmsh_from_stream(istream &in, bool close_region_db = true);
```
jbrezmorf: Why we need additional parameter? 
This method is used mainly in unit tests. Provide appropriate documentation.

src/mesh/region_set.cc
```diff
@@ -0,0 +1,291 @@
...
+
+
+namespace IT = Input::Type;
+
+
+/*******************************************************************
+ * implementation of RegionSetBase
+ */
+
+IT::Abstract & RegionSetBase::get_input_type() {
```
jbrezmorf: This and  following three methods replace by getter for mesh->region_db_, and call its methods directly in child classes.

src/mesh/region.hh
```diff
@@ -589,13 +600,7 @@ private:
...
-
-    /**
-     * Read two operands from input array of strings and check if given names
-     * are existing sets. Return pair of checked set names.
-     */
-    pair<string,string> get_and_check_operands(const Input::Array & operands);
+    void prepare_sets( const string & set_name_1, const string & set_name_2, RegionSet & set_1, RegionSet & set_2) const;
 
     /**
      * Create label of region in format: "region_"+id
```
jbrezmorf: Do we need method replace_region_dim any more?

jbrezmorf: Proposed change of add_region methods:
1. read elements from mesh, create regions (ans region sets) with label derived from ID.
2. read Physical names from the MSH file. Finds region by unique ID and dim, call 
RegionDB::rename_region(region, new_label)
3. process region list from mesh input (further rename regions, creating new etc.)
... calls mainly rename_region
.. special case is creating region from element list

src/mesh/region_set.cc
```diff
@@ -0,0 +1,291 @@
...
+
+
+
+const IT::Record & RegionSetFromElements::get_region_input_type()
+{
+    return IT::Record("From_Elements", "Region declared by name and enum of elements.")
+        .derive_from(RegionSetBase::get_input_type())
+		.declare_key("name", IT::String(), IT::Default::obligatory(),
+				"Label (name) of the region. Has to be unique in one mesh.\n")
+		.declare_key("id", IT::Integer(0), IT::Default::optional(),
```
jbrezmorf: Doc: "If not set the ID is set automatically to be unique. ID is used in output mesh, output field "region_id" and some error messages."

src/mesh/region_set.cc
```diff
@@ -0,0 +1,291 @@
...
+		}
+		mesh->modify_element_ids(region_id, element_ids);
+	}
+}
+
+
+
+const IT::Record & RegionSetFromElements::get_region_input_type()
+{
+    return IT::Record("From_Elements", "Region declared by name and enum of elements.")
```
jbrezmorf: Do we allow adding elements to an existing region?
We like to do so. Thus, document this in description and check implementation.

src/mesh/region_set.cc
```diff
@@ -0,0 +1,291 @@
...
+	unsigned int region_id;
+	string region_label = rec.val<string>("name");
+
+	Input::Iterator<unsigned int> it = rec.find<unsigned int>("id");
+	if (it) {
+		region_id = (*it);
+	} else {
+		Region reg = mesh->region_db().find_label(region_label);
+		if ( reg.is_valid() ) region_id = reg.id();
+		else THROW( ExcNonexistingLabel() << EI_Region_Label(region_label) );
```
jbrezmorf: Allow creating new non-existing region.

src/mesh/region_set.cc
```diff
@@ -0,0 +1,291 @@
...
+	}
```
jbrezmorf: For the case of new region and ID missing on input, we need new method in RegionDB to get an unique ID, e.g.
maximum of  all IDs +1.

src/mesh/region_set.cc
```diff
@@ -0,0 +1,291 @@
...
+		RegionSetFromLabel::get_region_input_type().size();
+
+
+
+/*******************************************************************
+ * implementation of RegionSetFromElements
+ */
+
+RegionSetFromElements::RegionSetFromElements(const Input::Record &rec, Mesh *mesh)
+{
```
jbrezmorf: call RegionSetBase(mesh), and set RegionSetBase::region_db_;

src/mesh/region_set.cc
```diff
@@ -0,0 +1,291 @@
...
+
+
+
+const IT::Record & RegionSetUnion::get_region_input_type()
+{
+    return IT::Record("Union", "Defines region as a union of given two or more regions.")
+        .derive_from(RegionSetBase::get_input_type())
+		.declare_key("name", IT::String(), IT::Default::obligatory(),
+				"Label (name) of the region. Has to be unique in one mesh.\n")
+		.declare_key("region_labels", IT::Array( IT::String(), 2,2 ), IT::Default::obligatory(),
```
jbrezmorf: Rename "region_labels" to "regions", array of size >=1.
Add key "region_ids".
Create union of all referenced regions.

src/mesh/region.cc
```diff
@@ -373,19 +398,27 @@ void RegionDB::prepare_sets( const string & set_name_1, const string & set_name_
 
 
 
-pair<string,string> RegionDB::get_and_check_operands(const Input::Array & operands)
+void RegionDB::prepare_set( const string & set_name, RegionSet & set) const {
+	std::map<std::string, RegionSet>::const_iterator it = sets_.find(set_name);
```
jbrezmorf: name the type

src/mesh/region.cc
```diff
@@ -327,8 +321,23 @@ RegionSet RegionDB::union_sets( const string & set_name_1, const string & set_na
 	return set_union;
 }
 
+RegionSet RegionDB::union_sets( RegionSet target_set, const string & source_set_name) const {
```
jbrezmorf: Check set names on call side, pass here reference to the set itself. The the method prepare_set is mere sort and coud be removed. Same for intersection and difference.

src/mesh/region_set.cc
```diff
@@ -0,0 +1,315 @@
...
+	}
+
+	if (regions) {
+		std::vector<string> set_names = mesh->region_db().get_and_check_operands(*regions);
+		for (string set_name : set_names) {
+			region_set = region_db_.union_sets( region_set, set_name );
+		}
+	}
+
+	region_db_.add_set(name_of_set, region_set);
```
jbrezmorf: Keep both keys "region_ids" and "regions" optional. Here check if the resulting region_set is not empty, possibly throw an input exception "Empty result of Union operation."

src/fields/field_interpolated_p0.impl.hh
```diff
@@ -68,6 +68,7 @@ void FieldInterpolatedP0<spacedim, Value>::init_from_input(const Input::Record &
     {
        source_mesh_ = new Mesh();
        reader_file_ = FilePath( rec.val<FilePath>("gmsh_file") );
+       ReaderInstances::instance()->get_reader(reader_file_)->read_physical_names( source_mesh_ );
```
jbrezmorf: No need to call read_physical_names.

src/mesh/mesh.cc
```diff
@@ -745,5 +736,28 @@ vector<int> const & Mesh::elements_id_maps( bool boundary_domain) const
     return bulk_elements_id_;
 }
 
+void Mesh::read_regions_from_input(Input::Array region_list)
+{
+	std::shared_ptr<RegionSetBase> reg_ptr;
+	for (Input::Iterator<Input::AbstractRecord> it = region_list.begin<Input::AbstractRecord>();
+				it != region_list.end();
+				++it) {
+		reg_ptr = (*it).factory< RegionSetBase, const Input::Record &, Mesh * >(*it, this);
```
jbrezmorf: Remove reg_ptr, comment that constructor have side effect in the mesh.

src/mesh/msh_gmshreader.cc
```diff
@@ -199,86 +203,87 @@ void GmshMeshReader::read_elements(Tokenizer &tok, Mesh * mesh) {
...
+            unsigned int id = lexical_cast<unsigned int>(*tok_); ++tok_;
+            string name = *tok_; ++tok_;
 
-            mesh->region_db_.add_region(id, name, dim);
+            mesh->region_db_.add_region(id, name, dim, "defined in '$PhysicalNames' section");
         }
 
     } catch (bad_lexical_cast &) {
-        xprintf(UsrErr, "Wrong format of number, %s.\n", tok.position_msg().c_str());
+        xprintf(UsrErr, "Wrong format of number, %s.\n", tok_.position_msg().c_str());
```
jbrezmorf: Turn to exception.

src/mesh/region.hh
```diff
@@ -361,21 +356,20 @@ public:
      * Parameter @p id is any unique non-negative integer, parameter @p label is unique string identifier of the region,
      * @p dim is dimension of reference elements in the region and @p boundary is true if the region consist of boundary elements
      * (where one can apply boundary condition).
-     *
      */
-    Region add_region(unsigned int id, const std::string &label, unsigned int dim);
+    Region add_region(unsigned int id, const std::string &label, unsigned int dim, const std::string &address ="");
```
jbrezmorf: comment meaning of parameter address.

src/mesh/region.cc
```diff
@@ -126,62 +95,73 @@ Region RegionDB::add_region( unsigned int id, const std::string &label, unsigned
...
+	region_table_.replace(
+			region_table_.get<Index>().find(index),
+            item);
+
+	Region new_reg = Region(index, *this);
+	// add region to sets
+	RegionSet region_set;
+	region_set.push_back( new_reg );
+	this->add_set(new_label, region_set);
+	if (old_boundary_flag != reg.is_boundary()) {
```
jbrezmorf: Print warning.

src/mesh/region.cc
```diff
@@ -126,62 +95,73 @@ Region RegionDB::add_region( unsigned int id, const std::string &label, unsigned
...
+        } else {
+            // region with same label and other ID exists
+            THROW(ExcNonuniqueLabel() << EI_Label(new_label) << EI_ID(reg.id()) << EI_IDOfOtherLabel(it_label->get_id()) );
+        }
     }
 
-    return insert_region(id, label, undefined_dim, boundary);
+	// replace region label
+	unsigned int index = reg.idx();
+	sets_.erase( reg.label() ); // remove RegionSet of old name
```
jbrezmorf: Try to just change label of appropriate item in sets_.

```
const int oldKey = 33;
const int newKey = 22;
const iterator it = m.find(oldKey);
if (it != m.end()) {
  // Swap value from oldKey to newKey, note that a default constructed value 
  // is created by operator[] if 'm' does not contain newKey.
  std::swap(m[newKey], it->second);
  // Erase old key-value from map
  m.erase(it);
}
```

src/mesh/region.cc
```diff
@@ -314,84 +294,41 @@ void RegionDB::add_set( const string& set_name, const RegionSet & set) {
...
 
-	std::stable_sort(set_1.begin(), set_1.end(), Region::comp);
-	std::stable_sort(set_2.begin(), set_2.end(), Region::comp);
+	if (it != sets_.end()) {
+		RegionSet & set = (*it).second;
+		auto set_it = std::find(set.begin(), set.end(), region);
+		if ( set_it!=set.end() ) {
+			set.erase(set_it);
+		} else {
+			xprintf(Warn, "Erased region was not found in set '%s'", set_name.c_str());
```
jbrezmorf: Turn to ASSERT. Both.

src/mesh/region.hh
```diff
@@ -571,9 +553,11 @@ private:
     unsigned int n_boundary_;
     /// Number of bulk regions
     unsigned int n_bulk_;
+    /// Maximal value of Region::id()
+    unsigned int max_index_;
```
jbrezmorf: Rename to max_id_. Same for the getter in RegionSetBase.

src/mesh/region.cc
```diff
@@ -126,62 +95,73 @@ Region RegionDB::add_region( unsigned int id, const std::string &label, unsigned
         THROW(ExcNonuniqueLabel() << EI_Label(label) << EI_ID(id) << EI_IDOfOtherLabel(it_label->get_id()) );
```
jbrezmorf: Check that these exceptions are caught and input address is provided in calling functions. Add into comment to that functions, that they may throw and address has to be provided.

src/mesh/region.cc
```diff
@@ -399,127 +336,13 @@ RegionSet RegionDB::get_region_set(const string & set_name) const {
...
+string RegionDB::create_label_from_id(unsigned int id) const {
 	stringstream ss;
 	ss << "region_" << id;
-	ss.str(label);
+	return ss.str();
 }
 
-Region RegionDB::insert_region(unsigned int id, const std::string &label, unsigned int dim, bool boundary) {
+Region RegionDB::insert_region(unsigned int id, const std::string &label, unsigned int dim, bool boundary, const std::string &address) {
 	if (closed_) THROW( ExcAddingIntoClosed() << EI_Label(label) << EI_ID(id) );
```
jbrezmorf: Check address for the exception.

src/mesh/region.cc
```diff
@@ -531,9 +354,25 @@ Region RegionDB::insert_region(unsigned int id, const std::string &label, unsign
 		n_bulk_++;
 	}
 	if (index >= max_n_regions) xprintf(UsrErr, "Too many regions, more then %d\n", max_n_regions);
-	if ( ! region_table_.insert( RegionItem(index, id, label, dim) ).second )
+	if ( ! region_table_.insert( RegionItem(index, id, label, dim, address) ).second )
 	   THROW( ExcCantAdd() << EI_Label(label) << EI_ID(id) );
```
jbrezmorf: Check address for the exception.

src/mesh/region.cc
```diff
@@ -568,3 +407,51 @@ Region RegionDB::find_by_dimid(DimIDIter it_id, unsigned int id, const std::stri
...
+
+	// print header
+	stream << endl << "----------- Table of all regions: -----------" << endl;
+	stream << std::setfill(' ') << setw(6) << "id" << " dim label" << setw(12) << "" << "contains regions" << endl;
+	// print data
+	for (RegionSetTable::const_iterator it = sets_.begin(); it != sets_.end(); ++it) { // iterates through RegionSets
+		unsigned int reg_id = RegionIdx::undefined;
+		string rset_label = it->first;
+		LabelIter label_it = region_table_.get<Label>().find(rset_label);
+		if ( label_it != region_table_.get<Label>().end() ) { // checks if Region with same label as RegionSet exists
```
jbrezmorf: This condition is supposed to give same result as the latter one (it->second.size() ==1. Rather merge both conditions.
Print single region set by line, reporting labels of elementary regions for size>1, and ID  and dim for size==1.

src/mesh/region.cc
```diff
@@ -568,3 +407,51 @@ Region RegionDB::find_by_dimid(DimIDIter it_id, unsigned int id, const std::stri
...
+			reg_id = label_it->get_id(); // stores ID of Region - used if RegionSet contains only one Region
+			stream << setw(6) << reg_id << setw(4) << label_it->dim();
+		} else {
+			stream << "     -   -";
+		}
+		unsigned int indent = 17 - rset_label.length();
```
jbrezmorf: possibly better: <<setw(17) << rset_label

src/mesh/region.cc
```diff
@@ -568,3 +407,51 @@ Region RegionDB::find_by_dimid(DimIDIter it_id, unsigned int id, const std::stri
...
+			}
+		} else if ( it->second.size() == 1 && (it->second)[0].id() != reg_id ) {
+			// for one Region in RegionSet prints ID only if RegionSet is not wrapper of Region
+			stream << (it->second)[0].id();
+		} else {
+			stream << "-";
+		}
+		stream << endl;
+	}
+	stream << "---------------------------------------------" << endl << endl;
```
jbrezmorf: ?? use fill?

src/mesh/region.cc
```diff
@@ -568,3 +407,51 @@ Region RegionDB::find_by_dimid(DimIDIter it_id, unsigned int id, const std::stri
...
+
+
+void RegionDB::check_regions() {
+	ASSERT(closed_, "RegionDB not closed yet.\n");
+
+	for (RegionTable::index<Index>::type::iterator it = region_table_.get<Index>().begin();
+			it!= region_table_.get<Index>().end();
+			++it) {
+		if (!it->used)
+			THROW(ExcUnusedRegion() << EI_Label(it->label) << EI_ID(it->get_id()) << EI_RegionAddress(it->address) );
```
jbrezmorf: How EI_RegionAddress and EI_Address differs? Why don't use it->ei_address()?

src/mesh/region.hh
```diff
@@ -268,17 +269,15 @@ region_sets = [
  * 1) Read PhysicalNames form GMSH file, populate RegionDB (DONE in GMSH reader, may need small modifications)
  * 2) Read region definitions from the input, see
  *
- * typedef std::map<unsigned int, unsigned int> MapElementIDToRegionID;
- * RegionDB::read_regions_from_input(Input::Array region_list, MapElementIDToRegionID &map);
+ * Mesh::read_regions_from_input(Input::Array region_list);
  *
  *
  *    (TODO in RegionDB, also creates (and return to Mesh) element regions modification map: std::map< unsigned int, RegionIdx>
```
jbrezmorf: Remove TODO, done.

src/mesh/region.hh
```diff
@@ -268,17 +269,15 @@ region_sets = [
...
  *     that maps element IDs to the new region names, GMSH reader should have setter method to accept this map
  *     and modify the elements during reading)
- * 3) Read region sets - TODO in RegionDB
- * 4) Read boundary key of the Mesh record and mark appropriate regions as boundary (TODO in RegionDB)
- * 5) Read nodes (DONE in GMSH reader)
- * 6) Read elements, per element:
+ * 3) Read boundary key of the Mesh record and mark appropriate regions as boundary (TODO in RegionDB)
```
jbrezmorf: This point is meaningless now.

src/mesh/region.hh
```diff
@@ -331,7 +326,7 @@ public:
     TYPEDEF_ERR_INFO( EI_NumOp, unsigned int);
     DECLARE_INPUT_EXCEPTION( ExcWrongOpNumber, << "Wrong number of operands. Expect 2, given: " << EI_NumOp::val);
 
-    DECLARE_INPUT_EXCEPTION(ExcUniqueRegionId, << "Id of region must be unique, id: " << EI_ID::val );
+    DECLARE_INPUT_EXCEPTION(ExcUniqueRegionId, << "Id of elementary region must be unique, id: " << EI_ID::val );
```
jbrezmorf: Seems to be same as ExcNonuniqueLabel.

src/mesh/region.hh
```diff
@@ -299,30 +298,26 @@ public:
...
-                                             << "other region with same ID but different label: " << EI_LabelOfOtherID::qval << " already exists\n");
-    DECLARE_EXCEPTION( ExcNonuniqueLabel, << "Non-unique label during add of region id: " << EI_ID::val << ", label: " << EI_Label::qval << "\n" \
-                                             << "other region with same label but different ID: " << EI_IDOfOtherLabel::val << " already exists\n");
-    DECLARE_EXCEPTION( ExcInconsistentBoundary, << "Inconsistent add of region with id: " << EI_ID::val << ", label: " << EI_Label::qval << "\n" \
-                                             << "both ID and label match an existing region with different boundary flag.");
-    DECLARE_EXCEPTION( ExcInconsistentDimension, << "Inconsistent add of region with id: " << EI_ID::val << ", label: " << EI_Label::qval << "\n" \
-                                             << "both ID and label match an existing region with different dimension.");
+    DECLARE_EXCEPTION( ExcNonuniqueID, << "Non-unique ID during add of elementary region id: " << EI_ID::val << ", label: " << EI_Label::qval << "\n" \
```
jbrezmorf: For all exceptions:
- consider turn them into INPUT_EXCEPTION

src/mesh/region_set.cc
```diff
@@ -0,0 +1,385 @@
...
+	Region reg;
+	string region_label = rec.val<string>("name");
+	unsigned int region_id = rec.val<unsigned int>("id");
+
+	reg = mesh->region_db().find_label(region_label);
+	if ( reg.is_valid() ) {
+		if ( region_id != reg.id() )
+			THROW( ExcInconsistentLabelId() << EI_Label(region_label) << EI_ID( region_id ) << rec.ei_address() );
+	} else {
+		try {
```
jbrezmorf: why this try, find_id doesn't throw.

src/mesh/msh_gmshreader.cc
```diff
@@ -199,86 +203,87 @@ void GmshMeshReader::read_elements(Tokenizer &tok, Mesh * mesh) {
...
 
-            unsigned int dim = lexical_cast<unsigned int>(*tok); ++tok;
-            unsigned int id = lexical_cast<unsigned int>(*tok); ++tok;
-            string name = *tok; ++tok;
+            unsigned int dim = lexical_cast<unsigned int>(*tok_); ++tok_;
+            unsigned int id = lexical_cast<unsigned int>(*tok_); ++tok_;
+            string name = *tok_; ++tok_;
 
-            mesh->region_db_.add_region(id, name, dim);
+            mesh->region_db_.add_region(id, name, dim, "defined in '$PhysicalNames' section");
```
jbrezmorf: Store just the input address without wrapping text. in this case form a special address:
(mesh: <file>, $PhysicalNames)

src/mesh/region_set.cc
```diff
@@ -0,0 +1,385 @@
...
+			reg = mesh->region_db().find_id(region_id);
+		} catch(RegionDB::ExcUniqueRegionId &e) {
+			e << rec.ei_address();
+		}
+		if ( reg.is_valid() ) {
+			region_db_.rename_region(reg, region_label);
+		} else {
+			stringstream ss;
+			ss << "declare at address \'" << rec.address_string() << "\'";
```
jbrezmorf: just address

src/mesh/region_set.cc
```diff
@@ -0,0 +1,385 @@
...
+const IT::Record & RegionSetFromId::get_region_input_type()
+{
+    return IT::Record("From_Id", "Region declared by id and name.\n"
+    					"Allows to create new region with given id and label\n"
+    					"or specify existing region by id which will be renamed.")
+        .derive_from(RegionSetBase::get_input_type())
+		.declare_key("name", IT::String(), IT::Default::obligatory(),
+				"Label (name) of the region. Has to be unique in one mesh.\n")
+		.declare_key("id", IT::Integer(0), IT::Default::obligatory(),
+				"The ID of the region to which you assign label.")
```
jbrezmorf: Could we add also an optional key for specification of the dimension?

src/mesh/region_set.cc
```diff
@@ -0,0 +1,385 @@
...
+    return IT::Record("From_Elements", "Region declared by name, ID and enum of elements.\n"
+    								   "Allows to get existing region or create new and assign\n"
+    								   "elements to its. Elements are specified by ids.")
+        .derive_from(RegionSetBase::get_input_type())
+		.declare_key("name", IT::String(), IT::Default::obligatory(),
+				"Label (name) of the region. Has to be unique in one mesh.\n")
+		.declare_key("id", IT::Integer(0), IT::Default::optional(),
+				"The ID of the region to which you assign label.\n"
+				"If new region is created and ID is not set, unique ID will be generated automatically.")
+		.declare_key("element_list", IT::Array( IT::Integer(0) ), IT::Default::optional(),
```
jbrezmorf: Any problem to make this key obligatory? Without the key it have no sense. Further the size of the list should be at least 1.

src/mesh/region_set.cc
```diff
@@ -0,0 +1,385 @@
...
+				"Specification of the region by the list of elements. This is not recommended")
```
jbrezmorf: Remove the last sentence.

src/mesh/region_set.cc
```diff
@@ -0,0 +1,385 @@
...
+ */
+
+RegionSetDifference::RegionSetDifference(const Input::Record &rec, Mesh *mesh)
+: RegionSetBase(mesh)
+{
+	string name_of_set = rec.val<string>("name");
+	Input::Iterator<Input::Array> labels = rec.find<Input::Array>("regions");
+
+	std::vector<string> set_names = mesh->region_db().get_and_check_operands(*labels);
+	if ( set_names.size() != 2 ) THROW(RegionDB::ExcWrongOpNumber() << RegionDB::EI_NumOp(set_names.size()) << labels->ei_address() );
```
jbrezmorf: No need to check the size, it comes from declaration of the input array.

src/mesh/region_set.cc
```diff
@@ -0,0 +1,385 @@
...
+	Input::Iterator<Input::Array> region_ids = rec.find<Input::Array>("region_ids");
+	Input::Iterator<Input::Array> regions = rec.find<Input::Array>("regions");
+
+	std::set<Region, bool (*)(const Region&, const Region&)> set(Region::comp);
+	if (region_ids) {
+		for (Input::Iterator<unsigned int> it_ids = region_ids->begin<unsigned int>();
+				it_ids != region_ids->end();
+		        ++it_ids) {
+			try {
+				Region reg = region_db_.find_id(*it_ids);
```
jbrezmorf: find_id doesn't throw.

src/mesh/region_set.cc
```diff
@@ -0,0 +1,385 @@
...
+				if (reg.is_valid()) {
```
jbrezmorf: Move condition after try block. Improve readability.

src/mesh/region_set.cc
```diff
@@ -0,0 +1,385 @@
...
+: RegionSetBase(mesh)
+{
+	unsigned int region_id;
+	string region_label = rec.val<string>("name");
+	Region reg = mesh->region_db().find_label(region_label);
+
+	Input::Iterator<unsigned int> it = rec.find<unsigned int>("id");
+
+	if ( reg.is_valid() ) { // region with the specified label exists
+		region_id = reg.id();
```
jbrezmorf: Rather throw always if the label already exists and change the exception message (region of same label already exists), current message is difficult to understand. Also check if RegionDB do not declare a similar exception. Could be reused.

src/mesh/region_set.cc
```diff
@@ -0,0 +1,385 @@
...
+					xprintf(Warn, "Region with id %d doesn't exist. Skipping\n", (*it_ids));
+				}
+			} catch(RegionDB::ExcUniqueRegionId &e) {
+				e << region_ids->ei_address();
+				throw;
+			}
+		}
+	}
+
+	if (regions) {
```
jbrezmorf: In order to reuse this code in update_history:
- make a union method in RegionDB getting vector of region names returning a region set.
- just convert IDs to labels for region_ids, merge that list with "regions" whole list pass to the new RegionDB::union method.

src/mesh/region.cc
dflanderka: Second condition (it->second.size() ==1) indicates case of region set which was created as difference or intersection and contains only one region. Label of this region is not same with label of region set. I think displaying of two regions with same ID, dim and different label would be confusing.

jbrezmorf: Result of set opeartions is just a region set i.e. non-elementary region and should have same format as all other sets. The point of my comment is, that we should distinguish just two kind of sets:
1) single region sets with name equal to region should be printed as elementary regions with ID and dim
2) other sets should be printed as list of region labels, here I suggest enclosing the list into brackets, e.g.

.region_xy 		boundary 	id: 101	dim: 2
region_yz 		bulk		 	id: 102	dim: 1
set				[ .region_xy, region_yz ]

src/mesh/region_set.cc
dflanderka: Done.

src/mesh/region.cc
dflanderka: All checks of exceptions done.

Issue #419, opened 2015-08-31
Python-markdown parser bug does not process "<" in html-doc


jstebel:
In documentation, string like "(($q<0$))" or "(($q\\\\<0$))" is not processed properly when invoking "make html-doc". Target "ref-doc" works OK.

x3mSpeedy:
Ad 1) ```(($q<0$))``` should be fixed now. New version of ```main.js``` will be downloaded upon html-doc call.
Ad 2) ```(($q\<0$))``` Is it valid expression? HTML documentation uses KaTeX library. Maybe some expressions are not implemented yet. @jstebel can you test it [here](https://khan.github.io/KaTeX/) (no need to surround expressions with ```(($``` on that site)

jstebel:
Ad  2) Indeed, \< is not valid LaTeX tag. Otherwise I would appreciate text formatting with \boldsymbol and \mathbb.

jstebel:
Another bug: In pdf manual the underscore is escaped both in text and math mode, so lower indices like (($q_w$)) do not work.

x3mSpeedy:
Javascript rendering engine added support for **bold** fonts. So expression 
```tex
$nobold\space \bold{bold}$
```
will look like this:
![image](https://cloud.githubusercontent.com/assets/894428/12946926/22f7f144-cff8-11e5-92eb-a8d7504dae75.png)

Issue with underscore should also be fixed now.

PR #506, opened 2016-01-25
JHy - add continuous testing to yaml template

x3mSpeedy:
Related to issue #184 
 - now testing Windows and Linux platform
 - testing following:
    - download&install
    - running flow123d binary
    - running flow123d binary and check embedded python functionality
    - uninstalling application


src/python/profiler/profiler_formatter_module.py
```diff
@@ -34,125 +31,146 @@ class ProfilerJSONDecoder (json.JSONDecoder):
...
-
-        # return True on success
-        #return True
-
-
-def convert (json_location, output_file, formatter):
-  """Simple iface method for c api"""
-  fmt = ProfilerFormatter ()
-  return fmt.convert (json_location, output_file, formatter)
+        Logger.instance().info('Processing file "%s"', json_location)
```
jbrezmorf: Do not now how the Logger works, however I prefer to simply have error messages coming from exceptions on the screen. So if anything went  wrong the Flow123d propagates the error message to its own exception. Any mechanism that hides exceptions prevents fixing bugs and leads to spurious behaviour.

jbrezmorf:
Unfortunately this pull also include lot of (GOOD) work on documentation and other unrelated python scripts, this hides essential changes.
On the other hand windows release builds on CI2 still fail, due to missing flow123d_wrapper.sh.
The package tests also fail.

x3mSpeedy:
@jbrezmorf this pull also changes to fix issue #419 
Whole Python IST processing was updated to more flexible requirements can be met (such as `Parameters`, `references` for individual nodes, more flexible `node types` and field names (`name == type_name`) and more...) 

For reference I'm also attaching generated [flow123d_doc.pdf](https://github.com/flow123d/flow123d/files/215045/flow123d_doc.pdf)

PR #512, opened 2016-02-08
Df update field descriptor

dflanderka:


src/fields/field.impl.hh
```diff
@@ -351,14 +351,22 @@ void Field<spacedim,Value>::update_history(const TimeStep &time) {
...
-			if (actual_list_item.opt_val("region", domain_name)) {
-				domain = mesh()->region_db().get_region_set(domain_name);
-				if (domain.size() == 0) {
-					THROW( RegionDB::ExcUnknownSetOperand()
-							<< RegionDB::EI_Label(domain_name) << actual_list_item.ei_address() );
+			if (actual_list_item.opt_val("region", domain_name_array)) {
+				std::set<Region, bool (*)(const Region&, const Region&)> set(Region::comp);
+				std::vector<string> domain_names;
+				domain_name_array.copy_to(domain_names);
+				for (std::string set_name : domain_names) {
```
jbrezmorf: We should postpone this after "mesh review". Move union operation back into RegionDB (sorry for that) and reuse it here

PR #517, opened 2016-02-11
JHy memory profiler

x3mSpeedy:


src/system/sys_profiler.hh
```diff
@@ -65,15 +67,23 @@ class MPI_Functions {
...
     static double max(double* val, MPI_Comm comm);
+    static long max(long* val, MPI_Comm comm);
 };
 
 // Assuming all compilers support constexpr
 #define CONSTEXPR_ constexpr
 
+// macro which does nothing but silence g++ compiler warning
+// about unused variable
+#define UNUSED(expr) (void)(expr)
```
jbrezmorf: Remove. To deal with ierr return codes use cherr or cherr_assert from system/global_defs.hh instead.

src/system/sys_profiler.hh
```diff
@@ -214,7 +224,60 @@ inline CONSTEXPR_ unsigned int str_hash(const char * str, unsigned int default_v
  */
 #define CODE_POINT_EXT(tag, subtag) CodePoint(tag, subtag, __FILE__, __func__, __LINE__)
 
+/**
+ * Simple allocator which uses default malloc and free functions. Dields allocated
```
jbrezmorf: Clearly document that this is only internal class. I suggest to move it into separate file and into the namespace 'internal'.

src/system/sys_profiler.hh
```diff
@@ -294,7 +357,19 @@ public:
...
+     */
+    void pause();
+    /**
+     * Resume current timer, load measured petsc memory information
+     */
+    void resume();
+    /**
+     * Debug information of the timer
+     */
+    void info();
```
jbrezmorf: Rather implement << operator.

src/system/sys_profiler.hh
```diff
@@ -294,7 +357,19 @@ public:
      * Start the timer. If it is already started, just increase number of starts (recursions) and calls.
      */
     void start();
-
+    /**
+     * Pause current timer, save measured petsc memory information util resume
```
jbrezmorf: Describe better how this mechanism interact with PETSC calls. Possibly pause and resume may be protected (since Profiler is a friend class) and the user should not call them directly.

src/system/sys_profiler.hh
```diff
@@ -376,8 +451,50 @@ protected:
...
-
-
+    /**
+     * Maximum number of bytes allocated at one time in this frame (not include subframes).
+     */
+    size_t max_allocated_;
+    /**
+     * Current number of bytes allocated in this frame (not include subframes).
+     */
+    size_t current_allocated_;
```
jbrezmorf: Do you mean "Total number of bytes allocated"?, i.e. opposite of total_deallocated_. Rename, modify comment.

src/system/sys_profiler.hh
```diff
@@ -376,8 +451,50 @@ protected:
...
+    int alloc_called;
+    /**
+     * Number of times delete/delete[] operator was used in this scope
+     */
+    int dealloc_called;
+    
+    /**
+     * Number of bytes used by Petsc at the start of time-frame
+     */
+    PetscLogDouble petsc_start_memory;
```
jbrezmorf: Make Profiller work also without PETSC,  so use the compiler flag FLOW123D_HAVE_PETSC to switch 
PETSC related code off.  Further even if we have petsc we need a means to turn its ussage off, e.g. in unit tests.
I suggest to add an optional flag to the constructor to turn PETSC memory tracking on.

src/system/sys_profiler.hh
```diff
@@ -376,8 +451,50 @@ protected:
...
+    /**
+     * Maximum amount of memory used that was PetscMalloc()ed at any time 
+     * during this run.
+     *
+     * The memory usage reported here includes all Fortran arrays (that may be
+     * used in application-defined sections of code).
+     */
+    PetscLogDouble petsc_peak_memory;
+    /**
+     * local aximum amount of memory used that was PetscMalloc()ed
```
jbrezmorf: Isn't this only a lower estimate of the true maximum?

src/system/sys_profiler.hh
```diff
@@ -376,8 +451,50 @@ protected:
...
+     *
+     * The memory usage reported here includes all Fortran arrays (that may be
+     * used in application-defined sections of code).
+     */
+    PetscLogDouble petsc_peak_memory;
+    /**
+     * local aximum amount of memory used that was PetscMalloc()ed 
+     * used during time-frame pause/resume
+     */
+    PetscLogDouble petsc_local_peak_memory;
```
jbrezmorf: Mark this as auxiliary variable for petsc_peak_memory.

src/system/sys_profiler.hh
```diff
@@ -524,6 +638,25 @@ public:
      */
     inline double actual_cumulative_time() const
         { return timers_[actual_node].cumulative_time(); }
+    /**
```
jbrezmorf: What is the purpose of these getters?

src/system/sys_profiler.hh
```diff
@@ -574,9 +707,41 @@ public:
      * Check if the instance was created.
      */
     static bool is_initialized() { return (_instance != NULL); }
```
jbrezmorf: Is this method useful anyway?

src/system/sys_profiler.hh
```diff
@@ -574,9 +707,41 @@ public:
...
+    
+    /**
+     * Check if the instance was created.
+     */
+    static void* operator new (size_t sz);
```
jbrezmorf: Wrong comment. Where are other overloaded allocation operators?

src/system/sys_profiler.hh
```diff
@@ -574,9 +707,41 @@ public:
...
+    
+    /**
+     * Method will propagate values from children timers to its parents
+     */
+    void propagate_timers ();
```
jbrezmorf: Should be private.

src/system/sys_profiler.hh
```diff
@@ -574,9 +707,41 @@ public:
...
+     * Public setter to turn on/off memory monitoring
+     * @param value whether to turn monitoring on or off
+     */
+    void static set_memory_monitoring(const bool value);
+    
+    /**
+     * Public getter to memory monitoring
+     * @return memory monitoring status
+     */
+    bool static get_memory_monitoring();
```
jbrezmorf: Who cares? Isn't it too Javish?

src/system/sys_profiler.hh
```diff
@@ -574,9 +707,41 @@ public:
...
+    /**
+     * Method will propagate values from children timers to its parents
+     */
+    void propagate_timers ();
 
+    /**
+     * Public setter to turn on/off memory monitoring
+     * @param value whether to turn monitoring on or off
+     */
+    void static set_memory_monitoring(const bool value);
```
jbrezmorf: Yep, put the petsc memory tracking flag here.

src/system/sys_profiler.cc
```diff
@@ -164,19 +263,22 @@ string Timer::code_point_str() const {
...
 Profiler* Profiler::_instance = NULL;
 CodePoint Profiler::null_code_point = CodePoint("__no_tag__", "__no_file__", "__no_func__", 0);
 
-void Profiler::initialize()
-{
-
-    if (!_instance)
+void Profiler::initialize() {
+    if (_instance == NULL) {
+        MemoryAlloc::malloc_map().reserve(100*1000);
```
jbrezmorf: Introduce appropriate constants in Profiler class

src/system/sys_profiler.cc
```diff
@@ -630,21 +829,143 @@ void Profiler::uninitialize() {
...
+
+void * Profiler::operator new (size_t size) {
+    return malloc (size);
+}
+
+
+void *operator new (std::size_t size) OPERATOR_NEW_THROW_EXCEPTION {
+	void * p = malloc(size);
+    
+    if (!Profiler::is_initialized()) {
```
jbrezmorf: Why to save the memory size if the profilling is off?

src/system/sys_profiler.cc
```diff
@@ -630,21 +829,143 @@ void Profiler::uninitialize() {
...
+        MemoryAlloc::init_overhead_map()[(long)p] = static_cast<int>(size);
+    }else
+    
+    if (Profiler::get_memory_monitoring()) {
```
jbrezmorf: ?wrong indent

src/system/sys_profiler.cc
```diff
@@ -630,21 +829,143 @@ void Profiler::uninitialize() {
...
+        Profiler::instance()->notify_malloc(size);
+        MemoryAlloc::malloc_map()[(long)p] = static_cast<int>(size);
+    }
+    
+
+	return p;
+}
+
+void *operator new[] (std::size_t size, const std::nothrow_t&) throw() {
```
jbrezmorf: Is there also similar variant of new() operator?

src/system/sys_profiler.cc
jbrezmorf: I suggest to call notify_malloc() always and check if we do memory monitoring and add into MemoryAlloc there. After all the functionality of MemoryAlloc may be internal to the Profiller itself.

src/system/sys_profiler.cc
```diff
@@ -630,21 +829,143 @@ void Profiler::uninitialize() {
...
+    if (Profiler::get_memory_monitoring()) {
+        Profiler::instance()->notify_malloc(size);
+        MemoryAlloc::malloc_map()[(long)p] = static_cast<int>(size);
+    }
+
+	return p;
+}
+
+void operator delete( void *p) throw() {
+    if (Profiler::get_memory_monitoring()) {
```
jbrezmorf: Similar to new operators call Profiler::notify_free, with the pointer as parameter and do the whole stuff there. This also reduce code redundancy.

src/system/sys_profiler.cc
```diff
@@ -97,20 +116,92 @@ Timer::Timer(const CodePoint &cp, int parent)
...
+    parent.total_allocated_ += child.total_allocated_;
+    parent.total_deallocated_ += child.total_deallocated_;
+    parent.alloc_called += child.alloc_called;
+    parent.dealloc_called += child.dealloc_called;
+
+    // add differences from child
+    parent.petsc_memory_difference += child.petsc_memory_difference;
+    parent.current_allocated_ += child.current_allocated_;
+    
+    // when computing maximum, we take greater value from parent and child
```
jbrezmorf: According to the comment the operation have no sense, I'm confused.

src/system/sys_profiler.cc
```diff
@@ -215,6 +316,7 @@ void Profiler::set_program_info(string program_name, string program_version, str
 
 
 int  Profiler::start_timer(const CodePoint &cp) {
+    Timer parent_timer = timers_[actual_node];
```
jbrezmorf: Shouldn't this be a reference? Otherwise pause have no effect.

src/system/sys_profiler.cc
```diff
@@ -269,7 +375,15 @@ void Profiler::stop_timer(const CodePoint &cp) {
                 }
                 // close 'node' itself
                 timers_[actual_node].stop(false);
-                actual_node=timers_[actual_node].parent_timer;
+                actual_node = timers_[actual_node].parent_timer;
+                
+                // workaround for time-frame 0, if (actual_node >=0) would make more sense
+                //   but time-frame 0 has also parent_timer equal to 0
+                if (actual_node == child_timer && actual_node == 0)
```
jbrezmorf: actual_node == child_timer should be enough. That's classical indication of the root of the rooted tree.

src/system/sys_profiler.cc
```diff
@@ -258,6 +363,7 @@ void Profiler::stop_timer(const CodePoint &cp) {
         if (timer.child_timers[i] != timer_no_child)
             ASSERT( ! timers_[timer.child_timers[i]].running() , "Child timer '%s' running while closing timer '%s'.\n", timers_[timer.child_timers[i]].tag().c_str(), timer.tag().c_str());
 #endif
+    unsigned int child_timer = actual_node;
```
jbrezmorf: both stop_timer methods now exhibit too big code redundancy, simplify.

src/system/sys_profiler.cc
```diff
@@ -425,13 +582,26 @@ std::shared_ptr<std::ostream> Profiler::get_default_output_stream() {
...
+
 void Profiler::output(MPI_Comm comm, ostream &os) {
     int ierr, mpi_rank, mpi_size;
     //wait until profiling on all processors is finished
     MPI_Barrier(comm);
     stop_timer(0);
+    propagate_timers();
+    
+    // stop monitoring memory
+    bool memory_monitoring_ = get_memory_monitoring();
```
jbrezmorf: This is completely against our naming convention having the ending underscore for class members.

src/system/sys_profiler.cc
```diff
@@ -630,21 +829,143 @@ void Profiler::uninitialize() {
...
         _instance->stop_timer(0);
+        set_memory_monitoring(false);
         delete _instance;
         _instance = NULL;
     }
 }
 
+bool Profiler::monitor_memory = false;
+void Profiler::set_memory_monitoring(const bool value) {
+    if (value && !monitor_memory) {
```
jbrezmorf: Should be on just for debugging.

src/system/sys_profiler.hh
```diff
@@ -690,6 +855,38 @@ public:
...
+     * Create static map containing <allocation address, allocation size> pairs
+     *   map is used for storing allocations and deallocations of all object not 
+     *   related to profiler after profiler initialization phase
+     */
+	static unordered_map_with_alloc & malloc_map();
+    /**
+     * Create static map containing <allocation address, allocation size> pairs
+     * map is used to store ONLY allocation before profiler is fully initialized
+     */
+    static unordered_map_with_alloc & init_overhead_map();
```
jbrezmorf: Why we need to monitor allocations before the Profiler is initialized?

src/system/sys_profiler.hh
```diff
@@ -690,6 +855,38 @@ public:
...
+    /**
+     * Sums given map and returns sum value
+     * @param  map unordered_map_with_alloc map
+     * @return     total sum of a given map
+     */
+    static int sum(unordered_map_with_alloc & map);
```
jbrezmorf: I don't see any place that use this method.

src/system/sys_profiler.hh
x3mSpeedy: They are used in unit tests. We can't directly access Timer values since they are private.

src/system/sys_profiler.hh
x3mSpeedy: Yes, before calling `malloc_notify`a nd `malloc_free` we have to make sure `Profiler` is created and static variable `_instance` is always available. We can't test `_instance` variable because it is private.

src/system/sys_profiler.cc
x3mSpeedy: There is  also `new`, `delete` and `delete[]` variant. But I'm not really sure we are using `nothrow` functionality:
```cc
int p* = new (std:nothrow) int [1024];
```

src/system/sys_profiler.hh
jbrezmorf: The design should not be dependent on unit tests. The better way is to make data members protected and in unit tests create Fixture - auxiliary class derived from the class to test.

x3mSpeedy:
Unit tests for Profiler are using inheritance of Gtest class

src/system/sys_profiler.cc
```diff
@@ -97,20 +117,124 @@ Timer::Timer(const CodePoint &cp, int parent)
   hash_idx_(cp.hash_idx_),
   parent_timer(parent),
   total_allocated_(0),
-  total_deallocated_(0)
+  total_deallocated_(0),
+  max_allocated_(0),
+  current_allocated_(0),
+  alloc_called(0),
+  dealloc_called(0),
+  petsc_start_memory(0),
```
jbrezmorf: Should be enough to put petsc related members under conditional compilation.

src/system/sys_profiler.hh
```diff
@@ -369,16 +390,76 @@ protected:
...
+    PetscLogDouble petsc_local_peak_memory;
+    #endif // FLOW123D_HAVE_PETSC
+    
     friend class Profiler;
+    friend std::ostream & operator <<(std::ostream&, const Timer&);
+    
+    /**
+     * if under unit testing, specify friend so protected members can be tested
+     */
+    #ifdef __UNIT_TEST__
```
jbrezmorf: Can we remove this now?

x3mSpeedy: It is still necessary. In testing we need to use `friend` pattern. If we were to use inheritance we would still not able to access protected members of Profiler since they belong to static instance of Profiler.

Right now, we set `ProfilerTest` as friend of `Profiler` and we inherit `Gtest` also in `ProfilerTest` class. This way we can access all protected members as well as create tests.

unit_tests/system/profiler_test.cpp
```diff
@@ -9,69 +9,70 @@
...
 #include <flow_gtest_mpi.hh>
 
-
-
+/**
+ * Fixture class for testing Profiler protected members
+ */
+class ProfilerTest: public testing::Test {
+    public:
+        void test_str_hash();
```
jbrezmorf: Using TEST_F you need not to decalare functions for individual tests.

x3mSpeedy: The main problem is that **instance** of Profiler already exists and is static meaning no other class have access to its members if they are protected.
I did not manage to find any other way to solve this issue. There are couple of ideas ([[1]](http://stackoverflow.com/a/30075309/2616048), [[2]](http://stackoverflow.com/a/3676936/2616048) and [[3]](https://github.com/google/googletest/blob/master/googletest/docs/AdvancedGuide.md#testing-private-code))

But the problem with using only `TEST_F` is following (taken from GoogleTest docs)

> Private class members are only accessible from within the class or by friends. To access a class' private members, you can declare your test fixture as a friend to the class and define accessors in your fixture. Tests using the fixture can then access the private members of your production class via the accessors in the fixture. **Note that even though your fixture is a friend to your production class, your tests are not automatically friends to it, as they are technically defined in sub-classes of the fixture.**

So we need to have testing code in fixture class methods, otherwise we have no access to it (and inheritance wont help us). Example of ProfilerTest fixture class:
```c++
class ProfilerTest: public testing::Test, public Profiler {
};

TEST_F(ProfilerTest, test_memory_propagation) {
    cout << Profiler::instance()->timers_.size() << endl;
}
```
compilation error:
```c++
sys_profiler.hh:730:53: error: ‘std::vector Profiler::timers_’ is protected


```

jbrezmorf: OK I get it. Didn't know that TEST_F is in fact derived class. Yet we
can get rid of PI, AN etc. macros and replace them by appropriate
accessors as suggested in google test doc, isn't it?

However I let you decide what you prefer..



On 04/13/2016 02:44 PM, Jan Hybš wrote:
> In unit_tests/system/profiler_test.cpp
> <https://github.com/flow123d/flow123d/pull/517#discussion_r59540582>:
> 
>>  #include <flow_gtest_mpi.hh>
>>  
>> -
>> -
>> +/**
>> + * Fixture class for testing Profiler protected members
>> + */
>> +class ProfilerTest: public testing::Test {
>> +    public:
>> +        void test_str_hash();
> 
> The main problem is that *instance* of Profiler already exists and is
> static meaning no other class have access to its members if they are
> protected.
> I did not manage to find any other way to solve this issue. There are
> couple of ideas ([1] <http://stackoverflow.com/a/30075309/2616048>, [2]
> <http://stackoverflow.com/a/3676936/2616048> and [3]
> <https://github.com/google/googletest/blob/master/googletest/docs/AdvancedGuide.md#testing-private-code>)
> 
> But the problem with using only |TEST_F| is following (taken from
> GoogleTest docs)
> 
>     Private class members are only accessible from within the class or
>     by friends. To access a class' private members, you can declare your
>     test fixture as a friend to the class and define accessors in your
>     fixture. Tests using the fixture can then access the private members
>     of your production class via the accessors in the fixture. *Note
>     that even though your fixture is a friend to your production class,
>     your tests are not automatically friends to it, as they are
>     technically defined in sub-classes of the fixture.*
> 
> So we need to have testing code in fixture class methods, otherwise we
> have no access to it (and inheritance wont help us). Example of
> ProfilerTest fixture class:
> 
> class ProfilerTest: public testing::Test, public Profiler {
> };
> 
> TEST_F(ProfilerTest, test_memory_propagation) {
>     cout << Profiler::instance()->timers_.size() << endl;
> }
> 
> compilation error:
> 
> sys_profiler.hh:730:53: error: ‘std::vector Profiler::timers_’ is protected
> 
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub
> <https://github.com/flow123d/flow123d/pull/517/files/b3bb808b313add2a5a28d5b844061b817228dcf7#r59540582>
> 

-- 
------------------------
Mgr. Jan Brezina, Ph. D.
Technical University in Liberec, New technologies institute
http://www.nti.tul.cz/cz/WikiUser:Jan.Brezina

x3mSpeedy: Main problem with replacing macros is following:
 `ProfilerTest` must be defined before `Profiler` (so it can be marked as friend)
but macros which should be replaced are working with unknown types since `Timer` is not included yet because it is in the same file as `Profiler`. These condition can never be met as long as both classes remain in the same file.

Since we are dealing with macros which only makes test more organized I suggest to keep macros for now but when we detach profiler to its own project I suggest to separate classes so we can have options when including.

PR #522, opened 2016-02-17
Df mesh multi proc

dflanderka:


src/mesh/partitioning.hh
```diff
@@ -44,6 +44,13 @@ public:
     static const Input::Type::Selection & get_tool_sel();
     static const Input::Type::Record & get_input_type();
 
+	TYPEDEF_ERR_INFO(EI_MeshFile, std::string);
+	TYPEDEF_ERR_INFO(EI_NElems, unsigned int);
+	TYPEDEF_ERR_INFO(EI_NProcs, unsigned int);
+	DECLARE_EXCEPTION(ExcDecomposeMesh,
+			<< "Decomposition of mesh defined in the input file " << EI_MeshFile::qval << " can't be done.\n"
```
jbrezmorf: Just slight improvement of the error message:

Number of processors (...) greater then number of elements (...). Can not make partitioning of the mesh (...).

PR #524, opened 2016-02-19
Df generic type params

dflanderka:


src/input/type_abstract.cc
```diff
@@ -60,8 +60,11 @@ TypeBase::TypeHash Abstract::content_hash() const
     boost::hash_combine(seed, "Abstract");
     boost::hash_combine(seed, type_name());
     boost::hash_combine(seed, child_data_->description_);
+	for (ParameterMap::iterator it=parameter_map_->begin(); it!=parameter_map_->end(); it++) {
```
jbrezmorf: prefered range for loop: for(auto &param : parameter_map_)

src/input/type_abstract.cc
```diff
@@ -60,8 +60,11 @@ TypeBase::TypeHash Abstract::content_hash() const
...
+		boost::hash_combine(seed, (*it).first );
+		boost::hash_combine(seed, (*it).second );
+	}
     //boost::hash_combine(seed, child_data_->generic_content_hash_);
```
jbrezmorf: Add hash of abstract.generic_type_

src/input/type_base.hh
```diff
@@ -252,6 +245,12 @@ protected:
     /// flag is true if type should be root of generic subtree
     bool root_of_generic_subtree_;
 
+    /// hash string of generic type if type is derived, or empty string
+    std::string generic_type_;
```
jbrezmorf: Rather use directly hash value.

src/input/type_base.hh
```diff
@@ -252,6 +245,12 @@ protected:
...
+
+    /// map of parameters if type is part of generic subtree
+    boost::shared_ptr<ParameterMap> parameter_map_;
```
jbrezmorf: Do we need shared pointer? Simple ParameterMap parameter_map_ fits better.

src/input/type_generic.cc
```diff
@@ -105,10 +105,12 @@ bool Instance::finish(bool is_generic) {
 
 std::string print_parameter_vec(std::vector<TypeBase::ParameterPair> vec) {
 	stringstream ss;
+	ss << "[ ";
```
jbrezmorf: How added brackets interferes with output?

src/input/type_base.cc
```diff
@@ -109,6 +110,20 @@ void TypeBase::add_attribute(std::string name, json_string val) {
 
 
 void TypeBase::write_attributes(ostream& stream) const {
+	// print hash of generic type and parameters into separate keys
+	if (generic_type_.size()) {
```
jbrezmorf: Output os generic_type_ and parameter_map_ do not match the name of method etc. Move as much as possible into TypeOutputBase. Any reason to have write_attributes is in TypeBase?

jbrezmorf: Remove this method completely even from unit_tests, use the fixture class instead. 
Touch the test: type_base_test.cpp

PR #535, opened 2016-03-08
Df it tuple

dflanderka:


src/input/accessors_impl.hh
```diff
@@ -240,7 +240,7 @@ template<class T>
 inline typename Iterator<T>::OutputType * Iterator<T>::operator ->() const {
     BOOST_STATIC_ASSERT(
```
jbrezmorf: Add comment:
OutputType has to be an accessor to call its method, e.g.  iter->val<AnyType>("xyz"). iter has to be e.g. Iterator<Record>.

src/input/reader_to_storage.cc
```diff
@@ -428,6 +437,57 @@ StorageBase * ReaderToStorage::make_storage(PathBase &p, const Type::Array *arra
...
+	if ( (arr_size = p.get_array_size()) != -1 ) {
+
+		StorageArray *storage_array = new StorageArray(tuple->size());
+        // check individual keys
+		for ( Type::Record::KeyIter it= tuple->begin(); it != tuple->end(); ++it) {
+        	if ( p.down(it->key_index) ) {
+                // key on input => check & use it
+                StorageBase *storage = make_storage(p, it->type_.get());
+                if ( (typeid(*storage) == typeid(StorageNull)) && it->default_.has_value_at_declaration() ) {
+                	storage_array->new_item(it->key_index, make_storage_from_default( it->default_.value(), it->type_ ) );
```
jbrezmorf: Just set storage pointer. First deallocate its content. Then use just one common call to storage_array->new_item
Check for deallocation also in other checks for the StorageNull.

src/input/type_tuple.cc
```diff
@@ -0,0 +1,216 @@
...
+	TypeHash seed=0;
+    boost::hash_combine(seed, "Tuple");
+    data_->content_hash(seed);
+    return seed;
+}
+
+
+Tuple & Tuple::allow_auto_conversion(const string &from_key)
+{
+	THROW( ExcForbiddenTupleMethod() << EI_Method("allow_auto_conversion") << EI_TupleName(this->type_name()) );
```
jbrezmorf: Replace with an ASSERT

src/input/type_tuple.cc
```diff
@@ -0,0 +1,216 @@
...
+        data_->auto_conversion_key=data_->keys.begin()->key_;
+    }
+
+    return (data_->finished);
+}
+
+
+Tuple &Tuple::derive_from(Abstract &parent)
+{
+	THROW( ExcForbiddenTupleMethod() << EI_Method("derive_from") << EI_TupleName(this->type_name()) );
```
jbrezmorf: ASSERT

src/input/type_tuple.cc
```diff
@@ -0,0 +1,216 @@
...
+	for ( KeyIter it= this->begin(); it != this->end(); ++it) {
+		if ( it->default_.is_obligatory() ) ++obligatory_keys_count;
+	}
+	return obligatory_keys_count;
+}
+
+
+TypeBase::MakeInstanceReturnType Tuple::make_instance(std::vector<ParameterPair> vec) const {
+	Tuple tuple = this->deep_copy();
+	ParameterMap parameter_map;
```
jbrezmorf: Try to move common body of this method into private method in Record.

dflanderka: Done.

PR #536, opened 2016-03-08
Js apply multifields

### LINKED ISSUE #323: Replace vector fields

jstebel:
Change vector-valued fields into multifields. Required some additional tuning of field classes.
Resolves #323.


src/fields/multi_field.hh
```diff
@@ -110,7 +110,10 @@ public:
      * We have to override the @p set_mesh method in order to call set_mesh method for subfields.
      */
     void set_mesh(const Mesh &mesh) override;
-
+    
+    
+    MultiField &set_bc(bool is_bc)
```
jbrezmorf: Make this same as for the Field, thus introduce the class BCMultiField.

PR #539, opened 2016-03-17
Js trans disable where

### LINKED ISSUE #114: TransportDG disable_where for boundary fields
TransportDG has Vector valued boundary fields bc_typ, bc_flux, bc_robin_sigma;

several issues:
-  We either have to generalize disable_where method or finally have individual equation for every substance.

- (the fields should not be vector valued for heat transport, how can we achieve that)

- finally currently we can not set disable_where for dirichlet condition field since it is added in model constructor.

- think over default values for transport_dg, they are little overused. This disable error checking
 However currently we at least report warning when the default value is used.

jstebel:
Resolves #114.

### DIFF
--- a/src/fields/multi_field.impl.hh
+++ b/src/fields/multi_field.impl.hh
@@ -31,7 +31,8 @@ namespace it = Input::Type;
 
 template<int spacedim, class Value>
 MultiField<spacedim, Value>::MultiField(bool bc)
-: FieldCommon()
+: FieldCommon(),
+  no_check_control_field_(nullptr)
 {
 	static_assert(Value::NRows_ == 1 && Value::NCols_ == 1, "");
 	this->multifield_ = true;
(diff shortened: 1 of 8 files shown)


src/fields/multi_field.impl.hh
```diff
@@ -58,6 +59,17 @@ it::Array MultiField<spacedim,Value>::get_multifield_input_type() {
 
 
 template<int spacedim, class Value>
+auto MultiField<spacedim, Value>::disable_where(
+        const MultiField<spacedim, typename FieldValue<spacedim>::Enum > &control_field,
+        const vector<FieldEnum> &value_list) -> MultiField &
+{
+    no_check_control_field_=&control_field;
```
jbrezmorf: Could we rather call disable_where for individual subfields? We should also check that the control_field have same size.
This way we save redundant member no_check_control_field_ of the MultiField.

src/fields/multi_field.impl.hh
```diff
@@ -143,6 +155,8 @@ void MultiField<spacedim, Value>::setup_components() {
 
     	sub_fields_.push_back( SubFieldType(i_comp, name(), full_name, is_bc()) );
     	sub_fields_[i_comp].units( units() );
+        if (no_check_control_field_ != nullptr)
```
jbrezmorf: Check that no_check_control_field_ is setup and have same size.

PR #540, opened 2016-03-23
Df field copy constr

dflanderka:


jbrezmorf:
Remains:
- do not use set_mesh since it have side effects on the original field.
- check copy constructor of FieldCommon, copy its data memebrs properly
- make copy of region_fields_
- check copy constructor for Multifield

PR #542, opened 2016-03-29
DF convert asserts

jbrezmorf:
- rename ASSERT to OLD_ASSERT
- replace OLD_ASSERTs in input dir


src/system/asserts.hh
```diff
@@ -0,0 +1,132 @@
...
+#define FEAL_ASSERT_B(x) FEAL_ASSERT_OP(x, A)
+/// Clever macro recursion
+#define FEAL_ASSERT_OP(x, next) \
+    FEAL_ASSERT_A.add_value((x), #x).FEAL_ASSERT_ ## next
+
+
+#ifdef FLOW123D_DEBUG_ASSERTS
+/// High-level macro
+#define FEAL_ASSERT( expr) \
+if ( (expr) ) ; \
```
jbrezmorf: semicolon ?!

src/system/asserts.hh
```diff
@@ -0,0 +1,132 @@
...
+/// Make an assertion
+static Assert make_assert(const std::string& expression) {
+	return Assert(expression);
+}
+
+} // namespace feal
+
+// Must define the macros afterwards
+/// Clever macro A
+#define FEAL_ASSERT_A(x) FEAL_ASSERT_OP(x, B)
```
jbrezmorf: use underscore prefix to mark these macros internal. Say it also in the comment.

src/system/asserts.hh
```diff
@@ -0,0 +1,132 @@
...
+	std::string expression_;                  ///< Assertion expression
+	const char* file_name_;                   ///< Actual file.
+	const char* function_;                    ///< Actual function.
+	int line_;                                ///< Actual line.
+	std::vector< std::string > current_val_;  ///< Formated strings of names and values of given variables.
+
+};
+
+/// Make an assertion
+static Assert make_assert(const std::string& expression) {
```
jbrezmorf: Is it necessary?

src/system/stack_trace.cc
```diff
@@ -0,0 +1,118 @@
...
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+ *
+ *
+ * @file    stack_trace.cc
+ * @brief
+ */
+
+#include "system/stack_trace.hh"
```
jbrezmorf: missing config.h

src/system/stack_trace.cc
```diff
@@ -0,0 +1,118 @@
...
+            && frame.find("throw_exception") != string::npos
+            ) break;
+    }
+    i_frame++;
+
+    unsigned int out_i_frame=0;
+    for(;i_frame< n_frames_; i_frame++, out_i_frame++) {
+        string frame(frames_[i_frame]);
+
+        bool is_to_cut = false; // check if frame is intended to cut
```
jbrezmorf: Move this before the loop. It should replace check for "boost", "exception_detail", "throw_exception". 
Th point is, that when exception is thrown we can remove begginging of the stack trace up to the throw_exception method from boost. However if we call what() in assert we should provide different condition to cut the stacktrace.
Nevertheless current call sequence do not allow to pass the "frames_to_cut" from the Assert down into this method. We must find a way to do so. 
Definitely this must be debugged under Linux in order to make stacktrace  available.

PR #560, opened 2016-04-20
JHy profiler memory leak fix

x3mSpeedy:
Pull which fixes r/w memory leaks caused by:
  1. getting reference from `vector`
  2. adding to vector `push_back` (may induce vector reallocating)
  3. accessing reference which may no longer be valid (depends whether reallocation occurred)

Also fixes unit-test for `profiler` running in parallel (`VecCreateSeq`) and adds precaution against inconsistent profiler tree reduction.

### DIFF
--- a/src/system/sys_profiler.cc
+++ b/src/system/sys_profiler.cc
@@ -555,25 +555,22 @@ void Profiler::add_timer_info(ReduceFunctor reduce, property_tree::ptree* holder
     for (int j = 0; j < Timer::max_n_childs; j++) {
         if (child_timers[j] != timer.child_timers[j]) {
             // this is severe error, timers indicies are different
-            if (child_timers[j] != timer_no_child) {
-                printf("Severe error: timers indicies are different\n");
+            ASSERT(child_timers[j] == timer_no_child, "Severe error: timers indicies are different");
+            
+            // explore the difference in more depth
+            if (timer.child_timers[j] == timer_no_child) {
+                // this processor does not have child timer
+                xprintf(Warn,
+                    "Inconsistent tree in '%s': this processor[%d] does not have child-timer[%d]\n",
+                    timer.tag().c_str(), mpi_rank, child_timers[j]);
             } else {
-                printf("Inconsistent tree: ");
-                if (timer.child_timers[j] == timer_no_child) {
-                    // this processor does not have child timer
-                    printf("this processor[%d] does not have child-timer[%d]\n", mpi_rank, child_timers[j]);
-                    printf("Child timer belongs to parent: \n");
-                    cout << timer << endl;
-                } else {
-                    // other processors do not have child timer
-                    printf("this processor[%d] contains child-timer[%d] which others do not\n", mpi_rank, timer.child_timers[j]);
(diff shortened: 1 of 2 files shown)


src/system/sys_profiler.cc
```diff
@@ -544,12 +544,44 @@ void Profiler::add_timer_info(ReduceFunctor reduce, property_tree::ptree* holder
...
+    // time-frame other processors will forget this time-frame (information lost)
+    int child_timers[ Timer::max_n_childs];
+    MPI_Allreduce(&timer.child_timers, &child_timers, Timer::max_n_childs, MPI_INT, MPI_MIN, MPI_COMM_WORLD);
+    #ifdef FLOW123D_DEBUG
+    int mpi_rank = MPI_Comm_rank(MPI_COMM_WORLD, &mpi_rank);
+    for (int j = 0; j < Timer::max_n_childs; j++) {
+        if (child_timers[j] != timer.child_timers[j]) {
+            // this is severe error, timers indicies are different
+            if (child_timers[j] != timer_no_child) {
+                printf("Severe error: timers indicies are different\n");
```
jbrezmorf: Use ASSERT here instead.

src/system/sys_profiler.cc
```diff
@@ -544,12 +544,44 @@ void Profiler::add_timer_info(ReduceFunctor reduce, property_tree::ptree* holder
...
+            } else {
+                printf("Inconsistent tree: ");
```
jbrezmorf: Use xprintf(Warn, ...) or xprintf(Log, ...) instead here and later on. We are working with David on new approach to logging, however its not ready yet.

src/system/sys_profiler.cc
```diff
@@ -544,12 +544,44 @@ void Profiler::add_timer_info(ReduceFunctor reduce, property_tree::ptree* holder
...
+                } else {
+                    // other processors do not have child timer
+                    printf("this processor[%d] contains child-timer[%d] which others do not\n", mpi_rank, timer.child_timers[j]);
+                    printf("Child timer belongs to parent timer: \n");
+                    cout << timer << endl;
+                }
+            }
+        }
+    }
+    cout << endl;
```
jbrezmorf: Remove.

src/system/sys_profiler.cc
```diff
@@ -544,12 +544,44 @@ void Profiler::add_timer_info(ReduceFunctor reduce, property_tree::ptree* holder
...
+            // this is severe error, timers indicies are different
+            if (child_timers[j] != timer_no_child) {
+                printf("Severe error: timers indicies are different\n");
+            } else {
+                printf("Inconsistent tree: ");
+                if (timer.child_timers[j] == timer_no_child) {
+                    // this processor does not have child timer
+                    printf("this processor[%d] does not have child-timer[%d]\n", mpi_rank, child_timers[j]);
+                    printf("Child timer belongs to parent: \n");
+                    cout << timer << endl;
```
jbrezmorf: can not use cout directly, sorry, postpone this to new logging approach, will support redirection.

Issue #556, opened 2016-04-18
Python libs under Windows


jbrezmorf:
- deal with inconsistent treatment of DLL under cygwin, Python seem's to fail loading its own libraries. The library is found, since when the file is replaced by the fake one it is loaded. The same DLL loads fine when python is started under cygwin shell, but fails to load when python is called from the windows cmd line.

jbrezmorf:
Will be resolved by usage of Docker.

PR #562, opened 2016-04-25
JHy - temp fix to windows-cygwin-python issue #556

x3mSpeedy:
- for now disable python conversion under win/cyg

### DIFF
--- a/src/system/sys_profiler.cc
+++ b/src/system/sys_profiler.cc
@@ -37,6 +37,7 @@
 #include "system/python_loader.hh"
 #include "mpi.h"
 #include "time_point.hh"
+#include "../main.h"
 
 // namespace alias
 namespace property_tree = boost::property_tree;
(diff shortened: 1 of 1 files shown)


src/system/sys_profiler.cc
```diff
@@ -845,6 +843,18 @@ void Profiler::transform_profiler_data (const string &output_file_suffix, const
     PyObject_CallObject (convert_method, arguments);
 
     PythonLoader::check_error();
+
+    #else
+
+    // print information about windows-cygwin issue and offer manual solution
+    cout << "# Note: converting json profiler reports is not";
```
jbrezmorf: Use stringstream  + xprintf(Mesg, ...

src/system/sys_profiler.cc
```diff
@@ -549,7 +550,10 @@ void Profiler::add_timer_info(ReduceFunctor reduce, property_tree::ptree* holder
     // using operation MPI_MIN, so if some processor does not contain some 
     // time-frame other processors will forget this time-frame (information lost)
     int child_timers[ Timer::max_n_childs];
-    MPI_Allreduce(&timer.child_timers, &child_timers, Timer::max_n_childs, MPI_INT, MPI_MIN, MPI_COMM_WORLD);
+    if (Application::petsc_initialized) {
+        MPI_Allreduce(&timer.child_timers, &child_timers, Timer::max_n_childs, MPI_INT, MPI_MIN, MPI_COMM_WORLD);
```
jbrezmorf: Sorry, this is not solution. method output() just should not use MPI at all. I disable this check, it would be better to implement our own MPI reduce operation for the tree sooner then planed.

jbrezmorf:
Cherry pick done.

Issue #570, opened 2016-05-03
Comfortable asserts


jbrezmorf:
- rename to ASSERT_DBG, FEAL_ASSERT_DBG

- (re)introduce macros ASSERT_LT,GT,EQ,LE,GE expadngin, e.g to:
 ASSERT_LT( idx, vector.size() ) ....
 FEAL_ASSERT( idx < vetor.size() ) (idx) (vector.size() )

thus one can use:
ASSERT_LT( idx, vector.size() ).warning("XYZ");

Postpone following after loggers( #280 )
- warning output goes to screen only in shorter form, to log in the full form.
  in release mode no output to screen

- possibly convert some assert warnings to log warnig in Input and Profiler

- ASSERT - in debug version count calls and warn for large count:
  macro could expand to:
  static int count_LINENUM;
  feal::Assert( #expr, count_LINENUM)...

jbrezmorf:
The first part done with asserts. The second part with logger.

PR #572, opened 2016-05-05
Df comfortable asserts

dflanderka:
Part of #570 
 - rename FEAL_DEBUG_ASSERT macro
 - introduce comparative asserts


src/system/global_defs.h
```diff
@@ -152,50 +152,71 @@ if ( !(expr) ) \
```
jbrezmorf: It would be better to move all new assert macros into asserts.hh and include asserts.hh from global_defs.h. That would be temporary solution until we finish system classes.

src/system/global_defs.h
```diff
@@ -152,50 +152,71 @@ if ( !(expr) ) \
...
   feal::Assert( #expr).set_context( __FILE__, __func__, __LINE__)._FEAL_ASSERT_A
 #else
-#define FEAL_DEBUG_ASSERT( expr)
+#define FEAL_ASSERT_DBG( expr) \
+if ( !(expr) ) \
+  feal::AssertNull()._FEAL_ASSERT_A
 #endif
 
+/// Definition of comparative assert macro (Less Than)
+#define ASSERT_LT(a, b) \
```
jbrezmorf: How to deal with conflict with gtest? For every macro without FEAL_ prefix we should check if it's not already defined, 
like that:
	#ifdefined ASSERT_LT
	#ifdefined FEAL_OVERRIDE_ASSERTS
		#undef ASSERT_LT
	#else 
		#warning "ASSERT_LT already defined out of FEAL."
        #endif
        #endif

These checks could possibly be done through another macro.
CHECK_DEFINITION(ASSERT_LT, FEAL_OVERRIDE_ASSERTS)

Issue #571, opened 2016-05-04
Test python copy


jbrezmorf:
- [x] sort paths before PYTON_COPY  (done in 2.0.0_rc)
- [ ] Try to remove *.bat wrapper in windows, since python search in correct paths relative to the binary anyway. It was reported that when running flow123d.exe with corruped python libs with YAML file in the same directory the ImportError  does not appear. Possibly if the python libs are not found the load of the pythonlib just ignores it, but do not fail.
- [ ] DONE make a wrapper for unit tests similar to the flow123d wrapper, setting PYTHON environment
- [ ] DONE and a unit test for PythonLoader, that check that the most important  python modules are loaded from 
  the python copy
- [ ] DONE ?? use the debug flow123d wrapper for execution of the integration scripts, ?? it is essential for package testing, but do not test integrity of python copy since the system python is used anyway.
- [ ] JB: update package tests in order to test every built package, check that these test fails when the python copy is corrupted.

- [ ] create and publish installation manifest (list of files) for every built package

- [ ] try to remove unnecessary files from the packages: 
(flow123d.exe, flow123d_wrapper.bat), error on line 952
- [x] Try to link flow123d under Cygwin with non-cygwin python. Could be ultimate solution of python problems under windows. NO WAY.
- [x] How to set virtual environment for embedded python? Answer: works in the same way as the interpreter.
- [ ] Try to start cygwin app using embedded python form the wincmd through the cygwin shell. 
       Does it solve the issue with python libs?
       Can we set activate.sh within the cygwin shell.
- [ ] Could we make the cygwin shell standalone as we do with flow123d?

- [ ] Try to use virtual environment for our python. 
	Advantages:
	- simpler (automatic) copy of python environment, need to copy libpython* in addition
	- we can test consistency of our copy of the python environment (setting PYTHONPATH
           do not prevent usage of the system environment)
	Possible problems (we must test):
	- [ ] Can we call "activate" and setup environment out of Cygwin? 
	 Possibly not, so this force us to use non-cygwin python!! But check it, it seems that under windows "activate" is *.bat and just setup some variables.
	- [ ] Could we make a wrapper to call activate before flow123d call? In particular we have to call flow123d from the same directory as the wrapper itself. (see http://stackoverflow.com/questions/3827567/how-to-get-the-path-of-the-batch-script-in-windows)

jbrezmorf:
No way to make embedded Python work fine under cygwin with complex dependencies.
New approach is to adopt docker.

PR #575, opened 2016-05-10
JHy python env

x3mSpeedy:
Changes:
 - partial issue #571 fix
 - create pythonenv script which will inject python variables and execute commands
 - edit cmake unit test call using pythonenv
 - edit tests also using pythonenv script


jbrezmorf:
I postpone this pull until we decide about usage of Docker. That may make this pull unnecessary.

jbrezmorf:
Wrong base branch, close.

PR #576, opened 2016-05-11
Js thc coupling

### LINKED ISSUE #567: THC sequential coupling
Extend current coupling to compute both solute transport and heat.

jstebel:
Resolves #567 .


jstebel:
Already merged to another branch which was then merged into master.

PR #578, opened 2016-05-13
Df apply logger

dflanderka:


src/input/reader_to_storage.cc
```diff
@@ -632,7 +633,7 @@ StorageBase * ReaderToStorage::make_storage_from_default(const string &dflt_str,
```
jbrezmorf: Why we do not pass Type::Default instead of string.

src/input/reader_to_storage.cc
```diff
@@ -632,7 +633,7 @@ StorageBase * ReaderToStorage::make_storage_from_default(const string &dflt_str,
 
     } catch (Input::Type::ExcWrongDefault & e) {
         // message to distinguish exceptions thrown during Default value check at declaration
-        xprintf(Msg, "Wrong default value while reading an input stream:\n");
+    	e << Type::EI_Desc("Wrong default value while reading an input stream:\n");
```
jbrezmorf: Introduce new exception in case of invalid JSON format (in method Default::check_validity).
Here catch both exceptions, formate their what message and make a fatal error (either using assert or throwing an fatel exception directly). Add there what message of the original exception.

src/input/type_abstract.cc
```diff
@@ -143,7 +143,8 @@ bool Abstract::finish(bool is_generic) {
 		try {
 			child_data_->selection_default_.check_validity(child_data_->selection_of_childs);
 		} catch (ExcWrongDefault & e) {
-			xprintf(PrgErr, "Default value '%s' for TYPE key do not match any descendant of Abstract '%s'.\n", child_data_->selection_default_.value().c_str(), type_name().c_str());
+			ASSERT(false)(child_data_->selection_default_.value())(this->type_name())
```
jbrezmorf: Check if this is similar case to the previous one.

src/system/file_path.cc
```diff
@@ -109,7 +109,9 @@ void FilePath::substitute_value() {
     for (std::map<std::string,std::string>::const_iterator it = this->placeholder.begin(); it != this->placeholder.end(); ++it) {
         size_t i = abs_file_path_.find(it->first,0);
         if (i != std::string::npos) {
-            if (it->second == "" ) xprintf(Warn, "Substituting placeholder %s with empty value.\n", it->first.c_str());
+            if (it->second == "" ) {
```
jbrezmorf: ASSERT().warning()

src/input/reader_to_storage.cc
dflanderka: Type string is necessary. Method check_validity creates storage only for Default with value at declaration and and it is thus ensured. For other types of Default method "does nothing" (storage == null) and subsequently causes segfault.

jbrezmorf: Either I attached the comment to the wrong code or was unconscious, this code is pretty clear.

PR #591, opened 2016-06-10
Pe output

Paulie14:
- create output mesh identical with computational one
- introduce output element and its iterator
- pass error_control_field (future usage in refinement)
- compute field data on the output mesh
- unify writing data vectors in output_vtk
- get rid of mesh in output_vtk, use output mesh only

- class OutputMesh is still under construction (refinement)

### DIFF
--- a/src/io/output_element.hh
+++ b/src/io/output_element.hh
@@ -32,7 +32,7 @@ class ElementAccessor;
 /** @brief Represents an element of the output mesh.
  * Provides element access on the data of the output mesh (nodes, connectivity, offsets etc.).
  * 
- * Vertex function suppose spacedim = 3 at the moment.
+ * Vertex function suppose spacedim = 3 at the moment, hard coded.
  */
 class OutputElement
 {
@@ -43,7 +43,7 @@ public:
     typedef Space<spacedim>::Point Point;
     
     /// Constructor.
-    OutputElement(unsigned int ele_idx, OutputMesh* output_mesh);
+    OutputElement(unsigned int ele_idx, std::shared_ptr<OutputMeshBase> output_mesh);
     
     /// @name Output Mesh Getters.
     //@{ 
@@ -54,17 +54,7 @@ public:
     /// Returns global index of the node.
     unsigned int node_index(unsigned int loc_idx) const;
     Point vertex(unsigned int loc_idx) const;  ///< Returns coordinates of node @p loc_idx.
-    std::vector<Point> vertex_list() const;    ///< Returns vector of nodes coordinates.
-    
-    /// Returns global index of the node. (DISCONTINOUS)
-    unsigned int node_index_disc(unsigned int loc_idx) const;
-    
-    //TODO: vertex_disc and vertex_list_disc return the same coordinates as vertex and vertex_list
-    // It is meaningful only if we have ONLY discontinuous vectors for nodes and connectivity
-    /// Returns coordinates of node @p loc_idx. (DISCONTINOUS)
-    Point vertex_disc(unsigned int loc_idx) const;
-    /// Returns vector of nodes coordinates. (DISCONTINOUS)
-    std::vector<Point> vertex_list_disc() const;
+    std::vector<Point> vertex_list() const;    ///< Returns vector of nodes coordinates.    
     
     Point centre() const;                      ///< Computes the barycenter.
     //@}
@@ -79,33 +69,23 @@ public:
     ElementAccessor<spacedim> element_accessor() const;
     //@}
     
-    void operator++();
+    void inc();
     bool operator==(const OutputElement& other);
 private:
     
-    /// Returns global index of the node. (DISCONTINUOUS)
-    unsigned int node_index_internal(unsigned int loc_idx, 
-                                     std::shared_ptr<MeshData<unsigned int>> connectivity) const;
-    /// Returns coordinates of node @p loc_idx.  (DISCONTINUOUS)
-    Point vertex_internal(unsigned int loc_idx,
-                               std::shared_ptr<MeshData<unsigned int>> connectivity,
-                               std::shared_ptr<MeshData<double>> nodes) const;
-    /// Returns vector of nodes coordinates.  (DISCONTINUOUS)
-    std::vector<Point> vertex_list_internal(std::shared_ptr<MeshData<unsigned int>> connectivity,
-                                            std::shared_ptr<MeshData<double>> nodes) const;
-    
-//     friend void OutputElementIterator::operator++();
-    unsigned int ele_idx_;      ///< index of the output element
-    OutputMesh* output_mesh_;   ///< pointer to the output mesh
+    /// index of the output element
+    unsigned int ele_idx_;
+    /// pointer to the output mesh
+    std::shared_ptr<OutputMeshBase> output_mesh_;
 };
 
 // --------------------------------------------------- OutputElement INLINE implementation -------------------
 
-inline OutputElement::OutputElement(unsigned int ele_idx, OutputMesh* output_mesh)
+inline OutputElement::OutputElement(unsigned int ele_idx, std::shared_ptr<OutputMeshBase> output_mesh)
 : ele_idx_(ele_idx), output_mesh_(output_mesh)
 {}
 
-inline void OutputElement::operator++()
+inline void OutputElement::inc()
 {
     ele_idx_++;
 }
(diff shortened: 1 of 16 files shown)


src/io/output_mesh.hh
```diff
@@ -0,0 +1,123 @@
...
+    std::shared_ptr<MeshData<double>> discont_nodes_;
+    std::shared_ptr<MeshData<unsigned int>> discont_connectivity_;
+    
+    unsigned int n_nodes();
+    unsigned int n_nodes_disc();
+    unsigned int n_elements();
+    
+private:
+    
+    struct AuxElement{
```
jbrezmorf: If this structure is used only as a local variable in create_refined_mesh I suppose, that creating just the vectors would be simpler.

src/io/output_mesh.hh
```diff
@@ -0,0 +1,123 @@
...
+        std::vector<double> coords;
+        std::vector<unsigned int> connectivity;
+    };
+    
+    void fill_vectors();
+    bool refinement_criterion();
+    
+    Mesh *orig_mesh_;
+    bool discont_data_computed_;
```
jbrezmorf: We can just check size of discont data vectors, can't we?

src/io/output_time.cc
```diff
@@ -212,6 +256,9 @@ void OutputTime::write_time_frame()
 			// Remember the last time of writing to output stream
 			write_time = time;
 			current_step++;
+            
+            // unset validity of the output mesh after data are written at the current write time
+            is_output_mesh_valid_ = false;
```
jbrezmorf: Possibly better to just delete the mesh  here and check its pointer for validity.

src/io/output_time.cc
```diff
@@ -91,12 +102,45 @@ OutputTime::~OutputTime(void)
...
+    
+    FieldCommon* field =  output_fields->field(error_control_field_name);
+    if( field && (typeid(*field) == typeid(Field<3,FieldValue<3>::Scalar>)) ) {
+        Field<3,FieldValue<3>::Scalar>* error_control_field = static_cast<Field<3,FieldValue<3>::Scalar>*>(field);
+        DBGMSG("Output mesh will be refined according to field '%s'.\n", error_control_field_name.c_str());
+        // create refined output mesh according to the field
+        output_mesh_->create_refined_mesh(error_control_field);
+    }
+    else{
+        DBGMSG("Output mesh identical with the computational one.\n");
```
jbrezmorf: We should throw  an input exception since the input field must be scalar. Even better to do such check in the constructor just after read from input.

src/io/output_time.hh
```diff
@@ -230,6 +236,19 @@ protected:
      * Cached pointer at mesh used by this output stream
      */
     Mesh *_mesh;
+    
+    OutputMesh *output_mesh_;
```
jbrezmorf: Rather use std::shared_ptr.

src/io/output_time.impl.hh
```diff
@@ -310,64 +309,67 @@ void OutputTime::compute_field_data(DiscreteSpace space_type, Field<spacedim, Va
...
             output_data.zero(idx);
 
-        // sum values
-        FOR_ELEMENTS(this->_mesh, ele) {
-            FOR_ELEMENT_NODES(ele, i_node) {
-                Node * node = ele->node[i_node];
-                unsigned int ele_index = ele.index();
-                unsigned int node_index = this->_mesh->node_vector.index(ele->node[i_node]);
-
+        for(OutputElementIterator it = output_mesh_->begin(); it != output_mesh_->end(); ++it)
```
jbrezmorf: Could it be simplified to `for(auto & ele : *output_mesh_)`?

src/io/output_time.impl.hh
```diff
@@ -310,64 +309,67 @@ void OutputTime::compute_field_data(DiscreteSpace space_type, Field<spacedim, Va
...
+        {
+            std::vector<Space<3>::Point> vertices = it->vertex_list();
```
jbrezmorf: Seems to copy the vector of nodes

src/io/output_element.hh
```diff
@@ -0,0 +1,290 @@
...
+    Point res({0,0,0});
+    for(auto& v : vertex_list() ) res += v;
+    return res/n_nodes();
+}
+
+// --------------------------------------------------- OutputElementIterator ---------------------------------
+/** @brief Output element iterator.
+ * Provides iterator over elements of the output mesh.
+ */
+class OutputElementIterator
```
jbrezmorf: Could you create an iterator template for just any internal accessor type. That would be used both here and for SideIter.
Usage:
typedef GenericIterator<OutputElement> OutputElementIterator;

src/io/output_element.hh
```diff
@@ -0,0 +1,290 @@
...
+#include "mesh/point.hh"
+#include "output_mesh.hh"
+
+
+/** @brief Represents an element of the output mesh.
+ * Provides element access on the data of the output mesh (nodes, connectivity, offsets etc.).
+ * 
+ * Vertex function suppose spacedim = 3 at the moment.
+ */
+class OutputElement
```
jbrezmorf: I don't like two things in this class:
- duplicated interface for non-disc and disc cases
- creating vertex_list vectors and passing them by value

For the first issue either introduce aux objects for non-disc and disc mesh, usage example:

	output_mesh_->continuous_elements().begin()
	output_mesh_->discontinuous_elements().begin()

Or possibly create whole new  OutputMesh for the discontinus case which can share some date with the continuous mesh due to shared pointers (so e.g. it is safe to delete the continuous mesh).

For the second issue, there are three solutions:
- use our own allocators for those lists in order to share the contents with the big vector
- have our own proxy object for those lists and add new variant of the Field<>::value_list
- use vector of vectors to store element data and create the global arrays only on demand (filling the field values is called  more times then creating the mesh big arrays).

jbrezmorf:
After resolving the other minor issues the documentation of the classes and methods should be provided.

src/io/output_element.hh
Paulie14: The template parameter would then be required to implement a function that increases index (in our case inc() ) in operator++. Should not then the template parameter itself implement operator++?

Secondly, couldn't the implementation of operator ==, be general:
*this == ref_to_other_object, or is it necessary to be able to implement operator== differently for each usage?

src/io/output_time.cc
Paulie14: This flag is checked in make_output_mesh() which is called from field_set.output().
If the flag is not there, the output mesh would be constructed twice - once for flow, once for transport.
Since we have the same output mesh for both, I assumed we can do it once at the moment...

Paulie14: Sorry, not for flow and transport..
In case there are some reaction terms, then the output mesh would be constructed separately for sorption and dual porosity, since there are more field sets wrtiting into the same output stream...

src/io/output_time.cc
Paulie14: OK. Then I would rather suggest to create a simple record for output mesh inside output stream record. When we implement the refinement, I think it would good to be able to control some setting of the refinement (max levels, interpolation error tolerance, type of refinement - it might be useful to just set uniform refinement of given level instead of error control - occurred to me during testing refinement..)

src/io/output_element.hh
jbrezmorf: The template parameter need to implement:
- inc() method
- == operator
Only the second one has to be modified both in OutputElementIterator as in the SideIterator.

src/io/output_time.cc
jbrezmorf: I agree.

src/io/output_time.cc
jbrezmorf: I don't understand. You have:

transport output : null ouput mesh -> make it, do fields output
reactions output: mesh exists reuse it, output other fields
call write_time_frame, output everything and delete output_mesh

So the point is that I see no save in resolution of invalid and null mesh since both mean just make it from scratch.

Paulie14: OK, you're right.
I did not deleted the mesh in the right place...
So now since it is shared pointer, I hope "output_mesh_.reset()" is safe, isn't it?

src/io/output_element.hh
Paulie14: I solved the first issue.
The second one:
To be honest, I don't understand much the first two options you named :(
And I don't like the third one either.. we would 3 version of the mesh saved side by side at some times...

PR #598, opened 2016-06-22
Df fix input paths

dflanderka:


src/system/file_path.hh
```diff
@@ -76,11 +76,11 @@ public:
     /**
      * Set:
      * - working directory (used only if the output directory is relative)
-     * - root directory (of the main input file)
+     * - input root directory (path of the main input file)
      * - input directory to replace ${INPUT} place holder
      * - output directory used as prefix to the output files (relative output dirs are relative to the working directory)
      */
-    static void set_io_dirs(const string working_dir, const string root_input,const string input,const string output);
+    static string set_io_dirs(const string main_input_file, const string input, const string output, const string working_dir=".");
```
jbrezmorf: Both output and input directory, if specified as relative should be relative to the root directory of the main_input_file.
Then we can possibly remove working directory completely. In order to simplify tests, where one may use several input files and setting non-existent main yaml file is awkward, I suggest to keep several set_io_* methods, see next comment.

jbrezmorf:
```
// for backward compatibility, need not to change all tests
// Have assert for working_dir=="." and just ignore it.
set_io_dirs( working_dir, root_dir, input, output)
// this do the main thing
set_dirs(root_dir, input, output)
// this calls previous internally
set_dirs_from_input(main_yaml, input, output)
```

jbrezmorf: Update documentation, in particular the return value doc.

src/system/file_path.cc
```diff
@@ -40,7 +40,7 @@ FilePath::FilePath(string file_path, const  FileType ft)
     if (ft == input_file) {
     	if ( FilePath::is_absolute_path(abs_file_path_) ) {
     	} else {
-            abs_file_path_ = root_dir + DIR_DELIMITER + abs_file_path_;
+            abs_file_path_ = input_root_dir + DIR_DELIMITER + abs_file_path_;
```
jbrezmorf: Does boost have a portable way of path joining? I.e. could we use boost to get rid of the DIR_DELIMITER, it is set constantly to "/" anyway. 
Even if we keep current approach we should move it from system into class FilePath and possibly make it private (in future) there are few places where it is used out of FilePath and some of them are not too correct, e.g. the match method of Input::Type::FilePath.

src/system/file_path.cc
```diff
@@ -40,7 +40,7 @@ FilePath::FilePath(string file_path, const  FileType ft)
...
     	}
     } else if (ft == output_file) {
         if ( FilePath::is_absolute_path(abs_file_path_) ) {
```
jbrezmorf: Possibly use boost to separate the file name and check the match parent_type() == output_dir.

src/system/file_path.cc
```diff
@@ -57,45 +57,43 @@ FilePath::FilePath(string file_path, const  FileType ft)
 
 
 
-void FilePath::set_io_dirs(const string working_dir, const string root_input_dir,const string input,const string output) {
-    // root directory
-    root_dir = root_input_dir;
-
+string FilePath::set_io_dirs(const string main_input_file, const string input, const string output, const string working_dir) {
     // relative output dir is relative to working directory
     // this is possibly independent of position of the main input file
```
jbrezmorf: Yeah, bad decision in the past. Taking known use cases I see no reason to interpret relative output dirs differently then input dirs. This only leads to confusion.

src/system/file_path.cc
jbrezmorf: boost way is so cute:
fs::path full_path = dir / file;

src/system/file_path.cc
```diff
@@ -57,45 +57,43 @@ FilePath::FilePath(string file_path, const  FileType ft)
...
-    	for (vector<string>::iterator it = dirs.begin(); it != dirs.end(); ++it) {
-    	    if ( !(*it).size() ) continue;
-    	    if ( !output_dir.size() ) output_dir = *it;
-    	    else output_dir = output_dir + DIR_DELIMITER + *it;
-    	    FilePath::create_dir(output_dir);
-        }
+    	full_output_path = boost::filesystem::path(working_dir + DIR_DELIMITER + output);
+    }
+    boost::filesystem::create_directories(full_output_path);
+    output_dir = full_output_path.string();
```
jbrezmorf: Why don't store all directories as boost::filepath objects?

src/system/file_path.cc
```diff
@@ -57,45 +57,43 @@ FilePath::FilePath(string file_path, const  FileType ft)
...
 
-    	FilePath::create_canonical_path(working_dir, output);
+    if ( !FilePath::is_absolute_path(output_dir) ) {
```
jbrezmorf: Add comment that we are treating relative working_dir (resp. root_dir) here.

src/system/file_path.cc
```diff
@@ -57,45 +57,43 @@ FilePath::FilePath(string file_path, const  FileType ft)
...
     // the relative input is relative to the directory of the main input file
     add_placeholder("${INPUT}", input);
+
+    // set input root directory
+	boost::filesystem::path input_path;
+	if ( FilePath::is_absolute_path(main_input_file) ) {
+    	input_path = boost::filesystem::path(main_input_file);
+    } else {
+    	stringstream dir;
+    	dir << "." << DIR_DELIMITER << main_input_file;
```
jbrezmorf: There are better ways to do string join:

src/system/file_path.cc
```diff
@@ -40,7 +40,7 @@ FilePath::FilePath(string file_path, const  FileType ft)
     if (ft == input_file) {
     	if ( FilePath::is_absolute_path(abs_file_path_) ) {
```
jbrezmorf: Could we implement FilePath::is_absolute_path using boost::filesystem::path.is_absolute()? Just to be as consistent as possible. Then some calls of this method may be replaced by direct call of the boost is_absolute. E.g. consider

	FilePath::is_absolute_path(output_dir)
vs.
	output_path.is_absolute()

src/system/file_path.cc
```diff
@@ -57,45 +57,43 @@ FilePath::FilePath(string file_path, const  FileType ft)
...
+	boost::filesystem::path input_path;
+	if ( FilePath::is_absolute_path(main_input_file) ) {
+    	input_path = boost::filesystem::path(main_input_file);
+    } else {
+    	stringstream dir;
+    	dir << "." << DIR_DELIMITER << main_input_file;
+    	input_path = boost::filesystem::path(dir.str());
+    }
+	input_root_dir = input_path.parent_path().string();
+    return input_path.filename().string();
```
jbrezmorf: I wonder how it did happen to work without returning converted main_path here and use it in Application. Good job.

unit_tests/system/file_path_test.cpp
```diff
@@ -17,6 +17,116 @@
...
+			boost::filesystem::create_directories(dir.str());
+			EXPECT_TRUE( boost::filesystem::is_directory(dir.str()) );
+			boost::filesystem::remove(dir.str());
+			EXPECT_FALSE( boost::filesystem::is_directory(dir.str()) );
+		}
+	}
+
+	// absolute path contains actual dir ('.'), format is current_path() + "/./x"
+	{
+		stringstream dir;
```
jbrezmorf: Either this is the test of create_directories or the test of path joining, which should not be done using redir operators.
Just compare resulting path to a string, or if you are not sure if this is portable implement a single checking method. 
It should be clear what is the reference value or correct behavior of the test. So do not mix tow tests together.

src/system/file_path.hh
```diff
@@ -19,6 +19,7 @@
 #define FILE_NAME_HH_
 
 #include <string>
+#include <boost/filesystem.hpp>
```
jbrezmorf: We have to avoid this include in the header. Move implementation to the cc file, even for the opeartor, inlining is not necessary here.

src/system/file_path.cc
```diff
@@ -16,123 +16,110 @@
...
+	root_dir = boost::filesystem::path(root);
+
+	// set output directory
+	// if param output is absolute output_dir = output
+	// else output_dir = root / output
+	// the resulting relative path is always completed to absulute path
+	boost::filesystem::path output_path = boost::filesystem::path(output);
+    if ( !output_path.is_absolute() ) {
+    	boost::filesystem::path root_output_path = boost::filesystem::path(root) / output_path;
+    	boost::filesystem::create_directories(root_output_path);
```
jbrezmorf: This has to be done for `output_dir` after it is definitely set.

src/system/file_path.cc
```diff
@@ -16,123 +16,110 @@
...
-#else
-    	    	output_dir = DIR_DELIMITER + *it;
-#endif // FLOW123D_HAVE_CYGWIN
-    	    } else {
-            	output_dir = output_dir + DIR_DELIMITER + *it;
-    	    }
-    	    FilePath::create_dir(output_dir);
-        }
+	root_dir = boost::filesystem::path(root);
+
```
jbrezmorf: Make canonical absolute path for the root_dir here.
1. It is necessary to prevent warning in `FilePath` constructor if the root is ".",
2. Simplifies further code for setting `output_dir`

src/system/file_path.hh
dflanderka: We can move implementation of operators to cc file, but we can't remove include, because FilePath class contains boost::filesystem::path class members.

src/system/file_path.cc
dflanderka: Method boost::filesystem::canonical needs existing directory, create_directories must be call before canonical.

src/system/file_path.hh
jbrezmorf: This include is really huge. To avoid it:
1. make output_path etc. shared_ptr<boost::fielsystem::path> and use make_shared for initialization 
2. use just forward declaration of boost::filesystem::path i.e.
	namespace boost {
		namespace filesystem  {
			class path
	}}

src/system/file_path.cc
jbrezmorf: Fine., just 
1. set e.g full_output_path for all cases
2. call create_directories for full_output_path
3. set output_dir to canonical (full_output_path)

The point is to call both `create_dirs` and `canonical` for all cases.

src/system/file_path.cc
jbrezmorf: The root dir should exist so canonical should work here. Otherwise throw.

PR #602, opened 2016-06-23
JHy python copy fix

x3mSpeedy:
Proposed new Python structure install change in CMake


config.cmake.template
```diff
@@ -50,13 +50,48 @@
...
+#             correctly, example:
+#             set(PYTHON_COPY_PACKAGES
+#                 bsddb compiler ctypes curses distutils email encodings hotshot importlib 
+#                 json lib-dynload lib-tk lib2to3 logging multiprocessing 
+#                 plat-x86_64-linux-gnu pydoc_data sqlite3 test unittest wsgiref xml)
+#             
+#   - PYTHON_COPY_DIST_PACKAGES
+#           - list of all 3rd party packages (installed using pip or other tool)
+#             each entry is either
+#                - relative path:
```
jbrezmorf: Can we restrict this list to be a list of paths relative to ${PYTHON_ROOT}/dist-packages ?

config.cmake.template
```diff
@@ -50,13 +50,48 @@
...
-# USE_PYTHON - only use python if this is set to "yes"
+# USE_PYTHON - use embedded python if this is set to "yes"
 # We add python library only on explicit request since it leads to tons of errors when debugging with valgrind.
 #
 # set(USE_PYTHON "yes")
 
 # PYTHON_COPY - make copy of python files, necessary for release builds
 # set(PYTHON_COPY "yes")
+#  - Following variables MUST ALSO BE SET:
+#    - PYTHON_COPY_ROOT
```
jbrezmorf: Definitely the python root should be same for the build as for the copy, so this is in fact HINT for FindPackage(Python) so we should possibly rename this to just PYTHON_ROOT.

config.cmake.template
```diff
@@ -50,13 +50,48 @@
...
+#                - relative path:
+#                   - relativity is taken towards PYTHON_COPY_ROOT
+#                   - all relativity is kept during copy so dist-packages/yaml
+#                     will create folder dist-packages/yaml
+#                - absolute path:
+#                   - only last 2 folders are taken into account for relativity
+#                     (we cannot conclusively estimate where relativity begins)
+#                     e.g. /usr/local/lib/dist-packages/markdown
+#                     will create folder dist-packages/markdown
+# - PYTHON_SYSPATH
```
jbrezmorf: This should be filled automatically since only the copy mechanism knows what is where. An reason why we can not make this automatic?

config.cmake.template
```diff
@@ -50,13 +50,48 @@
...
+#                 plat-x86_64-linux-gnu pydoc_data sqlite3 test unittest wsgiref xml)
+#             
+#   - PYTHON_COPY_DIST_PACKAGES
+#           - list of all 3rd party packages (installed using pip or other tool)
+#             each entry is either
+#                - relative path:
+#                   - relativity is taken towards PYTHON_COPY_ROOT
+#                   - all relativity is kept during copy so dist-packages/yaml
+#                     will create folder dist-packages/yaml
+#                - absolute path:
```
jbrezmorf: Is there any reason for absolute paths?

config.cmake.template
x3mSpeedy: It depends on platform. Some tools install packages to `dist-packages` but some to `site-packages`. So we cannot detect it reliably.

config.cmake.template
x3mSpeedy: Rename can be done. But hint cannot be used. There are two Python related FindPackages
- [`find_package(PythonLibs)`](https://cmake.org/cmake/help/v3.0/module/FindPythonLibs.html) - which finds include files and libraries (so)
- [`find_package(PythonInterp)`](https://cmake.org/cmake/help/v3.0/module/FindPythonInterp.html) - which interpret binary (e.g. python command)

`PythonLibs` (includes and libs) as well as `PythonInterp` can be outside PYTHON_ROOT

config.cmake.template
x3mSpeedy: Unfortunately it depends partially also on platform. We can deal with 3rd party packages however base modules can vary. On one platform base modules include path 
`/usr/lib/python2.7/plat-x86_64-linux-gnu`
but on other there is this path instead:
`/usr/lib/python2.7/plat-linux2`

What we can do is parse python paths (`sys.path`) a remove PYTHON_ROOT from them (if it contains) but if path will come from outside of PYTHON_ROOT we may not be able to use it.
Part of Python resides in:
`/usr/lib/python2.7/`
but some part may reside in
`/usr/lib/python2.7/`

config.cmake.template
x3mSpeedy: There is only one PYTHON_ROOT but some packages are located outside root.

x3mSpeedy:
@jbrezmorf right now we are dealing with 3rd party packages twice. 

1. in `PythonCopy` we can specify which package we want to use
2. in PythonScripts we use `pip` tool to install packages we want to use.

I suggest to use only `pip`, so we can control what to install and where to install.
- Downside of this approach is that some packages will be installed even though they are installed in system already. (Right now `pip` install mechanism skips packages already installed so they are installed once per build).
- Another possible downside is we cannot use packages which cannot be installed using `pip` (right now there are none)

config.cmake.template
jbrezmorf: OK. Leave the name, and explain how it is used. I guess that this path is stripped from other paths and resulting relative path is recreated under the build_tree.

PR #603, opened 2016-06-23
Js fix balance richards

jstebel:


jbrezmorf:
Seems OK, but test 10 and test 15 reports different water_balance (was OK before). Please investigate that.

jstebel:
Done.

PR #607, opened 2016-06-30
Df finish exceptions

dflanderka:


src/system/exceptions.hh
```diff
@@ -321,6 +322,37 @@ TYPEDEF_ERR_INFO( EI_XprintfMessage, std::string);
...
+
+
+/**
+ * Create shared pointer of given exception.
+ *
+ * Used for propagation exception message.
+ */
+template <class Exc>
+std::shared_ptr<Exc> make_exception_ei(Exc &e) {
+    return std::make_shared<Exc>(e);
```
jbrezmorf: check than Exc is derived from ExceptionBase using 
[type traits](http://en.cppreference.com/w/cpp/header/type_traits)
in combination with STATIC_ASSERT

src/input/type_record.cc
```diff
@@ -76,9 +76,11 @@ bool Default::check_validity(std::shared_ptr<TypeBase> type) const
 		storage_ = reader.get_storage()->get_item(0);
 		return true;
 	} catch ( Input::ReaderToStorage::ExcNotJSONFormat &e ) {
-		THROW( ExcWrongDefaultJSON() << EI_DefaultStr( value_ ) << EI_TypeName(type->type_name()));
+		THROW( ExcWrongDefaultJSON() << EI_DefaultStr( value_ ) << EI_TypeName(type->type_name())
+				<< EI_Nested(make_exception_ei<Input::ReaderToStorage::ExcNotJSONFormat>(e)) );
 	} catch ( Input::ReaderToStorage::ExcInputError &e ) {
-		THROW( ExcWrongDefault() << EI_DefaultStr( value_ ) << EI_TypeName(type->type_name()));
+		THROW( ExcWrongDefault() << EI_DefaultStr( value_ ) << EI_TypeName(type->type_name())
+				<< EI_Nested(make_exception_ei<Input::ReaderToStorage::ExcInputError>(e)) );
```
jbrezmorf: make_exception_ei should directly return EI_Nested.
test that template parameter need not to be specified.

PR #610, opened 2016-07-02
Df logger formater

dflanderka:


src/system/logger.hh
```diff
@@ -142,6 +145,13 @@ public:
 	/// Set flag every_process_ to true
 	Logger& every_proc();
 
+	/// Allow use functionality of fmtlib for formating message.
```
jbrezmorf: Document possibilities of the new formatting in the main documentation of the logging features. Provide examples.

PR #613, opened 2016-07-08
Df empty record

dflanderka:


src/input/reader_to_storage.cc
```diff
@@ -140,9 +140,31 @@ StorageBase * ReaderToStorage::make_storage(PathBase &p, const Type::TypeBase *t
...
+    			} else { // auto conversion
+    				return make_storage_empty_record(p, abstract_record_type->get_default_descendant());
+    			}
+    		} else {
+    			return make_storage_empty_record(p, &( abstract_record_type->get_descendant(descendant_name) ));
+    		}
+    	}
+
+    	// return Null storage if there is null on the current location
+		if (p.is_null_type()) {
```
jbrezmorf: Test this before scalar types resolution.

src/input/reader_to_storage.cc
```diff
@@ -140,9 +140,31 @@ StorageBase * ReaderToStorage::make_storage(PathBase &p, const Type::TypeBase *t
         return storage;
     }
 
-    // return Null storage if there is null on the current location
-    if (p.is_null_type())
-        return new StorageNull();
+    // create storage if Record is empty
+	if (typeid(*type) == typeid(Type::Record) && p.is_effectively_null()) {
```
jbrezmorf: Move this into specific make_storage methods.

src/input/reader_to_storage.cc
```diff
@@ -140,9 +140,31 @@ StorageBase * ReaderToStorage::make_storage(PathBase &p, const Type::TypeBase *t
...
+		return make_storage_empty_record(p, static_cast<const Type::Record *>(type));
```
jbrezmorf: No need for specific make_storage for empty record. Just check that path.get_record_key_set returns true and set empty set for null path.

src/input/reader_to_storage.hh
```diff
@@ -22,6 +22,7 @@
 #include <sstream>
 
 #include "input/input_type_forward.hh"
+#include "input/type_record.hh"
```
jbrezmorf: Remove include with new methods.

unit_tests/input/reader_to_storage_test.cpp
```diff
@@ -878,6 +879,106 @@ TEST_F(InputReaderToStorageTest, Tuple) {
...
+        .close();
+
+    static Type::Record descA_rec = Type::Record( "RecordA", "desc.")
+        .derive_from(a_rec)
+		.copy_keys(inrec)
+	    .close();
+
+    static Type::Record root_rec_with_abstract = Type::Record( "RootRecordWithAbstract", "desc.")
+        .declare_key("format", a_rec, Type::Default::obligatory(), "")
+	    .close();
```
jbrezmorf: Add check for null scalar type key, should be same as if key is not present.

PR #645, opened 2016-07-28
J hy python scripts

jbrezmorf:
Scripts for new test scheme.


CMake/Modules/python/PythonLibs.cmake
```diff
@@ -11,7 +11,6 @@
 find_package(PythonLibs 2.7 REQUIRED)
 
 include_directories(${PYTHON_INCLUDE_DIRS})
-flow_define(HAVE_PYTHON)
```
jbrezmorf: Flow still depends on this define. Keep it for now.

CMake/Modules/python/PythonScripts.cmake
```diff
@@ -24,9 +24,21 @@ macro(install_python_lib package_name package_dir)
         message(STATUS "-- cd ${Flow123d_SOURCE_DIR}/CMake/Modules/python && pip install --target=${PYTHON_3RD_PARTY} ${package_name}")
         # we start pip install from directory where setup.cfg is placed thus avoiding issue with --target usage
         # https://github.com/pypa/pip/pull/3450 and https://github.com/pypa/pip/issues/3056
+        # for now we try to install package using this fix (platform centos)
         execute_process(
             COMMAND pip install --target=${PYTHON_3RD_PARTY} ${package_name}
-            WORKING_DIRECTORY ${Flow123d_SOURCE_DIR}/CMake/Modules/python)
+            WORKING_DIRECTORY ${Flow123d_SOURCE_DIR}/CMake/Modules/python
+            RESULT_VARIABLE RETURNCODE)
```
jbrezmorf: What is difference here? Pip should not depend on working direcotry.

unit_tests/system/python_loader_test.cpp
```diff
@@ -49,23 +49,29 @@ TEST(PythonLoader, print_error) {
         "Message: global name 'a' is not defined\nTraceback");
 }
 
+
```
jbrezmorf: We should allow build and run of flow only with packages that are absolutely necessary.

bin/python/exec_parallel.py
```diff
@@ -0,0 +1,78 @@
...
+    '(using platform.node() or socket.gethostname()) and then search',
+    'it in the table "host_table.json" which assign logical hostname',
+    'possibly to multiple different real hostnames. ',
+    '',
+    'If the real host name is not found in the table ',
+    'it is used directly otherwise the logical ',
+    'hostname is used to select the setup script.',
+])
+parser.add('-v', '--valgrind', type=[True, str], name='valgrind', placeholder='[<VALGRIND ARGS>]', docs=[
+    'Run tests under valgrind, with python suppression with optional argument',
```
jbrezmorf: Not only tests.

bin/python/exec_parallel.py
```diff
@@ -0,0 +1,78 @@
...
+    ' <valgrind args> passed to the valgrind. (In PBS mode this arguments is ignored.)?',
+])
+parser.add('', '--batch', type=True, name='batch', docs=[
+    'Make output of this script more for an off-line reading',
```
jbrezmorf: I do not understand the doc.

bin/python/exec_parallel.py
```diff
@@ -0,0 +1,78 @@
...
+parser.add_section('Passable arguments to exec_with_limit.py')
+parser.add('-t', '--limit-time', type=Duration.parse, name='time_limit', placeholder='<time>', docs=[
+    'Obligatory wall clock time limit for execution in seconds or in HH:MM:SS format.',
+    'For precision use float value'
+])
+parser.add('-m', '--limit-memory', type=float, name='memory_limit', placeholder='<memory>', docs=[
+    'Optional memory limit per node in MB',
+    'For precision use float value'
+])
+parser.add('', '--root', hidden=True, type=str, name='root', placeholder='<ROOT>', docs=[
```
jbrezmorf: Is it necessary if this tool is for any executable?

bin/python/host_table.json
```diff
@@ -0,0 +1,6 @@
+{
```
jbrezmorf: We should move this into config dir. Possibly use YAML format namely to add comment with explanation of purpose of this file.

src/python/scripts/core/base.py
```diff
@@ -0,0 +1,463 @@
+#!/usr/bin/python
```
jbrezmorf: For all python sources: document at least purpose of every class.

src/python/scripts/exec_parallel_module.py
```diff
@@ -0,0 +1,250 @@
...
+
+        template = PBSModule.format(
+            module.template,
+            command=command,
+            json_output=case.fs.json_output # TODO remove
+        )
+
+        return template
+
+    def run_local_mode_one(self, proc):
```
jbrezmorf: document difference against run_local_mode

src/python/scripts/pbs/modules/pbs_hydra_kai_tul_cz.py
```diff
@@ -0,0 +1,71 @@
...
+    def __init__(self, job_id, case):
+        super(ModuleJob, self).__init__(job_id, case)
+        self.parser = self.parser_builder(
+            self, 4, JobState.UNKNOWN,
+            queue=7,
+            name=2,
+        )
+
+    @classmethod
+    def update_command(cls):
```
jbrezmorf: Seems that interface classes  to various PBS systems only define commands however parsing of the output should be also specific.

jbrezmorf:
TODO:
1. merge master
2. add doc for classes
3. Is comparison finished?
4. Some test of PBS specific  interface. So that we can run it on a particular parallel machine and be shure that it works. Need to test:
- can execute serial  job (np 1) under PBS
- can execute parallel job (np 2) under PBS
- can handle error in one process of the parallel job
- can handle job rejected by PBS (e.g. wrong queue)
- can handle job waiting to long for execution (How we treat this case?) 

Further discussion necessary.

CMake/Modules/python/PythonLibs.cmake
x3mSpeedy: `flow_define(HAVE_PYTHON)` is still present in `CMakeLists.txt` no change required

CMake/Modules/python/PythonScripts.cmake
x3mSpeedy: Its due to bug. In working directory is present special configuration file which `pip` recognize. Other solution is to copy this file to current working directory.

unit_tests/system/python_loader_test.cpp
x3mSpeedy: Do you mean during Python copy? Packed Python packages folder has `21 MB` I don't see any big benefit by reducing size even more. Further discussion may be required on this topic.

bin/python/exec_parallel.py
x3mSpeedy: It is for any executable but host_table.json is needed in fixed location. But will remove.

bin/python/exec_parallel.py
x3mSpeedy: Fixed

bin/python/exec_parallel.py
x3mSpeedy: Add more notes on this flag

unit_tests/system/python_loader_test.cpp
```diff
@@ -49,23 +49,29 @@ TEST(PythonLoader, print_error) {
...
+// only test embedded python if we actually copied out python
```
jbrezmorf: Change comment to Test that necessary python libraries are available in embedded python.

PR #648, opened 2016-07-29
Js sorption dg for master

jstebel:
Cherry-picked linear sorption in TransportDG from version 1.8.3.
Test file and reference results updated, manually checked to give the same results.


jbrezmorf:
Failing tests 16 and 03. In the test 03 just ignore difference in solute_observe.yaml.

jstebel:
Fixed.

PR #687, opened 2016-09-02
Pe intersection

jbrezmorf:


src/intersection/compute_intersection.cc
```diff
@@ -0,0 +1,1139 @@
...
+        
+        IP.set_coordinates(local_abscissa,local_triangle);
+        return true; // IP found
+    }
+    
+    return false;   // IP NOT found
+}
+
+
+bool ComputeIntersection<Simplex<1>, Simplex<2>>::compute(std::vector<IntersectionPointAux<1,2>> &IP12s,
```
jbrezmorf: I propose to distinguish:
- special case, when just 1 or 2 products are zero, this mean we have regular intersection but just on the boundary (edge, vertex). So in this case we comput as in the regular case and just mark topology info.
- degenerate case, when all 3 are zero, so the line and triangle are coplanar

jbrezmorf: When 1d-2d is called as part of 1d-3d or 2d-3d, we can always skip degenerate case since we get also special case
intersections for other faces. The only exception is 2d-2d case with coplanar triangles.

jbrezmorf: So, I suggest to split the compute method, store auxiliary variables into class as well as the resulting IntersectionPointAux and have ability to call the exact methods we want:
- compute_regular() // check  signs, for regular and special cases set the intersection point and its topology, return code should indicate "number of intersections" (0, 1, 2=degenerate)
- compute_degenerate()
- compute_final()
- intersection_point(uint index = 0) // return the intersection point reference,

src/intersection/compute_intersection.cc
```diff
@@ -0,0 +1,1139 @@
...
+    if(t >= -geometry_epsilon && t <= 1+geometry_epsilon){
+
+        IP.set_orientation(2);  // set orientation as a pathologic case ( > 1)
+        // possibly set abscissa vertex {0,1}
+        if( fabs(s) <= geometry_epsilon)       { s = 0; IP.set_topology_A(0,0);}
+        else if(fabs(1-s) <= geometry_epsilon) { s = 1; IP.set_topology_A(1,0);}
+        else                         {        IP.set_topology_A(0,1);}   // no vertex, line 0, dim = 1
+        
+        // possibly set triangle vertex {0,1,2}
+        if( fabs(t) <= geometry_epsilon)       { t = 0; IP.set_topology_B(RefElement<2>::interact(Interaction<0,1>(side))[RefElement<2>::normal_orientation(side)],0);}
```
jbrezmorf: Why we do "rounding" of the t parameter? I suggest to only set topology.

src/intersection/compute_intersection.cc
jbrezmorf: Partially done in JB_intersection branch.

src/intersection/compute_intersection.cc
```diff
@@ -0,0 +1,1139 @@
...
+            if( IP.dim_A() == 0 ) // if IP is vertex of triangle
+            {
+                // we are on line of the triangle, and IP.idx_A contains local node of the line
+                // E-E, we know vertex index
+                IP23.set_topology_A(RefElement<2>::interact(Interaction<0,1>(triangle_line))[IP.idx_A()], 0);
+            
+                if( IP.dim_B() < 3 )
+                    // if IP is on the surface of the tetrahedron,
+                    // we cannot trace it optimally
+                    intersection.pathologic_ = true;
```
jbrezmorf: It seems that IntersectionAux<>::pathologic_ flag is never used neither through is_pathologic() method. If we do not use this flag we can also drop IntersectionPointAux::is_pathologic() and setting orientation to  2 in 1d-2d degenerate case.

src/intersection/compute_intersection.cc
jbrezmorf: Do we even need to detect boundary of the line segment here? It seems useless in auxiliary 1d-2d intersection.

Paulie14: This is the place where we "snap" the IP to properly set topology info (e.i. almost node -> node);
Topology on the line is tested e.g. in 2d-2d.
Later tests suppose, that the almost IP are already snapped (coming from 1d-2d).
Honestly, if the coordinates are not rounded here, I do not know how the further test would behave.. I think at least that we would have to add tolerance to other tests..

src/intersection/compute_intersection.cc
Paulie14: As I commented in the code..
I found out, that in this case the optimal tracing does not work. To be consistent with the naming,
I set it pathologic, so the tracing algorithm could go to convex hull alg.

src/intersection/compute_intersection.cc
jbrezmorf: 2d-2d use compute_final. How every other call of 1d-2d compute is followed by trimming intersection to the abscissa limits. so we should move that check into 1d-2d intersection object. Then the trimming will be done always and calculation of the degenerate case only in 2d-2d.

src/intersection/compute_intersection.cc
jbrezmorf: I hope that we can improve optimal tracing algo to cover correctly all special cases, then we need not to use convex_hull.  However the tracing must be part of the 2d-3d intersection class.

jbrezmorf:
Could you specify what exactly you need in prolongation table. How you treat special cases where edge or vertex are part of the intersection boundary? Why there is no prolongation table for 1d-3d case?

jbrezmorf:
- [ ] remove test_units
- [ ] remove unit_tests/mesh/old_VF
- [ ] merge ref_element fixes from JB_intersections

Paulie14:
Removing done in PE_prolong.
Pe_intersection unit test fail due to commits 01538c2 and 6ad7bbe. 
After you merge JB_intersections -> PE_intersection and it is working, we can merge PE_prolong -> PE_intersections.
(PE_prolong includes unified prolongation for 2d-3d without prolongation table.)

PR #697, opened 2016-09-16
Pe assembly

Paulie14:
Suggestion of new assembly in Darcy Flow.

problems:
- preallocation is done by the assembly functions - hard to read and debug..
- new LocalSystem is dense at the moment - since we write it whole into linsys,
  more entries with zeros are added.
- test03 - test_output.yaml - the results are different in order of rounding errors, where the error in observe points comes from, since the numbers are the same...
- test 11,12 with simple 3tetra meshes with Dirichlet BC have errors in preallocation
  (when not calling schur_allocation in darcy_assembler, computation on 1 proc. works, but on more it gets stuck on balance->start**\* functions, I do not understand)
- test10 - richards_lmh.yaml - there is only total flux bc in contrast to the other test, but I failed to find a problem... (the solution is constant in time..)
- test01 - bddc test gets stuck on more than 1 proc.; I do not known why

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,6 +49,7 @@ doc/reference_manual/flow_version.tex
 .settings
 
 *.pyc
+*python.log
 /.idea/
 bin/pythonenv.sh
 tests/runtest.sh
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -147,6 +147,7 @@ add_library(la_lib
     la/linsys_BDDC.cc
     la/linsys_PETSC.cc
     la/sparse_graph.cc
+    la/local_system.cc
 )
 target_link_libraries(la_lib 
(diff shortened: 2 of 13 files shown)


jbrezmorf:
When things are messy (and Darcy is messy right now .. sorry) you must go in small steps, otherwise its nightmare to debug the changed code. So possibly create new branch and merge there single commits or even break them into smaller steps ...

It is possible that preallocations are wrong with dense LocalSystem and I'm not sure the Schur complement detect corectly local matrices in block A to compute its inverse.
1. Make separate pass through for preallocation just preallocating whole local system. Keep the old assembly. Test completely.
2.  Assembly thorugh LocalSystem in old assembly method.
3. Have Local system in EqData and move assembly into new framework bit by bit.

PR #701, opened 2016-09-20
Simplification of auxiliary 1D-2D intersection.

jbrezmorf:
- add macro for comfortable debug variable output
- fix 1D-2D and 1D-3D
- seems there are still some problems in 2D-3D tracking


src/intersection/compute_intersection.cc
```diff
@@ -296,26 +298,45 @@ bool ComputeIntersection<Simplex<1>, Simplex<2>>::compute(std::vector<Intersecti
...
         
         if(compute_plucker(IP, w))
         {
+            // edge of triangle
+            if (n_both == 2) // one zero product
+                IP.set_topology_B(zero_idx_sum, 1);
+            else if (n_both == 1) // two zero products
+                IP.set_topology_B(RefElement<2>::oposite_node(3-zero_idx_sum), 0);
+
+            DebugOut().VarFmt(n_positive).VarFmt(n_negative);
```
Paulie14: I do not get, how zero_idx_sum is index of triangle side.
The same with the oposite node and set_orientation is not obvious for me..

src/intersection/compute_intersection.cc
```diff
@@ -304,44 +304,45 @@ bool ComputeIntersection<Simplex<1>, Simplex<2>>::compute(std::vector<Intersecti
...
         if (w[i] > rounding_epsilon) n_positive++;
-        else if (w[i] < -rounding_epsilon) n_negative++;
-        else zero_idx_sum+=i;
-        //DebugOut().fmt("{} np: {} nn: {}\n", w[i], n_positive, n_negative);
+        else if ( w[i] > -rounding_epsilon) zero_idx_sum+=i;
+        else n_negative++;
     }
 
-    if (n_positive>0 && n_negative>0) return false;
+    if (n_negative>0) return false;
```
Paulie14: what if all are negative??

jbrezmorf: ..in comment.

src/intersection/compute_intersection.cc
Paulie14: There is no change.. Did you forget to push the comments?

Paulie14:
(sorry, wrong "review" comment before.. it's new here..)
Setting topology info: tracing VS prolongation:
PE_intersection
- the topology info is set as precise as it can be - node,edge,face..
- in prolongation (both 1d-2d-3d) we need this precise info, so we can cover all neighboring elements: e.g. if IP is at edge of tetra, we need to look at two neighbors, if at node, then 3 neighbors..
- in tracing alg. we did not suppose pathologic cases; therefore IP.dim_B is always 3 (IP is inside tetra) or 2 (IP inside tetras face)
- 2d-3d prolongation table is built during tracing; prolongation is done only by a triangle side or a tetra face; prolongation over an edge or a node is missing at the moment (one of the todos in the review issue..)
- 1d-3d prolongation is done without prolongation table (the one that is passed is empty); all cases are covered (to my best knowledge and effort) using these complicated functions get_bulk_element_edges, create_prolongations_over_bulk_element_edges


- now we want to resolve "pathologic" - now called special cases - in tracing alg. too; commenting 803,809 we have the info in which face the IP was created (neglecting whether it is inside, on the edge or at node); but we got rid of this precise info for prolongation... we can't have both at the moment...
- do I see this correctly? : if we build a prolongation table properly, then we do no need the precise info in prolongation routines

suggestion 1:
- DO NOT have the precise topology setting of IP (it will be just for the usage in tracing alg.)
(but what if we want to pass this info into the resulting object in future?)
- tracing alg. will work as it is in JB_intersection at the moment
- supposing special cases, the prolongation table will have to be two dimensional:
e.g. row on tetra: face; or 2 faces from edge; or 3 faces from node
- simplify the functions for 1d-3d prolongation with prolongation table; probably move the functionality of these partially into 1d-3d compute_intersection and build prolongation table (see the switch at inspect_elements_algorithm.cc:471)

suggestion 2:
- DO have the precise topology setting of IP
- then we need to pass some topology info into tracing alg. -> we can probably build the tracing table on the fly in 2d3d compute_intersection
- we can leave prolongation improvement for later...

suggestion3:
- we can do both and then it seems, that the topology in IP won't matter anymore as we build tracing and prolongation table directly..

src/intersection/compute_intersection.cc
jbrezmorf: Fixed.

jbrezmorf:
I'm thinking this way:
1. keep detailed topology info in IP as it was before
2. move tracking directly into 23d compute
	a. for every triangle side add (at most) two intersections in the order around the triangle, merging possible vertex IP on the way
	b. for IP on edge or vertex of tetra mark interacting edges
	c. compute edge-triangle intersections (skip marked edges)
        d. insert new IPs into existing list, there are only 3 places where we can add. This adding is tricky for IP on tetra vertex. Have to cover that case yet.

So we completely change (hopefully simplify) tracing, keep topology info, keep prolongations.

Nevertheless, I have to yet understand logic in 1d-3d prolongations and design a logic for 2d-3d case. This I want to describe in the paper even if it is not completely implemented.

Paulie14:
OK. I am not sure, what you mean by d) and the trick adding. I'll think about it..
I believe the logic of prolongation is correct. In future we need to complete all the cases in 2d-3d and move partially the logic of 1d-3d into compute_intersection using prolongation table..

I can help you write it down (at least the logic structure)..

jbrezmorf:
Back from lunch. I try to implement new tracking today. Please try to write down prolongation logic for the 1d-3d case. You can use the paper tex file. It would be enough if I can understand it.

PR #705, opened 2016-09-28
Df report exc address

dflanderka:


src/io/output_vtk.cc
```diff
@@ -98,7 +98,12 @@ int OutputVTK::write_data(void)
 
     if (! this->_base_file.is_open()) {
         this->fix_main_file_extension(".pvd");
-        this->_base_filename.open_stream( this->_base_file );
+        try {
+            this->_base_filename.open_stream( this->_base_file );
+        } catch (FilePath::ExcFileOpen &e ) {
+            e << FilePath::EI_Address_String(input_record_.address_string());
+            throw;
```
jbrezmorf: Introduce macro INPUT_CATCH:
```
try{ 
	some_call
} INPUT_CATCH(input_accessor);
```

jbrezmorf: add into input/accessors.hh

PR #724, opened 2016-11-06
Jb intersection

jbrezmorf:
Implement 23 intersection algorithm with topological polygon tracking for all special cases.


src/intersection/compute_intersection.cc
```diff
@@ -281,54 +285,86 @@ bool ComputeIntersection< Simplex< 1  >, Simplex< 2  > >::compute_pathologic(uns
...
-            }
         }
+    } else {
+        return IntersectionResult::degenerate;
     }
-    
-    return false;   // if IP not found before
 };
 
 unsigned int ComputeIntersection< Simplex< 1  >, Simplex< 2  > >::compute_final(vector< IntersectionPointAux< 1, 2 > >& IP12s)
```
Paulie14: function compute_final differs from compute
I think it should use the function compute, then possibly cut the abscissa; "compute_pathologic" then has to deal only with the degenerate case;

src/intersection/compute_intersection.cc
```diff
@@ -281,54 +285,86 @@ bool ComputeIntersection< Simplex< 1  >, Simplex< 2  > >::compute_pathologic(uns
...
+        //DebugOut().VarFmt(non_zero_idx).VarFmt(signed_plucker_product(non_zero_idx));
+
+        auto result = signed_plucker_product(non_zero_idx) > 0 ?
+                IntersectionResult::positive : IntersectionResult::negative;
+        IP.set_orientation(result);
+
+        IP12s.push_back(IP);
+        return result;
+
+    } else if(compute_zeros_plucker_products){
```
Paulie14: so the flag "compute_zeros_plucker_products" is not needed anymore, correct?
at the end, we have only compute and compute_final with same argument..

src/intersection/compute_intersection.cc
```diff
@@ -742,10 +778,11 @@ unsigned int ComputeIntersection<Simplex<1>, Simplex<3>>::compute(IntersectionAu
     unsigned int count = compute(intersection.i_points_);
     
     // set if pathologic!!
+    /*
     for(IntersectionPointAux<1,3> &p : intersection.i_points_){
         if(p.is_pathologic()) intersection.pathologic_ = true;
         break;
-    }
+    }*/
```
Paulie14: since tracing is not used, the flag "pathologic" is not needed at all in intersectionAux and IntersectionPointAux anymore

src/intersection/compute_intersection.cc
```diff
@@ -9,7 +9,7 @@
 #include "plucker.hh"
 #include "intersection_point_aux.hh"
 #include "intersection_aux.hh"
-#include "tracing_algorithm.hh"
+//#include "tracing_algorithm.hh"
```
Paulie14: we should then remove it from cmakelist..

unit_tests/intersection/compute_intersection_23_new_test.cpp
```diff
@@ -47,80 +47,110 @@ using namespace computeintersection;
...
-    c.push_back({
-                {0-1e-16, 0, 0.14},     //FIXME: rounding error, should be zero 
-                {0, 0.375, 0.125},
+                {0.8, 0.2, 0}}});
+    c.push_back({ "41_r", {
                 {0.2, 0.5, 0.2},
-                {0.4, 0, 0.3}});
-    c.push_back({
+                {0, 0.375, 0.125},
+                {0, 0, 0.14},     //FIXME: rounding error, should be zero
```
Paulie14: so this seems to be solved :)

unit_tests/intersection/intersect_problem_00.msh
```diff
@@ -0,0 +1,159 @@
+$MeshFormat
```
Paulie14: What is this mesh??

unit_tests/intersection/simple_meshes_23d_new/50_r_triangle_tetrahedron.msh.opt
```diff
@@ -0,0 +1,582 @@
+General.AxesFormatX = "%.3g";
```
Paulie14: is this committed intentionally?

src/intersection/compute_intersection.cc
jbrezmorf: Just do it.

unit_tests/intersection/intersect_problem_00.msh
jbrezmorf: Wrong intersection extracted from bigger mesh. Seems it is still not covered by new testsuite.

src/intersection/compute_intersection.cc
jbrezmorf: agree.

src/intersection/compute_intersection.cc
jbrezmorf: I hope so. Must be checked.

src/intersection/compute_intersection.cc
jbrezmorf: Done.

PR #725, opened 2016-11-07
Df units integral test

jbrezmorf:


jbrezmorf:
- move  time_function test into 02_generic_input
- try to figure out reason for interpolated flux on .right starting below zero at time zero, the first nonzero flux should be 0.08
- try to minimize output files (ref output)

PR #731, opened 2016-11-16
DF fix time marks

dflanderka:


jbrezmorf:
Add destruction of Balance for all equations.

src/tools/time_marks.hh
```diff
@@ -48,7 +48,31 @@ public:
...
+	struct Type {
+		Type() : bitmap_(0x1), equation_index_(1) {}
+		Type(unsigned long int bitmap, unsigned char equation_index) : bitmap_(bitmap), equation_index_(equation_index) {}
+
+		Type operator&(const Type& other) const {
+			//ASSERT(equation_index_ == other.equation_index_)((unsigned int)equation_index_)((unsigned int)other.equation_index_).error();
+			return Type( (bitmap_ & other.bitmap_), equation_index_);
+		}
+
+		Type operator|(const Type& other) const {
```
jbrezmorf: - if equation index of one operand is zero, set equation index of the result to the other one.
- if equation indices differs (and none is zero) ... error

jbrezmorf:
- make TimeMarks::marks vector of vectors, adding a new vector in new_mark_type.
- TimeMarks::begin, end - remove default time mark type

jbrezmorf:
TimeMarks::next, essert for same equation index of the TimeGovenor and the mask.

jbrezmorf:
Where is called TimeMarks::next ? Can we change TimeGovernor  to TimeStep?

jbrezmorf:
remove TimeMarkHash

src/tools/time_marks.hh
```diff
@@ -279,7 +319,7 @@ private:
     TimeMark::Type next_mark_type_;
 
     /// TimeMarks list sorted according to the their time.
-    std::vector<TimeMark> marks_;
+    std::vector< std::vector<TimeMark> > marks_;
```
jbrezmorf: Update comment

PR #747, opened 2016-12-02
Df time marks tests

dflanderka:


tests/03_generic_output/09_observe_output.yaml
```diff
@@ -57,6 +57,8 @@ problem: !Coupling_Sequential
       end_time: 1000000.0
     balance:
       cumulative: true
+      times:
+        - 20000.0
```
jbrezmorf: We should not search for inputs that leads to reference outputs.
In particular for this test, removing time 20000 from mass_balance output is fine. 
More over I wonder why there are any times in the water_balance ??

tests/03_generic_output/07_dg_dp_sorp_small_output_multi_fields.yaml
```diff
@@ -24,6 +24,7 @@ problem: !Coupling_Sequential
         bc_pressure: !FieldFormula
           value: x+y
     output:
+      add_input_times: true
```
jbrezmorf: This is OK.

tests/03_generic_output/08_fv_balance_times.yaml
```diff
@@ -36,7 +36,11 @@ problem: !Coupling_Sequential
         - pressure_p0
         - pressure_p1
         - velocity_p0
-    balance: {add_output_times: false}
+    balance: 
+      add_output_times: false
+      times:
```
jbrezmorf: Rather change output.

PR #757, opened 2016-12-13
J hy profiler memory

jbrezmorf:


janhybs:
Petsc calls are not reliable, closed

PR #783, opened 2017-01-23
Df vtk compressed output

jbrezmorf:


jbrezmorf:
- Fix description of compressed VTK variant, delate "not supported"
- move reading of the variant key into constructor. 
- Make compressed variant optional in the selection depending on the support of zlib detected by cmake.

PR #811, opened 2017-02-06
Df field interpolation

jbrezmorf:


src/fields/field_fe.hh
```diff
@@ -83,6 +110,43 @@ private:
     Mapping<2,3> *map2_;
     Mapping<3,3> *map3_;
 
+	FE_P_disc<0,1,3> fe1;
```
jbrezmorf: pointer to FiniteElement<Dim,3>

src/fields/field_fe.cc
```diff
@@ -290,11 +311,148 @@ void FieldFE<spacedim, Value>::value_list (const std::vector< Point >  &point_li
...
+}
+
+
+
+template <int spacedim, class Value>
+void FieldFE<spacedim, Value>::set_mesh(const Mesh *mesh, bool boundary_domain) {
+	// Mesh can't be set for field which is not initialized.
+	if (field_name_ == "") return;
+
+	dh_ = new DOFHandlerMultiDim( const_cast<Mesh &>(*mesh) );
```
jbrezmorf: New method make_dof_handler(), call it from set_mesh.
- set fe1,.. = new FE_disc_P<0,1,3> ... (will be passed through FieldCommon in future)
- create DohHandler, distribute dofs
- make maps
- allocate aux data and dof value vector (VecSequential)

src/fields/field_fe.cc
```diff
@@ -290,11 +311,148 @@ void FieldFE<spacedim, Value>::value_list (const std::vector< Point >  &point_li
...
+    // allocate data_
+	unsigned int data_size = dh_->n_global_dofs();
+	Vec v;
+    std::vector<double> data_vec;
+	data_vec.resize(data_size);
+	VecCreateSeqWithArray(PETSC_COMM_SELF, 1, data_size, &(data_vec[0]), &v);
+    data_vec_ = &v;
+    VecGetArray(*data_vec_, &data_);
+}
+
```
jbrezmorf: We should introduce a cache of DoF handlers.

src/fields/field_fe.cc
```diff
@@ -290,11 +311,148 @@ void FieldFE<spacedim, Value>::value_list (const std::vector< Point >  &point_li
...
+bool FieldFE<spacedim, Value>::set_time(const TimeStep &time) {
+	// Time can't be set for field which is not initialized.
+	if (field_name_ == "") return false;
+
+	ASSERT_PTR(dh_)(field_name_).error("Null target mesh pointer of elementwise field, did you call set_mesh()?\n");
+    if ( reader_file_ == FilePath() ) return false;
+
+    //walkaround for the steady time governor - there is no data to be read in time==infinity
+    //TODO: is it possible to check this before calling set_time?
+    //if (time == numeric_limits< double >::infinity()) return false;
```
jbrezmorf: delete

src/fields/field_fe.cc
```diff
@@ -290,11 +311,148 @@ void FieldFE<spacedim, Value>::value_list (const std::vector< Point >  &point_li
...
+    data_vec_ = &v;
+    VecGetArray(*data_vec_, &data_);
+}
+
+
+
+template <int spacedim, class Value>
+bool FieldFE<spacedim, Value>::set_time(const TimeStep &time) {
+	// Time can't be set for field which is not initialized.
+	if (field_name_ == "") return false;
```
jbrezmorf: Should be detected by FieldFlags.
```
if (equation_input && declare_input)
    if (field_name_ == "" ) error
    else reid from file ...
else skip
```
```

File: src/fields/field_fe.cc [hunk 6]
```diff
@@ -290,11 +311,148 @@ void FieldFE<spacedim, Value>::value_list (const std::vector< Point >  &point_li
...
+    FOR_ELEMENTS( dh_->mesh(), ele ) {
+    	searched_elements_.clear();
+    	((BIHTree *)bih_tree_)->find_point(ele->centre(), searched_elements_);
+    	found_point_.SetCoord( ele->centre()(0), ele->centre()(1), ele->centre()(2) );
+    	std::fill(sum_val.begin(), sum_val.end(), 0.0);
+    	std::fill(elem_count.begin(), elem_count.end(), 0);
+    	for (std::vector<unsigned int>::iterator it = searched_elements_.begin(); it!=searched_elements_.end(); it++) {
+    		ElementFullIter elm = source_mesh_->element( *it );
+			switch (elm->dim()) {
+				case 0: {
```
jbrezmorf: We should get rid of ngh. Possibly this kind of check is already part of Element ...

src/fields/field_fe.cc
dflanderka: We don't have access to FieldFlags in FieldFE (descendant of FieldAlgoBase). FieldFlags are part of FieldCommon.

jbrezmorf: Add field flags into FieldAlgoBaseInitData, see also issue #816

src/fields/field_fe.cc
jbrezmorf: Use:

	map =elm->get_map
	projection = elm->project_point(found_point_, map)
	if (projection.min() >= -BoundingBox::epsilon) {
		// projection in element
		projection_point = map.col(dim()) + map.cols(0, dim()-1) * projection.rows(0, dim()-1);
		if ( arma::norm(projection_point - found_point_, "inf") < 2*BoundingBox::epsilon ) {
			// point on the element
		}
	}

jbrezmorf: Possible optimization:
- compute explicitely projection matrix 3xele->dim and cache them for elements of the source mesh, 
- move all this projection a mapping code from Element into Mapping, which however must be reconsidered.  
  (Mapping does not store mappings, just use reinitialization per element)

tests/02_generic_input/06_bulk_interpolated.yaml
```diff
@@ -0,0 +1,33 @@
...
+    mesh_file: ../00_mesh/cube_1x1x1_frac21_center.msh
+  flow_equation: !Flow_Darcy_MH
+    output_specific:
+    nonlinear_solver:
+      linear_solver: !Petsc
+        a_tol: 1.0e-07
+    input_fields:
+      - region: .BOUNDARY
+        bc_type: dirichlet
+        bc_piezo_head: 0.1
```
jbrezmorf: bc_type: dirichlet
        bc_pressure: !FieldFormula
          value: x-z

src/fields/field.impl.hh
```diff
@@ -508,7 +508,7 @@ template<int spacedim, class Value>
 typename Field<spacedim,Value>::FieldBasePtr Field<spacedim,Value>::FactoryBase::create_field(Input::Record rec, const FieldCommon &field) {
 	Input::AbstractRecord field_record;
 	if (rec.opt_val(field.input_name(), field_record)) {
-		FieldAlgoBaseInitData init_data(field.input_name(), field.n_comp(), field.units(), field.limits());
+		FieldAlgoBaseInitData init_data(field.input_name(), field.n_comp(), field.units(), field.limits(), const_cast<FieldCommon &>(field).flags());
```
jbrezmorf: Why the const cast here? Rather introduce get_flags() which returns copy of flags and is const method.

jbrezmorf: Why the const cast here? Rather introduce get_flags() which returns copy of flags and is const method.

src/fields/field_fe.cc
```diff
@@ -34,38 +36,56 @@ namespace it = Input::Type;
...
+const Input::Type::Record & FieldFE<spacedim, Value>::get_input_type()
+{
+    return it::Record("FieldFE", FieldAlgorithmBase<spacedim,Value>::template_name()+" Field given by finite element approximation.")
+        .derive_from(FieldAlgorithmBase<spacedim, Value>::get_input_type())
+        .copy_keys(FieldAlgorithmBase<spacedim, Value>::get_field_algo_common_keys())
+        .declare_key("mesh_data_file", IT::FileName::input(), IT::Default::obligatory(),
+                "GMSH mesh with data. Can be different from actual computational mesh.")
+        .declare_key("field_name", IT::String(), IT::Default::obligatory(),
+                "The values of the Field are read from the ```$ElementData``` section with field name given by this key.")
+		//.declare_key("unit", FieldAlgorithmBase<spacedim, Value>::get_input_type_unit_si(), IT::Default::optional(),
```
jbrezmorf: can delete this

src/fields/field_fe.cc
```diff
@@ -290,11 +310,132 @@ void FieldFE<spacedim, Value>::value_list (const std::vector< Point >  &point_li
...
+	   // no call to mesh->setup_topology, we need only elements, no connectivity
+    }
+
+	field_name_ = rec.val<std::string>("field_name");
+}
+
+
+
+template <int spacedim, class Value>
+void FieldFE<spacedim, Value>::set_mesh(const Mesh *mesh, bool boundary_domain) {
```
jbrezmorf: Why having separate set_mesh at all? It can not work since set_mesh is not virtual. Just move this code into set_time? Any issue with that?

src/fields/field_fe.hh
```diff
@@ -69,20 +81,90 @@ public:
...
+	std::shared_ptr<Mesh> source_mesh_;
+
+	/// mesh reader file
+	FilePath reader_file_;
+
+	/// field name read from input
+	std::string field_name_;
+
+	/// stored index to last computed element
+	unsigned int computed_elm_idx_ = numeric_limits<unsigned int>::max();
```
jbrezmorf: What is it used for?

src/fields/field_fe.cc
```diff
@@ -290,11 +310,132 @@ void FieldFE<spacedim, Value>::value_list (const std::vector< Point >  &point_li
...
+
+		GMSH_DataHeader search_header;
+		search_header.actual = false;
+		search_header.field_name = field_name_;
+		search_header.n_components = this->value_.n_rows() * this->value_.n_cols();
+		search_header.n_entities = source_mesh_->element.size();
+		search_header.time = time.end();
+
+		bool boundary_domain_ = false;
+		std::vector<double> data_vec = *(ReaderInstances::instance()->get_reader(reader_file_)->template get_element_data<double>(search_header,
```
jbrezmorf: Split into getting reader and getting data.

src/fields/field_fe.cc
```diff
@@ -290,11 +310,132 @@ void FieldFE<spacedim, Value>::value_list (const std::vector< Point >  &point_li
...
+void FieldFE<spacedim, Value>::init_from_input(const Input::Record &rec, const struct FieldAlgoBaseInitData& init_data) {
+	this->init_unit_conversion_coefficient(rec, init_data);
+	flags_ = init_data.flags_;
+
+
+	// read mesh, create tree
+    {
+       reader_file_ = FilePath( rec.val<FilePath>("mesh_data_file") );
+       auto reader_data = ReaderInstance::get_instance(reader_file_);
+       source_mesh_ = reader_data.mesh_;
```
jbrezmorf: Move reading the mesh into set_time. And rather make source_mesh_ a local variable in order to free lot of memory. On the other hand we would like to keep it to reuse it when more FieldFE are interpolated from the same source mesh. But this is work for future.

src/fields/field_fe.cc
```diff
@@ -290,11 +310,132 @@ void FieldFE<spacedim, Value>::value_list (const std::vector< Point >  &point_li
...
+		search_header.n_entities = source_mesh_->element.size();
+		search_header.time = time.end();
+
+		bool boundary_domain_ = false;
+		std::vector<double> data_vec = *(ReaderInstances::instance()->get_reader(reader_file_)->template get_element_data<double>(search_header,
+				source_mesh_->elements_id_maps(boundary_domain_), this->component_idx_));
+		std::vector<double> sum_val(4);
+		std::vector<unsigned int> elem_count(4);
+
+		FOR_ELEMENTS( dh_->mesh(), ele ) {
```
jbrezmorf: Make separate (private) method for the interpolation. Further we shall probably make separate generic class from that for interpolation between any fields. An then we can have different generic interpolation algorithms.

src/fields/field_fe.hh
```diff
@@ -21,11 +21,18 @@
 #include "petscmat.h"
 #include "system/system.hh"
 #include "fields/field_algo_base.hh"
+#include "fields/vec_seq_double.hh"
+#include "mesh/mesh.h"
 #include "mesh/point.hh"
+#include "mesh/bih_tree.hh"
+#include "mesh/ngh/include/ngh_interface.hh"
```
jbrezmorf: Remove NGH dependency.
Try to minimize includes, replace by forward declarations where possible.

src/fields/field_fe.hh
```diff
@@ -69,20 +81,90 @@ public:
...
+
+    /// Destructor.
+	virtual ~FieldFE();
+
+private:
+	/// Create DofHandler object
+	void make_dof_handler(const Mesh *mesh);
+
+	/// DOF handler object
+    DOFHandlerMultiDim *dh_;
```
jbrezmorf: Review data members, make them local variables where possible.

src/fields/field_fe.hh
```diff
@@ -69,20 +81,90 @@ public:
...
+	std::string field_name_;
+
+	/// stored index to last computed element
+	unsigned int computed_elm_idx_ = numeric_limits<unsigned int>::max();
+
+	/// vector stored suspect elements in calculating the intersection
+	std::vector<unsigned int> searched_elements_;
+
+	/// 3D (tetrahedron) element, used for computing intersection
+	TTetrahedron tetrahedron_;
```
jbrezmorf: NGH object can be removed.?

src/mesh/reader_instances.hh
```diff
@@ -20,12 +20,38 @@
...
+#include "mesh/mesh.h"
 #include "system/file_path.hh"
 
+#include <memory>
+
 
 /**
  * Auxiliary class to map filepaths to instances of readers.
  */
+class ReaderInstance {
```
jbrezmorf: How is this class finally related to ReaderInstances?

src/fields/field_fe.cc
dflanderka: We have no way how to pass target mesh into set_time method.

src/mesh/reader_instances.hh
dflanderka: ReaderInstances will be replaced after full implementation of FieldFE.

src/fields/field_fe.cc
jbrezmorf: I see. So, set_mesh is virtual so it should work even if I prefere to just set the mesh pointer in that method. Nevertheless, the base implementation of set_mesh is called in update_history. First make
FieldFE::set_mesh as override to be sure that argument types match. Second, output typename of the field object just before set_mesh is called to get which instance was in fact created by the factory. ... so debug FieldFE creation process.

dflanderka: Field factory creates FieldFE, but for this case (piezo_head_p0) FieldFE is apparently encapsulated in FieldAddPotential. Is it possible?
I suggest to implement FieldAddPotential::set_mesh method (similarly as set_time).

jbrezmorf:
Of course, this is a hidden bug. It would be problem also for
FieldInterpolated and FieldElementwise isn't it?
Add the set_mesh there which has to call set_mesh of the inner field.

src/fields/field_fe.cc
```diff
@@ -97,14 +115,14 @@ typename Value::return_type const & FieldFE<spacedim, Value>::value(const Point
 		if (dh_->fe<1>()->is_scalar()) {
```
jbrezmorf: Make a protected method template (with element dimension as parameter) for evaluation in order to have common implementation for all element dimensions. Check that it works also for dim==0, which should resolve problems with usage on boundary.

src/fields/fe_value_handler.cc
```diff
@@ -0,0 +1,159 @@
...
+		map_ = new MappingP1<elemdim,3>();
+    } else {
+    	map_ = map;
+    }
+}
+
+
+template <int elemdim, int spacedim, class Value>
+typename Value::return_type const &FEValueHandler<elemdim, spacedim, Value>::value(const Point &p, const ElementAccessor<spacedim> &elm)
+{
```
jbrezmorf: value just call value_list

src/fields/fe_value_handler.cc
```diff
@@ -0,0 +1,159 @@
...
+		double value = 0;
+		for (unsigned int i=0; i<dh_->fe<elemdim>()->n_dofs(); i++)
+			value += (*data_vec_)[dof_indices[i]]*fe_values.shape_value(i, 0);
+		this->value_(0,0) = value;
+	}
+	else {
+		arma::vec3 value;
+		value.zeros();
+		for (unsigned int i=0; i<dh_->fe<elemdim>()->n_dofs(); i++)
+			value += (*data_vec_)[dof_indices[i]]*fe_values.shape_vector(i, 0);
```
jbrezmorf: this->value_.zeros()
this->r_value += ...

src/fields/fe_value_handler.cc
```diff
@@ -0,0 +1,159 @@
...
+		Quadrature<elemdim> quad(1);
+		quad.set_point(0, im*p_rel);
+
+		FEValues<elemdim,3> fe_values(*this->get_mapping(), quad, *dh_->fe<elemdim>(), update_values);
+		fe_values.reinit(cell);
+
+		Value envelope(value_list[k]);
+
+		if (dh_->fe<elemdim>()->is_scalar()) {
+			double value = 0;
```
jbrezmorf: envelope.zeros()

src/fields/fe_value_handler.cc
```diff
@@ -0,0 +1,159 @@
...
+			for (unsigned int i=0; i<dh_->fe<elemdim>()->n_dofs(); i++)
+				value += (*data_vec_)[dof_indices[i]]*fe_values.shape_value(i, 0);
```
jbrezmorf: value_list[k] += ...

src/fields/fe_value_handler.cc
```diff
@@ -0,0 +1,159 @@
...
+			envelope(0,0) = value;
+		}
+		else {
+			arma::vec3 value;
+			value.zeros();
+			for (unsigned int i=0; i<dh_->fe<elemdim>()->n_dofs(); i++)
+				value += (*data_vec_)[dof_indices[i]]*fe_values.shape_vector(i, 0);
```
jbrezmorf: Make helper resolution function fe_value(fe_value, i_dof, i_qpoint), specialize for different ranks
template<dim, spacedim>
fe_value<0>(fe_value, i_dof, i_qp)
{ shape_value}

call like:
fe_value<Value::rank>(fe_values, i,0)

PR #812, opened 2017-02-06
Switch to Python 3

x3mSpeedy:


src/system/python_loader.cc
```diff
@@ -77,45 +79,9 @@ PyObject * PythonLoader::load_module_from_file(const std::string& fname) {
...
-			ss << setw(pos) << "^" << endl;
-			THROW(ExcPythonError() << EI_PythonMessage( ss.str() ));
-    	} else {
-    		THROW(ExcPythonError() << EI_PythonMessage( PyString_AsString(pvalue) ));
-    	}
-    }
-
-    PyObject * result = PyImport_ExecCodeModule(tmp_name, compiled_string);
+    PyObject * compiled_string = Py_CompileString(source_string.c_str(), module_name.c_str(), Py_file_input);
+    PyObject * result = PyImport_ExecCodeModule(module_name.c_str(), compiled_string);
```
jbrezmorf: We should keep dedicated error checking code. It is not simply replaceable by check_error, since here we check success of compilation. Do not remove it unless you compare results of this one and check_error.

src/system/python_loader.cc
```diff
@@ -133,22 +99,22 @@ PyObject * PythonLoader::load_module_by_name(const std::string& module_name) {
...
     if (PyErr_Occurred()) {
         PyObject *ptype, *pvalue, *ptraceback;
-
         PyErr_Fetch(&ptype, &pvalue, &ptraceback);
-        string error_description = string(PyString_AsString(PyObject_Str(pvalue)));
-
+        string error_description = string(PyUnicode_AsUTF8(PyObject_Str(pvalue)));
+        
+        // clear error indicator
+         PyErr_Clear();
```
jbrezmorf: Is it OK to clear it here? Didn't we loose the traceback?

src/system/sys_profiler.cc
```diff
@@ -830,15 +830,15 @@ void Profiler::transform_profiler_data (const string &output_file_suffix, const
     PyObject * arguments = PyTuple_New (3);
 
     // set json path location as first argument
-    PyObject * tmp = PyString_FromString (json_filepath.c_str());
+    PyObject * tmp = PyUnicode_FromString (json_filepath.c_str());
```
jbrezmorf: Please, document (python_loader.hh) various string formats used in interaction with python and list corresponding conversion functions provided in PythonLoader or by Python API itself. 

In particular,  here: What is difference between PyUnicode_FromString and PythonLoader::to_py_string?

CMakeLists.txt
```diff
@@ -554,8 +554,9 @@ if(USE_PYTHON)
 endif()
 
 # find python 
-find_package(PythonInterp 2.7 REQUIRED)
-find_package(PythonLibs 2.7   REQUIRED)
+set(Python_ADDITIONAL_VERSIONS 3.4) # for cmake version 2.8
```
x3mSpeedy: Yes, problem before was in `CMakeCache`

src/system/python_loader.cc
x3mSpeedy: PyErr_Clear()
> Clear the error indicator. If the error indicator is not set, there is no effect.

We fetch error indicator beforehand using `PyErr_Fetch()`

src/system/sys_profiler.cc
x3mSpeedy: Add comments in header file

src/system/python_loader.cc
x3mSpeedy: This error check routine does not offer anything new compared to check_error.

x3mSpeedy:
Python side should now be also done, ready for merge if build [No. 352](http://ciflow.nti.tul.cz:8080/job/Flow123d-ci2runner-debug-multijob/352/) passes @jbrezmorf

PR #815, opened 2017-02-09
Pe 2d2d

Paulie14:
2D-2D intersections

- fix fundamental algorithm (compute_final, compute_degenerate)
- component numbering for 2D
- 2D-2D intersections in 3D bulk mesh
- new unit tests


src/intersection/compute_intersection.cc
```diff
@@ -301,24 +303,25 @@ IntersectionResult ComputeIntersection<Simplex<1>, Simplex<2>>::compute(std::vec
...
         w = w / w_sum;
-
-        for (unsigned int i=0; i < 3; i++) {
-            //DebugOut().VarFmt(i).VarFmt(w[i]);
-            if (w[i] > rounding_epsilon) n_positive++;
-            else if ( w[i] > -rounding_epsilon) zero_idx_sum+=i;
-            else n_negative++;
-        }
+        
+    for (unsigned int i=0; i < 3; i++) {
```
Paulie14: This FOR cycle used to be under the IF condition before. I believe it was wrong. 
I had to solve problem, where products where -0.5, -0.5, 1.0 which summed to zero and let to degenerate case (and should not).
In compute_final and compute_degenerate I consider that the degenerate case means all three zero products.
Do you agree?
I add 12 unit test mesh for this.

PR #840, opened 2017-03-28
Df death test option

dflanderka:


jbrezmorf:
Try to minimize input files, leaving only necessary input. Mark lines with errors.

PR #842, opened 2017-03-29
Df barycentric coords

dflanderka:


src/mesh/ref_element.cc
```diff
@@ -392,9 +385,9 @@ auto RefElement<dim>::barycentric_on_face(const BaryPoint &barycentric, unsigned
     ASSERT_EQ_DBG(barycentric.n_rows, dim+1);
     FaceBaryPoint face_barycentric;
     for(unsigned int i=0; i < dim; i++) {
-        unsigned int i_sub_node = (i+1)%dim;
+//        unsigned int i_sub_node = (i+1)%dim;
 //         unsigned int i_bary = (dim + side_nodes_[i_face][i_sub_node])%(dim+1);
-        unsigned int i_bary = (dim + interact_<0,dim-1>(i_face)[i_sub_node])%(dim+1);
+        unsigned int i_bary = (dim + interact_<0,dim-1>(i_face)[i])%(dim+1);
```
jbrezmorf: Seems that we should use just 'i_bary = interact...'

PR #851, opened 2017-04-21
Js fe system 300

jstebel:
Implements #5.


src/fem/dofhandler.hh
```diff
@@ -75,6 +75,8 @@ public:
      * @brief Returns the offset of the local part of dofs.
      */
     const unsigned int loffset() const { return loffset_; }
+    
+    const unsigned int max_elem_dofs() const { return max_elem_dofs_; }
```
jbrezmorf: Documentation, purpose of the new method?

src/fem/fe_p.cc
```diff
@@ -19,6 +19,20 @@
 #include "fe_p.hh"
 #include "mesh/ref_element.hh"
```
jbrezmorf: We can definitely remove templating for all FE classes, since evaluation is done once on reference element in all quadrature points so there is no need for any optimization. The only exception could be XFEM, but that case should be treated in different way.

src/fem/fe_p.hh
```diff
@@ -354,7 +368,9 @@ FE_P<degree,dim,spacedim>::FE_P()
         unit_support_points.push_back(dof_distribution.unit_support_points[i]);
 
     order = degree;
-
+    
+    this->setup_components();
```
jbrezmorf: ??

src/fem/fe_p.hh
```diff
@@ -464,73 +492,5 @@ FE_P_disc<degree,dim,spacedim>::~FE_P_disc()
 
 
 
-#ifndef DOXYGEN_SHOULD_SKIP_THIS
```
jbrezmorf: Why we can remove this?

src/fem/fe_rt.hh
```diff
@@ -39,7 +39,7 @@ class FE_RT0 : public FiniteElement<dim,spacedim>
     using FiniteElement<dim,spacedim>::number_of_sextuples;
     using FiniteElement<dim,spacedim>::generalized_support_points;
     using FiniteElement<dim,spacedim>::order;
-    using FiniteElement<dim,spacedim>::is_scalar_fe;
+    using FiniteElement<dim,spacedim>::is_primitive_;
```
jbrezmorf: Why this using stuff? Is that because of templates?

src/fem/fe_rt.hh
```diff
@@ -63,18 +63,28 @@ public:
...
-     * @brief Returns the @p ith basis function evaluated at the point @p p.
-     * @param i Number of the basis function.
-     * @param p Point of evaluation.
+     * @brief Calculates the value of the @p comp-th component of
+     * the @p i-th raw basis function at the
+     * point @p p on the reference element (for vector-valued FE).
+     *
+     * @param i    Number of the basis function.
+     * @param p    Point of evaluation.
+     * @param comp Number of vector component.
```
jbrezmorf: Convenction for numbering components should be described, in FiniteElement class description.

src/fem/fe_rt.hh
```diff
@@ -93,7 +103,7 @@ public:
      * @param flags Flags that indicate what quantities should be calculated.
      */
     FEInternalData *initialize(const Quadrature<dim> &q, UpdateFlags flags);
```
jbrezmorf: Is it necessary to flag unneeded data? Is this computed just few times?

src/fem/fe_system.cc
```diff
@@ -0,0 +1,416 @@
...
+//   return finite_elements;
+// }
+
+
+
+template<unsigned int dim, unsigned int spacedim>
+FESystem<dim,spacedim>::FESystem(FiniteElement<dim,spacedim> *fe, unsigned int n)
+{
+  FiniteElement<dim,spacedim>::init(0, false);
+  std::shared_ptr<FiniteElement<dim,spacedim> > fe_ptr(fe);
```
jbrezmorf: 1. We should make a copy of *fe, which should be a reference.
2. We should support system of different finite elements, e.g constructor:
FeSystem(vector<pair<FiniteElement, uint>> fe_powers)

or
FeSystem(vector<FiniteElement> fes, vector<uint> multiplicity = {})

src/fem/fe_system.cc
```diff
@@ -0,0 +1,416 @@
...
+  initialize();
+}
+
+
+
+template<unsigned int dim, unsigned int spacedim>
+void FESystem<dim,spacedim>::initialize()
+{
+  // indices of nodal dofs (for each node)
+  std::vector<unsigned int> node_basis[dim+1];
```
jbrezmorf: What about edge and face dofs? Can we have dimension as index? 
vector<vector<uint>> basis; // basis[dim][i_dof]

src/fem/fe_system.cc
```diff
@@ -0,0 +1,416 @@
...
+  // for each base FE add components, support points, and other 
+  // information to the system
+  for (auto fe : fe_)
+  {
+    number_of_dofs += fe->n_dofs();
+    n_components_ += fe->n_components();
+
+    for (int i=0; i<=dim; ++i)
+    {
+      number_of_single_dofs[i] += fe->n_object_dofs(i, DOF_SINGLE);
```
jbrezmorf: Remind me why we have these pairs, triples, ...

src/fem/fe_system.cc
```diff
@@ -0,0 +1,416 @@
...
+      number_of_pairs[i] += fe->n_object_dofs(i, DOF_PAIR);
+      number_of_triples[i] += fe->n_object_dofs(i, DOF_TRIPLE);
+      number_of_sextuples[i] += fe->n_object_dofs(i, DOF_SEXTUPLE);
+    }
+
+    for (int i=0; i<fe->get_generalized_support_points().size(); i++)
+        generalized_support_points.push_back(fe->get_generalized_support_points()[i]);
+
+    if (fe->polynomial_order() > order) order = fe->polynomial_order();
```
jbrezmorf: Does it have a sense?

src/fem/fe_system.cc
```diff
@@ -0,0 +1,416 @@
...
+
+    for (int i=0; i<=dim; ++i)
+    {
+      number_of_single_dofs[i] += fe->n_object_dofs(i, DOF_SINGLE);
+      number_of_pairs[i] += fe->n_object_dofs(i, DOF_PAIR);
+      number_of_triples[i] += fe->n_object_dofs(i, DOF_TRIPLE);
+      number_of_sextuples[i] += fe->n_object_dofs(i, DOF_SEXTUPLE);
+    }
+
+    for (int i=0; i<fe->get_generalized_support_points().size(); i++)
```
jbrezmorf: Improve doc for get_generalized_support_points, it is not clear what is should be for general finite element, e.g. RT.

jbrezmorf: Can we somehow define functionals?

src/fem/fe_system.cc
```diff
@@ -0,0 +1,416 @@
...
+    double dof_index = 0;
+    comp_offset = 0;
+    // add footprint of nonzero components for each dof in FESystem
+    for (auto fe : fe_)
+    {
+      for (int i=0; i<fe->n_dofs(); ++i)
+      {
+        std::vector<bool> nonzeros(n_components_, false);
+        for (unsigned int c=0; c<fe->n_components(); c++)\
+          nonzeros[comp_offset+c] = fe->get_nonzero_components(fe_dof_indices_[dof_index++].basis_index)[c];
```
jbrezmorf: Usage of get_nonzero_components?

src/fem/fe_system.cc
```diff
@@ -0,0 +1,416 @@
...
+    }
+//   }
+
+  compute_node_matrix();
+}
+
+
+
+template<unsigned int dim, unsigned int spacedim>
+double FESystem<dim,spacedim>::basis_value(const unsigned int i, const arma::vec::fixed<dim> &p) const
```
jbrezmorf: This is confusing: 
- base_value returns scalar, even if called for FeSystem, and it probably fails for vector valued Fe
- do we need this method at all?

src/fem/fe_system.cc
```diff
@@ -0,0 +1,416 @@
...
+{
+  OLD_ASSERT(i <= number_of_dofs, "Index of basis function is out of range.");
+  
+  unsigned int bi = dof_basis_[i];
+  // component index in the base FE
+  int l_comp = comp-fe_dof_indices_[bi].component_offset;
+  OLD_ASSERT(l_comp >= 0 && l_comp < fe_[fe_dof_indices_[bi].fe_index]->n_components(),
+    "Index of component is out of range.");
+
+  if (fe_[fe_dof_indices_[bi].fe_index]->n_components() == 1)
```
jbrezmorf: So this can be simplified if we have only this method and not the previous one.

src/fem/fe_system.cc
```diff
@@ -0,0 +1,416 @@
...
+    for (auto fe : fe_)
+      f |= fe->update_each(flags);
+
+    return f;
+}
+
+
+template<unsigned int dim, unsigned int spacedim>
+void FESystem<dim,spacedim>::compute_node_matrix()
+{
```
jbrezmorf: Is this used out of individual finite elements?

src/fem/fe_system.cc
```diff
@@ -0,0 +1,416 @@
...
+    }
+  }
+  
+  for (auto d : fe_data) delete d;
+
+  return data;
+}
+
+template<unsigned int dim, unsigned int spacedim> inline
+void FESystem<dim,spacedim>::fill_fe_values(
```
jbrezmorf: What have to be computed on every cell? The implementation in FiniteElement just copy values on reference element.
For RT, we should just apply mapping to values on reference element.

src/fem/fe_p.hh
jbrezmorf: Removed unnecessary declaration of specializations.

src/fem/fe_rt.hh
jbrezmorf: ... ok this is because of templates

src/fem/fe_values_views.hh
```diff
@@ -0,0 +1,156 @@
...
+    unsigned int component_;
+  };
+  
+  class Vector {
+  public:
+    Vector(unsigned int comp) : first_vector_component_(comp) {};
+    
+    unsigned int first_vector_component_;
+  };
+
```
jbrezmorf: add empty Tensor

jbrezmorf: Make methods: FeValues<>::get_scalar_view, get_vector_view, get_tensor_view which takes component index as parameter and return particular extractor. This allows performing checking of compatibility of component with the system structure. Add specification of quantinty in constructor of FeSystem, optional parameter saying that I;m creating Vector or Tensor quantity.

src/fem/fe_system.cc
jbrezmorf: Two constructors:
FESystem(shared_ptr<FiniteElement...> fe, quantity=Vector) // Vector , Tensor ... determines number of 
FESystem(shared_ptr<FiniteElement...> fe, n) // Mixed system 
FESystem(vector<shared_ptr<FiniteElement...>> fe_vec)

src/fem/fe_system.cc
jbrezmorf: Where we need this method? Try to remove or make it protected or private.

jbrezmorf: Make also is_primitive() private.

src/fem/fe_system.cc
jbrezmorf: Remove variable order. Only keep it as parameter of some constructors to construct particular order of element.

src/fem/fe_p.cc
jstebel: Postponed to #887.

src/fem/fe_p.hh
jstebel: Added documentation of FiniteElement::setup_components().

src/fem/fe_rt.hh
jstebel: Numbering of components is clear, what is not intuitive, is the ordering of basis functions (differs from ordering of dofs in dof handler).

src/fem/fe_system.cc
jstebel: OK

src/fem/fe_system.cc
jstebel: Will be removed in #887

src/fem/fe_system.cc
jstebel: OK

src/fem/fe_system.cc
jstebel: Postponed to #887

src/fem/fe_system.cc
jstebel: Postponed to #887.

src/fem/fe_system.cc
jstebel: Postponed to #887

src/fem/fe_system.cc
jstebel: Postponed to #887

src/fem/fe_system.cc
jstebel: Postponed to #887

src/fem/fe_values_views.hh
jstebel: If I understand well, the new methods should have names get_{scalar,vector,tensor}_extractor (not view). Then I have 2 objections:
- the FEValues object must exist before the extractor
- the extractor does not change for different FEValues when using the same FiniteElement and hence it is more natural to place these new methods into FiniteElement

jstebel:
Review comments processed, ready to merge.

src/fem/fe_system.cc
jbrezmorf: This seems simple make it now.

src/fem/fe_system.cc
jbrezmorf: This seems simple make it now.

src/fem/fe_system.cc
jbrezmorf: This seems simple make it now.

src/fem/fe_system.cc
jbrezmorf: Make these methods private or protected.

jstebel: Done.

src/fem/fe_system.cc
jstebel: Done

src/fem/fe_system.cc
jstebel: Done

src/fem/fe_system.cc
jstebel: Done

PR #862, opened 2017-07-25
Js dof handler

jstebel:


src/mesh/duplicate_nodes.cc
```diff
@@ -0,0 +1,256 @@
...
+    }
+    
+    // divide node_elements into components connected by node_edges
+    std::vector<std::set<unsigned int> > components;
+    while (node_elements.size() > 0)
+    {
+      // create component containing the first element in node_elements
+      std::queue<unsigned int> q;
+      q.push(node_elements[0]);
+      node_elements.erase(node_elements.begin());
```
jbrezmorf: Avoid erasing it is expensive, O(n). Rather use real graph representation and component marking.

src/mesh/duplicate_nodes.cc
```diff
@@ -0,0 +1,256 @@
...
+ * 1) Create the lists of elements (node_elements) and edges (node_edges) sharing the node.
+ * 2) Divide node_elements into groups (components) which share one of node_edges.
+ * 3) If more than one component was created, create for each extra component a new node
+ *    and update its index in the elements from this component.
+ *    Also set node_dim_ (dimension of elements using the node).
+ */
+void DuplicateNodes::duplicate_nodes()
+{
+  FOR_NODES(mesh_, n){
+    // create list of elements
```
jbrezmorf: For every node we should create a graph, list of:
```
Vertex is a struct containing {
	uint el_idx, component;
	uint n_neigbours;
	uint neighbours[30];  }
```
We need unordered map: el_idx to vtx_idx
For an mesh edge add the graph edge for every pair of elements connected by the mesh edge.
`vertices[el_to_vtx[el1]].neighbours.push_back(el_to_vtx[el2]`

src/mesh/duplicate_nodes.cc
```diff
@@ -0,0 +1,256 @@
...
+    std::vector<unsigned int> node_elements = mesh_->node_elements[n.index()];
+    
+    // create list of edges sharing the node
+    std::set<unsigned int> node_edges;
+    for (unsigned int el_idx : node_elements)
+    {
+      FOR_ELEMENT_SIDES(&mesh_->element[el_idx], sid)
+        node_edges.insert(mesh_->element[el_idx].side(sid)->edge_idx());
```
jbrezmorf: This includes also edges not associated with the node, but possibly it does not matter.

src/mesh/duplicate_nodes.h
```diff
@@ -0,0 +1,151 @@
...
+#define DUPLICATE_NODES_H
+
+
+
+
+class Mesh;
+
+template<int dim>
+class MeshObject {
+public:
```
jbrezmorf: Document this class, probably it represents any edge, face or element (n-face) of the mesh in terms of 
(n-1)-face.

src/mesh/duplicate_nodes.h
```diff
@@ -0,0 +1,151 @@
...
+  
+  /// Duplicate nodes that are lying on interfaces with fractures.
+  void duplicate_nodes();
+  
+  
+  /// The mesh object.
+  Mesh *mesh_;
+  
+  /// Internal indices of nodes (including duplicated nodes).
+  std::vector<unsigned int> nodes_;
```
jbrezmorf: So nodes_[i] is mesh node index for every duplicated node index?

jbrezmorf: Try to distinguish "mesh node" and "dof node" or "duplicated node".

src/mesh/duplicate_nodes.h
```diff
@@ -0,0 +1,151 @@
...
+  
+  /// Vector of space dimensions of elements using the particular node.
+  std::vector<unsigned int> node_dim_;
+  
+  /// Internal indices of points (0d elements).
```
jbrezmorf: Mesh objects are one to one with mesh nodes, edges and elements?

src/mesh/duplicate_nodes.h
```diff
@@ -0,0 +1,151 @@
...
+  std::vector<MeshObject<1> > lines_;
+  
+  /// Internal indices of triangles (2d elements).
+  std::vector<MeshObject<2> > triangles_;
+  
+  /// Internal indices of tetrahedra (3d elements).
+  std::vector<MeshObject<3> > tetras_;
+  
+  /// Vector of object indices for each mesh element.
+  std::vector<unsigned int> obj_4_el_;
```
jbrezmorf: Can we use just the same order as in 
Mesh?

src/mesh/duplicate_nodes.cc
```diff
@@ -0,0 +1,256 @@
...
+        }
+        q.pop();
+      }
+      components.push_back(component);
+    }
+    
+    // for each component except the first one
+    // create a duplicate of the node
+    if (components.size() > 0) {
+      node_dim_[obj_4_node_[n.index()]] = mesh_->element[*components[0].begin()].dim();
```
jbrezmorf: So from here it seems that node_dim_ gives dim for every MeshObj<0> which should be for every mesh node and this doesn't make sense. more over since later on the push_back is used, this line can simply overwrite a previous value.

src/mesh/duplicate_nodes.cc
```diff
@@ -0,0 +1,256 @@
...
+      components.erase(components.begin());
+    }
+    for (auto component : components) {
+      unsigned int new_nid = nodes_.size();
+      nodes_.push_back(new_nid);
```
jbrezmorf: So what is the point to store sequence 0,1,...

src/mesh/duplicate_nodes.cc
```diff
@@ -0,0 +1,256 @@
...
+      node_dim_.push_back(mesh_->element[*component.begin()].dim());
+      for (auto el_idx : component) {
+        switch (mesh_->element[el_idx].dim()) {
+          case 1:
+            {
+              MeshObject<1> *elem = &lines_[obj_4_el_[el_idx]];
+              for (unsigned int i=0; i<2; i++)
+                if (elem->nodes[i] == n.index())
+                  elem->nodes[i] = new_nid;
+              for (unsigned int fi=0; fi<2; fi++)
```
jbrezmorf: Is it enough to change node references just for first two levels of recursion? Isn't then better to store nodes only for MeshObject<0> and apply recursion when getting the nodes of other MeshObjects

jbrezmorf:
I didn't review parallel Dof distribution yet. We first have to narrow the sequential case.

src/fem/discrete_space.cc
```diff
@@ -0,0 +1,56 @@
...
+  }
+}
+
+
+unsigned int EqualOrderDiscreteSpace::n_node_dofs(unsigned int nid) const
+{
+  unsigned int dim = mesh_->tree->node_dim()[nid];
+  switch (dim)
+  {
+    case 1: return fe1_->n_object_dofs(0, DofMultiplicity::DOF_SINGLE)/(dim+1);
```
jstebel: dim+1=number of nodes per element (FiniteElement::n_object_dofs() returns number of dofs on all nodes). We may redefine this to return number of dofs per one entity (node, edge, face).

src/fem/discrete_space.hh
```diff
@@ -0,0 +1,110 @@
...
+ * - p-refined spaces using variable FE order,
+ * - XFEM with arbitrary shape functions on every element.
+ * 
+ */
+class DiscreteSpace {
+public:
+    
+  /// Number of dofs associated to node. @p nid is the node index in the mesh tree.
+  virtual unsigned int n_node_dofs(unsigned int nid) const = 0;
+
```
jstebel: Methods for dofs on edges and faces are not implemented yet (depends also on DuplicateNodes).

The number of dofs for one n-face (point/line/triangle) must be unique (elements sharing this n-face may not have different number of dofs on it).

src/mesh/duplicate_nodes.cc
jstebel: Right, this makes no sense (it is a relict from version where I had pointers instead of indices). So in fact we only need number of (duplicated) nodes.

src/mesh/duplicate_nodes.h
jstebel: Indeed, nodes_ are indices of duplicated nodes. As commented above, this makes no sense and we only need the number of duplicated nodes.

src/mesh/duplicate_nodes.h
jstebel: Yes. We need to have global indexing and information about subfaces for n-faces of all dimensions (so far we only have elements, edges (dim-1) and nodes).

src/mesh/duplicate_nodes.h
jstebel: No. Elements are all together, objects are grouped by dimension. So obj_4_el_[i] returns the index of i-th element in the vector lines_, triangles_ or tetras_, depending on the element dimension.

src/mesh/duplicate_nodes.h
jstebel: OK.

src/mesh/duplicate_nodes.cc
jstebel: Indeed, node_edges is a local set of edges in the vicinity of the node, used to determine which elements belong to the same component. Checking if edge is associated with the node would not pay off, I guess.

src/mesh/duplicate_nodes.cc
jstebel: In fact, node_dim_ should give the dim for every duplicated node. The code is not very clear, but I think that obj_4_node_[i] = i (so it is useless). The value of n.index() is always less than the index of any subsequently added (duplicated) node, so no overwriting happens.

src/mesh/duplicate_nodes.cc
jstebel: Currently it is enough since currently, MeshObjects are only elements, their 1-codimensional faces and nodes. In the future, when we will have complete graph of n-faces, it would not be sufficient.

src/fem/discrete_space.hh
jstebel: Added empty method n_face_dofs().

jstebel:
Concerning efficiency of duplicating the nodes: The time of (1) reading, (2) setting up topology and (3) duplicating nodes are almost the same. Tested on several meshes up to 800k elements. See repo flod123d/bench_data, 03_mesh_read.

src/mesh/duplicate_nodes.h
```diff
@@ -0,0 +1,161 @@
...
+  
+  DuplicateNodes(Mesh *mesh);
+
+  // Getters (see private section for explanation).  
+  Mesh *mesh() const { return mesh_; }
+  
+  unsigned int n_nodes() const { return n_duplicated_nodes_; }
+  const std::vector<unsigned int> &node_dim() const { return node_dim_; }
+  
+  const std::vector<MeshObject<0> > &points() const { return points_; }
```
jbrezmorf: We definitely should provide way to pass dimension as parameter either as template parameter or runtime parameter having different method for different object dimension leads to lot of code duplication when using the DuplicateNodes class. Definitely there is lack of documentation.

PR #877, opened 2017-09-27
Df output tests

dflanderka:
- add new integral tests with different types of field interpolation


src/fields/equation_output.cc
```diff
@@ -134,7 +134,10 @@ void EquationOutput::read_from_input(Input::Record in_rec, const TimeGovernor &
     auto fields_array = in_rec.val<Input::Array>("fields");
     for(auto it = fields_array.begin<Input::Record>(); it != fields_array.end(); ++it) {
         string field_name = it -> val< Input::FullEnum >("field");
+        FieldCommon *found_field = field(field_name);
```
jbrezmorf: - Make separate loop over all EquationOutput fields, containing:
- check for field interpolation UNDEFINED, replace by default interpolation of the field
- call stream_->add_field_interpolation

PR #878, opened 2017-10-02
Partial merge of output code from PE_xfem.

Paulie14:
Output mesh edit, mainly regarding adaptive refinement.


jbrezmorf:
Make an integration test in 04_generic_output, to demonstrate usage of the introduced parameters.

src/fields/field.impl.hh
```diff
@@ -642,7 +641,10 @@ void Field<spacedim,Value>::compute_field_data(OutputTime::DiscreteSpace space_t
     }
     break;
     case OutputTime::ELEM_DATA: {
-    	std::shared_ptr<OutputMesh> output_mesh = std::dynamic_pointer_cast<OutputMesh>( stream->get_output_mesh_ptr() );
+        std::shared_ptr<OutputMeshBase> output_mesh = stream->get_output_mesh_ptr(true);
```
jbrezmorf: Does it mean that output on refined mesh is supported only for ELEM_DATA? What about NODE_DATA, CORNER_DATA, NATIVE_DATA.

src/io/output_time.cc
```diff
@@ -114,7 +118,7 @@ Input::Iterator<Input::Record> OutputTime::get_output_mesh_record() {
 
 std::shared_ptr<OutputMeshBase> OutputTime::create_output_mesh_ptr(bool init_input) {
 	bool discont = (used_interpolations_.find(DiscreteSpace::CORNER_DATA) != used_interpolations_.end());
-	if (discont) {
+	if (discont || this->get_output_mesh_record()) {
```
Paulie14: I did this, so that the discontinuous output mesh is created when OutputMesh record is specified (i has only keys defining refinement).

Issue #875, opened 2017-09-11
Memory consumption


Paulie14:
Motivated by the problem at: 
`bacula.nti.tul.cz/intec/lide/jirina.kralovcova/00_share/2017_Vypocet_flow3_06`
which can be solved with 2.1.2 and not in 3.0.0 due to larger memory consumption.

Smaller test (33k elements) was run:
`bacula.nti.tul.cz/intec/lide/pavel.exner/00_share/flow_problems/2017_memory_3_0_0`
which showed the following:
- over 10 % increase in Petsc allocation `MatSeqAIJSetPreallocation_SeqAIJ` in `LinSys_PETSC::preallocate_matrix() (linsys_PETSC.cc:237)`
- computation in schur complement takes additional memory in 3.0.0_dev `MatAXPY_SeqAIJ` in `SchurComplement::form_schur() (schur.cc:218)`
- Balance takes quite a lot of memory and allocation is called 3 times `Balance::lazy_initialize() (balance.cc:282)`
- we do not know why allocation `MatILUFactorSymbolic_SeqAIJ` in `LinSys_PETSC::solve() (linsys_PETSC.cc:381)` is called 4 times
- in 2.1.2 max peak is 284.3 MB, in 3.0.0_dev max peak is 320.5 MB

See the results at bacula link, use massif visualizer:
```
add-apt-repository ppa:kubuntu-ppa/backports
apt-get update && apt-get install massif-visualizer
```
Increase of the number of detailed snapshots in massif might be necessary.

Paulie14:
I also tried MatInfo on schur0 matrix:
```
        MatInfo info;
        MatGetInfo(*schur0->get_matrix(),MAT_GLOBAL_SUM,&info);
        DBGVAR(info.nz_allocated);    // 2.67694e+06
        DBGVAR(info.nz_used);         // 2.61379e+06
        DBGVAR(info.nz_unneeded);     // 63150
```
The unneeded nz are 2,36%, so the larger allocation itself might not be the main issue.

jstebel:
Try to reduce allocated memory by setting ```MatSetOption(m, MAT_IGNORE_ZERO_ENTRIES, PETSC_TRUE)``` where applicable.

Paulie14:
`MatSetOption(m, MAT_IGNORE_ZERO_ENTRIES, PETSC_TRUE)` works great, from 320 MB peak I got to 270 MB peak.
[...]

PR #884, opened 2017-10-26
Pe 300 leaks

jbrezmorf:


src/flow/darcy_flow_mh.cc
```diff
@@ -712,17 +719,14 @@ void DarcyMH::allocate_mh_matrix()
         ls->mat_set_values(loc_size, local_dofs, loc_size, local_dofs, zeros);
         
 
-        std::vector<int> tmp_rows;
-        tmp_rows.reserve(200);
-
         // compatible neighborings rows
         unsigned int n_neighs = ele_ac.full_iter()->n_neighs_vb;
-        for (unsigned int i = 0; i < n_neighs; i++) {
+        for (i = 0; i < n_neighs; i++) {
```
jbrezmorf: Why removed 'unsigned int'?

src/flow/darcy_flow_mh.cc
```diff
@@ -712,17 +719,14 @@ void DarcyMH::allocate_mh_matrix()
...
             // every compatible connection adds a 2x2 matrix involving
             // current element pressure  and a connected edge pressure
             Neighbour *ngh = ele_ac.full_iter()->neigh_vb[i];
             int neigh_edge_row = mh_dh.row_4_edge[ ngh->edge_idx() ];
-            tmp_rows.push_back(neigh_edge_row);
+            tmp_rows[i] = neigh_edge_row;
```
jbrezmorf: This is wrong, reserve is not resize. tmp_rows have still size 0.

src/flow/darcy_flow_assembly.hh
```diff
@@ -150,7 +150,9 @@ public:
         assemble_element(ele_ac);
         assemble_source_term(ele_ac);
         
-        
+        //due to time dependent flow, fill the whole diagonal (otherwise it is allocated)
```
jbrezmorf: So it is necessary to keep matrix structure unchanged? But what can change for unsteady case? Only possibility is switching from steady to unsteady. Comment seems misleading since this have noting to do with allocation, but with allowing reuse of matrix structure.

src/la/local_system.hh
```diff
@@ -28,6 +28,8 @@ class LocalSystem
 public:
     typedef arma::uvec DofVec;
     
+    static constexpr double artificial_zero = std::numeric_limits<double>::min();
```
jbrezmorf: What about 'almost_zero', thats more descriptive. Also add detailed explanation of the whole problem here together with suggestions where to use it.

unit_tests/la/schur_compl_test.cpp
```diff
@@ -135,7 +135,7 @@ TEST(schur, complement) {
 		const PetscScalar *vals;
 		for (unsigned int i=0; i<block_size; i++) {
 			MatGetRow(schurComplement->get_a_inv(), i + rank*block_size, &ncols, &cols, &vals);
-			EXPECT_FLOAT_EQ( (1.0 / (double)(rank + 2)), vals[i] );
+			EXPECT_FLOAT_EQ( (1.0 / (double)(rank + 2)), vals[0] );
```
jbrezmorf: Why this change?

Paulie14: Thanks to setting `MatSetOption(matrix_, MAT_IGNORE_ZERO_ENTRIES, PETSC_TRUE);`,
the matrix `get_a_inv()` is now sparser:
``` 
row 0: (0, 0.5) 
row 1: (1, 0.5)
```
therefore vector vals has NOT size==block_size, but only 1 (only the diagonal value is nonzero in the test)

src/flow/darcy_flow_assembly.hh
Paulie14: This was a fast fix:
- in steady case, we want the zeros on the diagonal to be ignored (and not allocated)
- in unsteady case, if the zeros on the diagonal are ignored (and not allocated), then it crashes when assembling time terms (in `setup_time_term()`) because we need to write to the diagonal and not enough memory is allocated in that rows..

Any better suggestions, what to do with it?

src/flow/darcy_flow_mh.cc
Paulie14: agree, my bad..

Paulie14:
Here, we are solving issue #875.

Paulie14:
Please see the last commit 8f9b5a9 for the sparsity pattern.

Paulie14:
Ready to merge now.

PR #886, opened 2017-11-06
Df gmsh output

jbrezmorf:
Use OutputMesh in GMSH output.


src/io/output_mesh.cc
```diff
@@ -109,6 +109,33 @@ unsigned int OutputMeshBase::n_nodes()
...
+			node_idx[0] = i;
+			node_ids_->store_value( i, node_idx );
+		}
+	}
+	return node_ids_;
+}
+
+std::shared_ptr<ElementDataCacheBase> OutputMeshBase::get_element_ids_cache()
+{
+	if (!elem_ids_) {
```
jbrezmorf: Make OutputElement::get_node_ids, which returns original Node IDs if the mesh is same as original mesh, and returns indices of nodes in output mesh otherwise.
```
	id =	orig_mesh->node_vector.get_id(node_idx)
```

src/io/output_mesh.cc
```diff
@@ -109,33 +114,42 @@ unsigned int OutputMeshBase::n_nodes()
...
+	unsigned int elm_idx[1];
+	unsigned int node_idx[1];
+	elem_ids_ = std::make_shared< ElementDataCache<unsigned int> >("elements_ids", (unsigned int)1, 1, this->n_elements());
+	node_ids_ = std::make_shared< ElementDataCache<unsigned int> >("node_ids", (unsigned int)1, 1, this->n_nodes());
+	OutputElementIterator it = this->begin();
+	for (unsigned int i = 0; i < this->n_elements(); ++i, ++it) {
+		elm_idx[0] = it->get_idx();
+		elem_ids_->store_value( i, elm_idx );
+
+		std::vector< unsigned int > node_list = it->node_list();
```
jbrezmorf: 1. remove is_refined, is_equal_with_orig methods, test here directly
2. remove get_idx, treat both element and node idx in the same way
3. use orig IDs only if mesh is orig (can use this also for parallel case later on)

PR #888, opened 2017-11-10
Df parallel vtk output

dflanderka:


jbrezmorf:
- make method is_created, set nodes_, offsets_,connectivity_ to default null shared ptr in constructors
-

src/io/output_mesh.cc
```diff
@@ -198,6 +198,39 @@ bool OutputMesh::refinement_criterion()
...
+
+    orig_element_indices_ = std::make_shared<std::vector<unsigned int>>(n_local_elements);
+
+    offsets_ = std::make_shared<ElementDataCache<unsigned int>>("offsets", (unsigned int)ElementDataCacheBase::N_SCALAR, 1, n_local_elements);
+    Node* node;
+    unsigned int ele_id = 0,
+                 connect_id = 0,
+                 offset = 0,    // offset of node indices of element in node vector
+                 li;            // local node index
+    auto &offset_vec = *( offsets_->get_component_data(0).get() );
```
jbrezmorf: - loop over elements to fill offset and map of orig node IDX to submesh node IDX (use unordered hash map)
- loop over elements to fill connectivity
- loop over map to fill nodes

src/io/output_mesh.cc
```diff
@@ -514,6 +547,12 @@ void OutputMeshDiscontinuous::refine_aux_element(const OutputMeshDiscontinuous::
     }
 }
 
+void OutputMeshDiscontinuous::create_sub_mesh()
```
jbrezmorf: We should make a method that constructs discont mesh from continuous in order to have just one

src/io/output_mesh.cc
jbrezmorf: ... rather support only discont mesh for parallel submesh, there we can construct mesh in just two loops without using node map.

src/io/output_time.hh
```diff
@@ -194,8 +192,16 @@ public:
     template <typename T>
     ElementDataCache<T> & prepare_compute_data(std::string field_name, DiscreteSpace space_type, unsigned int n_rows, unsigned int n_cols);
 
-    /// Add information about field to map of used interpolations.
-    void add_field_interpolation(DiscreteSpace space_type, const std::string &field_name, unsigned int value_type);
+    inline Mesh * get_orig_mesh() {
```
jbrezmorf: Check if we can remove orig mesh from OutputTime at all so it should work only with OutputMesh and data caches.

src/io/output_time.hh
```diff
@@ -194,8 +192,16 @@ public:
...
+    	return _mesh;
+    }
+
+    /**
+     * Return if file format is in parallel / serial version.
+     */
+    inline bool is_parallel() {
```
jbrezmorf: Remove "is_parallel".

src/fields/field.impl.hh
```diff
@@ -582,7 +582,7 @@ void Field<spacedim,Value>::compute_field_data(OutputTime::DiscreteSpace space_t
 	typedef typename Value::element_type ElemType;
 
     /* It's possible now to do output to the file only in the first process */
-    if( stream->get_rank() != 0) {
+    if( (stream->get_rank() != 0) && (!stream->is_parallel())) {
```
jbrezmorf: move call of stream->get_output_mesh_ptr here and return null_ptr for case where current processor performs no output (and thus have no output mesh).

src/fields/fe_value_handler.cc
```diff
@@ -114,7 +114,7 @@ void FEValueHandler<elemdim, spacedim, Value>::value_list(const std::vector< Poi
 	ASSERT_EQ( point_list.size(), value_list.size() ).error();
 
     DOFHandlerBase::CellIterator cell = dh_->mesh()->element( elm.idx() );
-	dh_->get_loc_dof_indices(cell, dof_indices);
+	dh_->get_dof_indices(cell, dof_indices);
```
jbrezmorf: Check and change other calls to get_loc_dof_indices (in methods: interpolate, calculate_native_values, fill_data_to_cache), comment this method since it is currently wrong (the shift by loc_offset_ is not enough for local <-> global mapping. Only local to global mapping should be used in future. 

Check if we can merge methods where this method is used, in particular  calculate_native_values should use FeValueHandler.

jbrezmorf:
- move integration tests into 03_generic_output direcotry
- make separate tests for vtk_pralallel_<type>, type is one of element data, corner data, node data
  tests will be based on test 12/06 with output of pressure_p0, velocity, and anisotropy (tensor data)
- describe aim of the test in description string and/or in YAML comment

- other changes into separate issue, include comments in code + change from vector<int> to vector<unsigned int> in dofhadler::get_dof_indices.

PR #898, opened 2017-12-05
Df improve input

dflanderka:
- simplify inpute structure
- use default values of selected records


tests/02_generic_input/08_field_init_check.yaml
```diff
@@ -55,6 +55,7 @@ problem: !Coupling_Sequential
         - velocity_p0
```
jbrezmorf: Test default for the stream in 03_generic_output dir. Add comment in description.

PR #905, opened 2018-01-08
Df improve output

jbrezmorf:


src/fields/equation_output.cc
```diff
@@ -213,7 +220,7 @@ void EquationOutput::make_output_mesh()
     if(stream_->enable_refinement()) {
         if(it) {
             // create output meshes from input record
-            auto output_mesh = std::make_shared<OutputMeshDiscontinuous>(*stream_->get_orig_mesh(), *stream_->get_output_mesh_record());
+            auto output_mesh = std::make_shared<OutputMeshDiscontinuous>(*mesh_, *stream_->get_output_mesh_record());
```
jbrezmorf: Set discont flag at beginning and apply ||= for parallel and refinement case.

src/fields/equation_output.cc
```diff
@@ -235,20 +242,19 @@ void EquationOutput::make_output_mesh()
     // create output mesh identical with the computational one
 	std::shared_ptr<OutputMeshBase> output_mesh;
 	bool discont = (used_interpolations_.find(OutputTime::CORNER_DATA) != used_interpolations_.end());
-	if (discont || it || stream_->is_parallel()) {
-		output_mesh = std::make_shared<OutputMeshDiscontinuous>(*stream_->get_orig_mesh());
+	if (discont || it || parallel) {
```
jbrezmorf: Test just for discont.

PR #915, opened 2018-01-22
JHy - change arg logic within runtest script

janhybs:
Previously:
 - All yaml files contained in the processing folder were executed
   even if no explicit setting was present in config.yaml

Current:
 - Only yaml files which are specified in the config.yaml
   will be executed.

This change only affect runtest script if directories are passed, e.g.:
  `runtest -n 1 tests/10_darcy`

Directly specifying yaml file will ALWAYS execute the test, e.g.:
  `runtest-n 1 tests/10_darcy/01_source.yaml`

### DIFF
--- a/src/python/scripts/runtest_module.py
+++ b/src/python/scripts/runtest_module.py
@@ -6,7 +6,7 @@ import time
 import sys
 # ----------------------------------------------
 from scripts.pbs import pbs_control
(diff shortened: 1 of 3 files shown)


jbrezmorf:
Cherry picked to 3.0.0_dev.

PR #925, opened 2018-02-21
Js fill fe values

### LINKED ISSUE #887: FE System and FE simplification

jstebel:
Move fill_fe_values() to FEValues.
Remove template parameter spacedim from FiniteElement.

Partially resolves #887.


src/fem/finite_element.hh
```diff
@@ -343,7 +352,7 @@ protected:
      * @param q Quadrature rule.
      * @param flags Update flags.
      */
-    virtual FEInternalData *initialize(const Quadrature<dim> &q);
+    FEInternalData *initialize(const Quadrature<dim> &q);
```
jbrezmorf: Move this to FeValues together with FeInternalData.

src/fem/finite_element.hh
```diff
@@ -299,7 +308,7 @@ public:
      * @param p    Point of evaluation.
      * @param comp Number of vector component.
      */
-    virtual double basis_value(const unsigned int i,
+    double basis_value(const unsigned int i,
```
jbrezmorf: Rename and reimplement to shape_value. Same for basis_grad. Called just once per assembly loop so until we need to call it many times (in XFEM) we can live with slightly reduced efficiency.

src/fem/finite_element.hh
jstebel: Done.

src/fem/finite_element.hh
jstebel: Done.

PR #926, opened 2018-02-22
Df time step func

dflanderka:


src/tools/time_governor.cc
```diff
@@ -55,7 +55,19 @@ const Tuple & TimeGovernor::get_input_time_type(double lower_bound, double upper
 
 
 const Record & TimeGovernor::get_input_type() {
-	return Record("TimeGovernor",
+    static const Tuple &dt_step =
+        Tuple("DtLimits", "Time dependent changes in min_dt and max_dt limits.")
```
jbrezmorf: Use Tuple for all three components.

src/tools/time_governor.cc
```diff
@@ -364,20 +422,88 @@ void TimeGovernor::init_common(double init_time, double end_time, TimeMark::Type
...
+
+        if (min < time_step_precision) {
+    		THROW(ExcTimeGovernorMessage() << EI_Message("'min_dt' in 'dt_limits' smaller than machine precision.\n") );
+        }
+        if (max < min) {
+    		THROW(ExcTimeGovernorMessage() << EI_Message("'max_dt' in 'dt_limits' smaller than 'min_dt'.\n") );
+        }
+
+    	dt_limits_table_.push_back( DtLimitRow(time, min, max) );
+    	this->marks().add(TimeMark(time, this->equation_fixed_mark_type()));
```
jbrezmorf: Introduce a switch for setting the times in the time table as fixed time marks.

src/tools/time_governor.cc
```diff
@@ -548,6 +674,17 @@ void TimeGovernor::next_time()
     lower_constraint_ = min_time_step_;
     lower_constraint_message_ = "Permanent minimal constraint, in next time.";
     upper_constraint_message_ = "Permanent maximal constraint, in next time.";
+
+    if (dt_limits_iter_->time == step().end()) set_permanent_constraint();
```
jbrezmorf: Use inequality.

src/tools/time_governor.cc
```diff
@@ -364,20 +422,88 @@ void TimeGovernor::init_common(double init_time, double end_time, TimeMark::Type
...
+    // fill table limits with two records (start time, end time)
+    dt_limits_table_.push_back( DtLimitRow(init_time_, min_dt, max_dt) );
+    dt_limits_table_.push_back( DtLimitRow(end_time_, min_dt, max_dt) );
+    dt_limits_iter_ = dt_limits_table_.begin();
+
+    set_permanent_constraint();
+}
+
+
+void TimeGovernor::set_dt_limits( double min_dt, double max_dt, Input::Array dt_limits_list)
```
jbrezmorf: Try to merge these two functions, pass an empty array as a default value for the corresponding input key.

Issue #538, opened 2016-03-08
Under surface depth


jbrezmorf:
Introduce a parameter D into FieldFormula, taking the distance of the point to the surface. 
Two other (optional) keys must be added into FeildFormula record:

surface_direction (default [0,0,-1]) the vector used to project evaluation point onto the surface
surface_region - the name of region set considered as the surface. If there are more intersection points the closest one is used.
max_depth - If no intersection is found, we use this value. By default we use 

Implementation should use an array storing the depth of the center of every element. And reusing this information in further evaluations.

Implementace: BIH tree pro XY projekce povrchovych elementu, 
Pro dany xy,z bod najit povrchovy element -> spocitat projekci bodu na povrch. Pomale pokud bychom to provadeli pro kazdy kvadraturni bod zvlast. Bylo by vhodne mit pro to cache nebo to napocitat do FieldFE. S tim souvisi moznost pouzit ve FieldFormula nejake jine Field, asi mozne pokud budeme mit FiedlValueCache

jbrezmorf:
Implementation in own class SurfaceDepth.
1. Have this object separately in every FieldFormula. Precompute the BIH tree for surface elements, but evaluate projection for every quadrature point independently (no precomputing)
2. ... possible optimizations

jbrezmorf:
Nemyslim, ze je potreba konstruovat celou submesh, staci do BIH Tree pridat jen hranicni elementy z vybranych regionu.
Zavislost BIHTree na Mesh je zbytecna, podivej se na

https://github.com/flow123d/bih/blob/master/src/bih_tree.cc

Zde jsem tu tridu exportoval do samostatneho projektu s Pythonim rozhranim, bylo by fajn to ve Flow123d mit stejne.

jbrezmorf:
Co se tyce projekce:
Je zadany vektor N, ten se vydeli velikosti, aby byl jednotkovej a vytvori se bazove vektory pro kolmou plochu:

Ex = (1,0,0), Ey=(0,1,0)
Ta = Ex - (Ex . N) N
Ta = Ta / |Ta|
Tb = Ey - (Ey . N) N - (Ex . Ta) Ta
Tb = Tb/|Tb|

Udela se matice M s vektory Ta a Tb v radcich.

Libovolny 3d bod P se pak projektuje do souradnic (a,b) v rovine kolme k N, prenasobenim tou matici:

(a,b) = M . P

[...]

PR #930, opened 2018-03-06
Df surface depth

dflanderka:
Issue #538


src/fields/field_formula.cc
```diff
@@ -44,9 +45,17 @@ const Input::Type::Record & FieldFormula<spacedim, Value>::get_input_type()
...
-                                        "Formula can contain variables ```x,y,z,t``` and usual operators and functions." )
+                                        "Formula can contain variables ```x,y,z,t,d``` and usual operators and functions." )
 			//.declare_key("unit", FieldAlgorithmBase<spacedim, Value>::get_input_type_unit_si(), it::Default::optional(),
 			//							"Definition of unit.")
+			.declare_key("surface_direction", it::String(), it::Default("\"0 0 1\""),
+										"The vector used to project evaluation point onto the surface.")
+			.declare_key("surface_region", it::String(), it::Default::optional(),
+										"The name of region set considered as the surface. You have to set surface region if you "
+										"want to use formula variable ```d```.")
+			.declare_key("max_depth", it::Double(0.0), it::Default("1e06"),
```
jbrezmorf: Why such limit? Not very useful. To deal with failed projections we should 
implement sort of projection in BIH  to find nearest box. Then we just take e.g. barycenter of corresponding triangle. If to far we can report an error. Alternatively we can introduce a key for default value of 'd'.

src/fields/field_formula.cc
```diff
@@ -152,15 +170,28 @@ bool FieldFormula<spacedim, Value>::set_time(const TimeStep &time) {
...
+
+
 /**
  * Returns one value in one given point. ResultType can be used to avoid some costly calculation if the result is trivial.
  */
 template <int spacedim, class Value>
 typename Value::return_type const & FieldFormula<spacedim, Value>::value(const Point &p, const ElementAccessor<spacedim> &elm)
 {
+
+    auto p_depth = this->eval_depth_var(p);
```
jbrezmorf: p_depth should be always double.

src/fields/field_formula.cc
```diff
@@ -173,21 +204,36 @@ template <int spacedim, class Value>
...
             }
     }
 }
 
 
+template <int spacedim, class Value>
+arma::vec FieldFormula<spacedim, Value>::eval_depth_var(const Point &p)
+{
+	arma::vec p_depth(spacedim+1);
+	for (unsigned int i=0; i<spacedim; i++) p_depth(i) = p(i);
```
jbrezmorf: Why this?

src/fields/field_formula.cc
```diff
@@ -173,21 +204,36 @@ template <int spacedim, class Value>
...
+	if (surface_depth_ && has_depth_var_) {
+		// add value of depth
+		p_depth(spacedim) = std::min( surface_depth_->compute_distance(p), max_depth_ );
+	} else {
+		p_depth(spacedim) = 0;
```
jbrezmorf: just return 'p'

src/fields/field_formula.cc
```diff
@@ -173,21 +204,36 @@ template <int spacedim, class Value>
...
 
 
+template <int spacedim, class Value>
+arma::vec FieldFormula<spacedim, Value>::eval_depth_var(const Point &p)
+{
+	arma::vec p_depth(spacedim+1);
+	for (unsigned int i=0; i<spacedim; i++) p_depth(i) = p(i);
+	if (surface_depth_ && has_depth_var_) {
+		// add value of depth
+		p_depth(spacedim) = std::min( surface_depth_->compute_distance(p), max_depth_ );
```
jbrezmorf: create extended variable vec only in this case

src/fields/surface_depth.cc
```diff
@@ -0,0 +1,126 @@
...
+    bih_tree_.add_boxes( boxes );
+    bih_tree_.construct();
+}
+
+double SurfaceDepth::compute_distance(arma::vec3 point)
+{
+	double distance = std::numeric_limits<double>::max();
+	arma::vec projection = m_ * point;
+	arma::vec3 project_point;
+	project_point(0) = projection(0); project_point(1) = projection(1); project_point(2) = 0;
```
jbrezmorf: Just make 'm_' 3x3 matrix with last row zero.

src/fields/surface_depth.cc
```diff
@@ -0,0 +1,126 @@
...
+
+	RegionSet region_set = mesh->region_db().get_region_set(surface_region);
+    if (region_set.size() == 0)
+        THROW( RegionDB::ExcUnknownSet() << RegionDB::EI_Label(surface_region) );
+
+    // make element boxes
+    unsigned int i=0;
+    unsigned int i_node;
+    arma::vec3 project_node("0 0 0");
+    for( ElementFullIter it = mesh->bc_elements.begin(); it != mesh->bc_elements.end(); ++it) {
```
jbrezmorf: We should iterate through all elements to allow projections to internal interfaces. Depends on
changes in mesh. We should merge mesh changes ASAP.

src/fields/surface_depth.cc
```diff
@@ -0,0 +1,126 @@
...
+        if (it->region().is_in_region_set(region_set)) {
+        	if (it->dim()==3) { // dimension 3 is forbidden for surface elements
+        		THROW( ExcSurfaceProjection()
```
jbrezmorf: Throw for all dimensions != 2.

src/fields/surface_depth.cc
```diff
@@ -0,0 +1,126 @@
...
+        		    			<< EI_Message("Surface region contains one or more elements of dimension 3. Maximal dimension is 2!") );
+        	} else if (it->dim()<2) continue; // dimension 0,1 has no effect
+        	ASSERT_EQ(it->n_nodes(), 3);
+
+        	arma::mat coords(3,3);
+        	coords.col(0) = it->node[0]->point();
+        	arma::vec projection = m_ * it->node[0]->point();
+        	project_node(0) = projection(0); project_node(1) = projection(1);
+            BoundingBox bb(project_node);
```
jbrezmorf: Why not treat col(0) in the loop.

src/fields/surface_depth.cc
```diff
@@ -0,0 +1,126 @@
...
+            for(i_node=1; i_node<it->n_nodes(); i_node++) {
+            	coords.col(i_node) = it->node[i_node]->point();
+                arma::vec project_coords = m_ * it->node[i_node]->point();
+                project_node(0) = project_coords(0); project_node(1) = project_coords(1);
+                bb.expand(project_node);
+            }
+            boxes.push_back(bb);
+            nodes_coords_.push_back(coords);
```
jbrezmorf: Armadillo small fixed vectors are very inefficient in terms of memory, causing also bad cache usage. We should use Armor library due to Petr Kral to store fixed vectors as fixed double arrays but still work with them using armadillo functions. Same should be considered for the BoundingBox points.

src/fields/surface_depth.hh
```diff
@@ -0,0 +1,65 @@
...
+	BIHTree bih_tree_;
+
+	/// normal vector of surface plane
+	arma::vec3 surface_norm_vec_;
+
+	/// projection matrix
+	arma::mat m_;
+
+	/// vector of nodes coordinates of elements above which BIH tree is created
+	std::vector<arma::mat> nodes_coords_;
```
jbrezmorf: Why not store pointers to elements or pointers to element nodes?

src/mesh/bih_tree.hh
```diff
@@ -46,16 +46,19 @@ public:
...
 	 */
-	BIHTree(Mesh* mesh, unsigned int soft_leaf_size_limit = 20);
+	BIHTree(unsigned int soft_leaf_size_limit = 20);
 
 	/**
 	 * Destructor
 	 */
 	~BIHTree();
 
+	void add_boxes(std::vector<BoundingBox> boxes);
```
jbrezmorf: make 'boxes' const reference.

jbrezmorf:
An integration test should be made and we have to measure difference between evaulation of FieldConst, FieldFormula without 'd' , FieldFormula with 'd'. We should try to optimize calculation of depth, however in near future we should allow precomputing the FieldFormula to suitable FieldFE.

src/fields/field_formula.cc
```diff
@@ -173,21 +204,36 @@ template <int spacedim, class Value>
...
             for(unsigned int col=0; col < this->value_.n_cols(); col++) {
-                envelope(row,col) = this->unit_conversion_coefficient_ * parser_matrix_[row][col].Eval(point_list[i].memptr());
+                envelope(row,col) = this->unit_conversion_coefficient_ * parser_matrix_[row][col].Eval(p_depth.memptr());
             }
     }
 }
 
 
+template <int spacedim, class Value>
+arma::vec FieldFormula<spacedim, Value>::eval_depth_var(const Point &p)
```
jbrezmorf: Make it as an inlined function in CC file.

src/fields/field_formula.cc
```diff
@@ -137,7 +155,7 @@ bool FieldFormula<spacedim, Value>::set_time(const TimeStep &time) {
                     xprintf(UsrErr, "ParserError: %s\n in the FieldFormula[%d][%d] == '%s'\n at the input address:\n %s \n",
                         parser_matrix_[row][col].ErrorMsg(),
                         row,col,formula_matrix_.at(row,col).c_str(),
-                        value_input_address_.c_str());
+                        value_input_address.c_str());
```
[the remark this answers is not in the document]
dflanderka: Done, FunctionParser hasn't compare operator. New flag was added.

src/fields/surface_depth.cc
dflanderka: We fill matrix and construct bounding box at the same step, we must initialize bounding box first.

jbrezmorf:
- rename key 'global_observe_search_radius' to 'global_snap_radius' describe that it is used to 
  limit maximum snapping distance in  various search operations, in particular it is used in ObservePoint   to find closest mesh element and in FieldFormula to find closest surface element in plan view (Z projection).
- SurfaceDepth: for distance larger then the limit, throw an input exception: Report:
    - region and time of the field, address v YAML
    -  XY coordinates of projected point, vzdalenost nejblissiho elementu.

- integration test for FieldFormula with 'd':
    - final intergration test on some existing 3d mesh, compare results against FieldFormula usign just 'z'
      for flat surface.
    - some benchmarks usign the same test but larger mesh (100k elements) and compare time of 'assembly_steady_mh_matrix' speed against simple FieldFormula and FieldConstant.

src/fields/surface_depth.cc
```diff
@@ -0,0 +1,158 @@
...
+    			<< EI_Message("Region " + surface_region + " contains no boundary element! Probably bulk region was set.") );
+    }
+
+    bih_tree_.add_boxes( boxes );
+    bih_tree_.construct();
+}
+
+double distance_from_abscissa(arma::vec3 a, arma::vec3 b, arma::vec3 &p)
+{
+	arma::vec3 u = b-a;
```
jbrezmorf: p_projected = x.cols(0,1)
proj_ref = RefElement<2>::snap_point(p_projected)
proj_3d = a_mat.cols(0,1) * proj_ref
dist = arma:norm(proj_3d - p)

src/fields/surface_depth.cc
```diff
@@ -0,0 +1,158 @@
...
+		a_mat.col(1) = vb;
+		a_mat.col(2) = surface_norm_vec_;
+		arma::vec3 b_vec = point - coords.col(0);
+		arma::vec3 x;
+		arma::solve(x, a_mat, b_vec);
+		if ( (x(0)>=0) && (x(1)>=0) && (x(0)+x(1)<=1) ) { // point is in triangle
+			double new_distance = -x(2);
+			if ( (new_distance>=0) && (new_distance<distance) || (snap_dist>0.0) ) distance = new_distance;
+			snap_dist = 0.0;
+		} else if (snap_dist > 0.0) { // check snap distance of point from triangle
```
jbrezmorf: Perform clipping only if no element is found. Second loop over searched_elements

src/fields/surface_depth.cc
```diff
@@ -0,0 +1,158 @@
...
+double SurfaceDepth::compute_distance(arma::vec3 point)
+{
+	double distance = std::numeric_limits<double>::max();
+	double snap_dist = std::numeric_limits<double>::max();
+	arma::vec3 project_point;
+	arma::vec3 proj_to_surface_plane;
+	project_point.subvec(0,1) = m_ * point;
+	project_point(2) = 0;
+
+	std::vector<unsigned int> searched_elements;
```
jbrezmorf: Move into class and use reserve to allocate vector.

PR #932, opened 2018-03-08
Jb raw ngh

jbrezmorf:


jbrezmorf:
After PE_intersections, separate neighboring info into separate raw output and we shall merge it into master. 
I assume that both raw outputs will be possible only for serial runs.

Paulie14:
Question:
Should we output the ngh data also at all output times, like we do with the flow data?
As long as the mesh does not change, ngh data can be output only once...

Paulie14:
Sofar, I left the raw ngh output in the same manner as raw flow output,
i.e. add new table for every output time.
If it is ok, it can be merged.

jbrezmorf:
Move key for raw_ngh_output into Mesh record. Move implementation into Mesh class. Optional raw output call in setup_topology.

jbrezmorf:
Move key for raw_ngh_output into Mesh record. Move implementation into Mesh class. Optional raw output call in setup_topology.

jbrezmorf:
Just single integration test in 04_generic_mesh.

Paulie14:
Done.

Paulie14:
I do not know why, but on the second run on CI (no changes made), every test passed.
Every test passed locally on my PC too.
Can be merged now..

PR #933, opened 2018-03-09
Df field time shift

dflanderka:
- allow time shift for FieldElementwise etc.


src/fields/field_elementwise.cc
```diff
@@ -47,8 +47,11 @@ const Input::Type::Record & FieldElementwise<spacedim, Value>::get_input_type()
 		//		"Definition of unit.")
         .declare_key("default_value", IT::Double(), IT::Default::optional(),
                 "Allow set default value of elements that have not listed values in mesh data file.")
+        .declare_key("read_from_time", IT::Double(), IT::Default("0.0"),
```
jbrezmorf: rename key: "read_time_shift"

PR #935, opened 2018-03-19
Df optimize mesh elements

dflanderka:
- store boundary and bulk elements in common structure


src/mesh/mesh.h
```diff
@@ -357,9 +342,58 @@ class Mesh {
     /// Maximal distance of observe point from Mesh relative to its size
     double global_observe_radius() const;
 
-    /// Number of elements read from input.
-    unsigned int n_all_input_elements_;
+    /// Initialize element_vec_, set size and reset counters of boundary and bulk elements.
+    void init_element_vector(unsigned int size);
 
+    /// Returns range (of indexes in element_vector_) of boundary or bulk elements
+    inline pair<unsigned int, unsigned int> get_element_range(bool boundary=false) const {
```
jbrezmorf: bulk_elements(), boundary_elements()
```
return  make_iterator_range(make_iter(ElementAccessor(mesh,0), make_iter(ElementAccessor(mesh, n_bulk_elements))
```

src/mesh/mesh.h
```diff
@@ -357,9 +342,58 @@ class Mesh {
...
+    }
+
+    /// Returns count of boundary or bulk elements
+    inline unsigned int n_elements(bool boundary=false) const {
+    	if (boundary) return boundary_size_;
+    	else return bulk_size_;
+    }
+
+    /// Begin iterator to bulk part of element vector
+    inline ElementIterator bulk_begin() {
```
jbrezmorf: replace by ranges.

src/mesh/mesh.h
```diff
@@ -357,9 +342,58 @@ class Mesh {
...
+    /// End iterator to boundary part of element vector
+    inline ElementIterator boundary_end() {
+    	return element_vec_.end();
+    }
+
+    // For each node the vector contains a list of elements that use this node
+    vector<vector<unsigned int> > node_elements_;
+
+    /// For element of given elem_id returns index in element_vec_ or (-1) if element doesn't exist.
+    inline int elem_index(int elem_id) const
```
jbrezmorf: Make a class for bi-dir map, make its instance, e.g. Mesh::element_ids. Kid of this method would be method of that bi-dir map class.

src/mesh/mesh.h
jbrezmorf: Rename `Tools::GeneralIterator` to `Iter` and add there templated method make_iter for simplified cration of iterators from accessors. Try to implement also subtract() or overloaded - in accessors, so that the range class can return size().

src/mesh/mesh.h
```diff
@@ -357,9 +342,58 @@ class Mesh {
     /// Maximal distance of observe point from Mesh relative to its size
     double global_observe_radius() const;
 
-    /// Number of elements read from input.
-    unsigned int n_all_input_elements_;
+    /// Initialize element_vec_, set size and reset counters of boundary and bulk elements.
+    void init_element_vector(unsigned int size);
 
+    /// Returns range (of indexes in element_vector_) of boundary or bulk elements
```
jbrezmorf: Implement Mesh::make_element_ac(index) as replacement of mesh->element[index].

src/mesh/mesh.h
jbrezmorf: Try to implement own Range class (having begin, end and size methods). Test if such a class can be used in the for statement.

jbrezmorf:
- remove includes of XYZ_impl.hh in mesh.hh
- move XYZ_impl.hh (only for accessor classes) to XYZ.hh, inlined methods.

src/coupling/balance.cc
```diff
@@ -208,12 +210,12 @@ void Balance::lazy_initialize()
 	// construct vector of regions of boundary edges
     for (unsigned int loc_el = 0; loc_el < mesh_->get_el_ds()->lsize(); loc_el++)
     {
-        Element *elm = mesh_->element(mesh_->get_el_4_loc()[loc_el]);
+        ElementAccessor<3> elm = const_cast<Mesh *>(mesh_)->element_accessor( mesh_->get_el_4_loc()[loc_el] );
         if (elm->boundary_idx_ != nullptr)
         {
-            FOR_ELEMENT_SIDES(elm,si)
+        	for(unsigned int si=0; si<elm->n_sides(); si++)
```
jbrezmorf: for(auto side_it : elm->side_range())

PR #947, opened 2018-05-09
Pk bih tree calibrated

jbrezmorf:


jbrezmorf:
Will merge sucessor JB-armor-fix instead.

PR #948, opened 2018-05-09
Pe xfem master

Paulie14:
Initial pull request to see changes made in xfem development.


src/fem/fe_values.cc
```diff
@@ -139,6 +146,7 @@ template<unsigned int dim, unsigned int spacedim>
 void FEValuesBase<dim,spacedim>::allocate(Mapping<dim,spacedim> & _mapping,
         Quadrature<dim> & _quadrature,
         FiniteElement<dim> & _fe,
+        unsigned int ndofs,
```
jbrezmorf: Why ndofs is set explicitely?

src/fem/fe_values.cc
```diff
@@ -505,8 +514,20 @@ FEValues<dim,spacedim>::FEValues(Mapping<dim,spacedim> &_mapping,
          UpdateFlags _flags)
 :FEValuesBase<dim, spacedim>()
 {
-    this->allocate(_mapping, _quadrature, _fe, _flags);
+    this->allocate(_mapping, _quadrature, _fe, _fe.n_dofs(), _flags);
 
+    //UpdateFlags map_update_flags = this->data.update_flags;
+    
+    if(typeid(_quadrature) == typeid(QXFEM<dim,spacedim>)){
```
jbrezmorf: We should make this more general, either add a method "reset_quadrature" or introduce derived class where quadrature is not fixed. Some similarities to p-adaptivity.

src/fem/fe_values.hh
```diff
@@ -149,6 +149,13 @@ public:
 //      */
 //     std::vector<std::vector<arma::mat::fixed<spacedim,spacedim> > > shape_grad_vectors;
 
+    /**
+     * @brief Divergence of shape functions (for vectorial finite elements).
+     * 
+     * TODO: compute div from grads; (is not trivial in rt xfem)
+     */
+    std::vector<std::vector<double> > shape_divergence;
```
jbrezmorf: Not clear what is stored in vectors. Neither why this is necessary for XFEM in contradiction to  FEM.

src/fem/fe_values.hh
```diff
@@ -256,9 +263,10 @@ public:
      * @param _fe The finite element.
      * @param flags The update flags.
      */
-    void allocate(Mapping<dim,spacedim> &_mapping,
+    virtual void allocate(Mapping<dim,spacedim> &_mapping,
```
jbrezmorf: Why making this virtual, if it is called only from constructor which is separate for FEM and XFEM case.

src/fem/fe_values.hh
```diff
@@ -453,7 +475,7 @@ protected:
      *
      * @param fe_data Precomputed finite element data.
      */
-    void fill_data(const FEInternalData &fe_data);
+    virtual void fill_data(const FEInternalData &fe_data);
```
jbrezmorf: Why this must be virtual? Where it is called from?

src/fem/finite_element.hh
```diff
@@ -39,9 +39,7 @@ template<unsigned int dim, unsigned int spacedim> class FEValuesBase;
 template<unsigned int dim, unsigned int spacedim> class FEValuesData;
 template<unsigned int dim> class FE_P_disc;
 template<unsigned int dim> class Quadrature;
-
-
-
+template<unsigned int dim, unsigned int spacedim> class XFEValues;
```
jbrezmorf: This is too interdependent. Seems that Quadrature is not necessary. What is used by FEValues etc. should be public.
FE_P_disc should not be necessary as it is derived from FiniteElement (indirectly).

src/fem/mapping_p1.cc
```diff
@@ -298,7 +298,17 @@ auto MappingP1<dim,spacedim>::clip_to_element(BaryPoint &barycentric) -> BaryPoi
     return RefElement<dim>::clip(barycentric);
 }
 
-
+template<unsigned int dim, unsigned int spacedim>
+bool MappingP1<dim,spacedim>::is_point_inside(const BaryPoint &bp){
```
jbrezmorf: There is same functionality in Element, we should deduplicate this. The right place is in Mapping.

src/flow/darcy_flow_mh.cc
```diff
@@ -1241,18 +1320,77 @@ void DarcyMH::make_serial_scatter() {
...
             }
-            for (auto ele : mesh_->bulk_elements_range()) {
+            else{
+                
+            //velocity
+            for (auto ele : mesh_->bulk_elements_range())
+            	for (unsigned int si=0; si<ele->n_sides(); si++)
+                    loc_idx[i++] = mh_dh.side_row_4_id[ mh_dh.side_dof( ele.side(si) ) ];
+            
+//             //enriched velocity
```
jbrezmorf: What about this commented code?

src/flow/darcy_flow_mh_output.cc
```diff
@@ -70,8 +70,12 @@ const it::Instance & DarcyFlowMHOutput::get_input_type_specific() {
         .copy_keys(OutputSpecificFields::get_input_type())
         .declare_key("compute_errors", it::Bool(), it::Default("false"),
                         "SPECIAL PURPOSE. Computing errors pro non-compatible coupling.")
+        .declare_key("python_solution", it::FileName::input(), it::Default::optional(),
+                        "SPECIAL PURPOSE. Python script for computing analytical solution.")
         .declare_key("raw_flow_output", it::FileName::output(), it::Default::optional(),
                         "Output file with raw data from MH module.")
+        .declare_key("output_linear_system", it::Bool(), it::Default("false"),
```
jbrezmorf: This should be implemented in LinSys, shouldn't be?

src/flow/field_velocity.hh
```diff
@@ -0,0 +1,212 @@
...
+ * This program is free software; you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License version 3 as published by the
+ * Free Software Foundation. (http://www.gnu.org/licenses/gpl-3.0.en.html)
+ * 
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+ *
+ * 
+ * @file    field_velocity.hh
```
jbrezmorf: What is in this file?

src/flow/mh_dofhandler.cc
```diff
@@ -37,7 +46,14 @@ edge_4_loc(nullptr),
 row_4_edge(nullptr),
 edge_ds(nullptr),
 el_ds(nullptr),
-side_ds(nullptr)
+side_ds(nullptr),
+row_4_sing(nullptr),
```
jbrezmorf: We are going to remove mh_dofhandler so no code here please. Probably there should be an xfem_dofhandler. Possibly DiscreteSpace interface is simple enough to support it.

src/mesh/mesh.cc
```diff
@@ -697,6 +697,10 @@ MixedMeshIntersections & Mesh::mixed_intersections() {
     if (! intersections) {
         intersections = std::make_shared<MixedMeshIntersections>(this);
         intersections->compute_intersections();
+//             static_cast<IntersectionType> (
```
jbrezmorf: Remove?

src/quadrature/quadrature.hh
```diff
@@ -109,7 +110,8 @@ protected:
      * To be filled by the constructors of the derived classes.
      */
     std::vector<double> weights;
-
+    
+    friend class QXFEMFactory;
```
jbrezmorf: Why necessary?

src/fem/finite_element.hh
Paulie14: Quadrature, FE_P_disc removed.
problem with FEValues, XFEValues left for future.

src/quadrature/quadrature.hh
Paulie14: removed

src/fem/mapping_p1.cc
Paulie14: solved

Issue #784, opened 2017-01-24
slow sorptions


jbrezmorf:
Linearni sorpce volaji reinit pri kazdem vyhodnoceni, nepouzivaji se tabulky. Vysledkem je:

 vypocet bez sorpci 6s

vypocet s linearni sorpci: 134s

jbrezmorf:
smb://jan.brezina@bacula.nti.tul.cz/intec/projekty/flow123d/input_data/bug_data/slow_sorption

jstebel:
Sorptions are calculated by interpolation tables only if solubility or table_limits are given in the yaml file. Otherwise the sorption is calculated elementwise using a nonlinear solver.

jbrezmorf:
We need to implement automatic extension of the tables.
TODO:
- if tables are not initialized pass through the conc values, determine range (0, max) of values and initailize tables for interval (0, 2*max). If solubility is set use range given by this parameter as is done now.
- during each calculation keep maximum of values out of tables, if this maximum is greater then table range, recompute tables using twice the new maximum as the upper bound

Paulie14:
Finished by 1be74ab5f58faf3a383c6f370505225f2f4f1573
Can be closed when pull request  #968 is accepted.

This also fixed the following problem (now in test 30_sorption/11_direct_vs_interpolation):
- computing this simple problem with solubility limit did not give same results when computed directly or using interpolation
- problem was the solubility limit condition in direct computation, where it was done in the concentration axial system instead of total mass system
- is fixed in ccdfd66422a270ca25619ec21668b2ea425572d6

jbrezmorf:
done

PR #968, opened 2018-07-19
Pe slow sorption

Paulie14:
- separate functionality of solubility_limit and table_limit
- automatic increase of interpolation table range according to max conc (this is now by default)
- isotherm reinit call available for single isotherm (one can reinit, the others don't)
- simplify code in sorptions
- move implementation of isotherm to cc file

- fixed issue of slow sorptions #784


src/reaction/sorption_base.cc
```diff
@@ -409,22 +412,126 @@ void SorptionBase::update_solution(void)
...
+}
+
 void SorptionBase::make_tables(void)
 {
+    START_TIMER("SorptionBase::make_tables");
     try
     {
+        
         ElementAccessor<3> elm;
         BOOST_FOREACH(const Region &reg_iter, this->mesh_->region_db().get_region_set("BULK") )
```
jbrezmorf: use c++11 for

src/reaction/sorption_base.cc
```diff
@@ -409,22 +412,126 @@ void SorptionBase::update_solution(void)
...
-                    isotherms[reg_idx][i_subst].make_table(n_interpolation_steps_);
+                    call_make_table = true;
+                    subst_table_limit = 2*max_conc[reg_idx][i_subst];
+//                     DebugOut().fmt("limit: max conc\n");
+                }
+                // make tables if isotherm reinited
+                else if(call_reinit)
+                {
+                    call_make_table = true;
+                    subst_table_limit = isotherms[reg_idx][i_subst].table_limit();
```
jbrezmorf: Clarify

PR #976, opened 2018-08-21
Optimize PadeApproximant by splitting timestep.

### LINKED ISSUE #974: Pade approximation produces negative concentration
In radioactive decays, if time step is large and halflife small, the Pade approximation of exponential can be too inaccurate, leading to negative or too inexact values of concentrations. For example, for the default choice of polynomial degrees, m=1, n=3, the approximating function is:

R(x) = (1+x/4) / (1 - 3/4*x + x^2/4 - x^3/24)

For x<-4 we get R(x) < 0, while for x~2.6258 the denominator is zero, so R(x) is not defined. On the other hand, if |x|<1 then |R(x)-exp(x)|<0.01.

In decays, x is the product of kinetic constant (= ln(2)/halflife) and time step. We should provide some bounds within which Pade approximant is OK. Then divide the time step into L substeps so that the argument is not far from zero and use (R(x/L))^L instead of R(x).

Proposed solution:
- Implement PadeApproximant method that estimates the region of good approximation (e.g. find roots of nominator and denominator closest to zero and in this interval find upper and lower bounds to get error between exp(x) and R(x) less than some tolerance). This method should be called in constructor (as well as calculation of nominator and denominator coefs).
- In PadeApproximant::update_solution estimate the largest eigenvalue of system_matrix_, set the number of substeps so that lambda*step_/L is in the admissible interval. Then compute the solution_matrix_ as before and power it to L.

Test problem for verification:
bacula-intec/jan.stebel/00_share_flow123d_issues/#974

jstebel:
The artificial time step is chosen so that the difference between
Pade approximant and exponential is less than 0.001.
This fixes problems with negative concentrations obtained after reaction update with large time step.

Resolves #974.


jstebel:
Reworked in #982.

PR #977, opened 2018-08-23
Js dof handler 2

jstebel:
Parallel distribution of dofs: supported nodal and internal dofs.

Status:
- New class DiscreteSpace that provides FiniteElement for every mesh element.
- Each Dof handler objects creates indices of ghost elements and their owning processors. This should once be a part of the parallel mesh. Should it be done already now?
- Distribution of dofs is split into several parts: initialization, distribution on local elements, communication/update, possibly sequentialization.
- The data structures are parallel with exception to cell_starts


src/fem/discrete_space.cc
```diff
@@ -0,0 +1,85 @@
...
+ * the terms of the GNU General Public License version 3 as published by the
+ * Free Software Foundation. (http://www.gnu.org/licenses/gpl-3.0.en.html)
+ * 
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+ *
+ * 
+ * @file    dofhandler.hh
+ * @brief   Declaration of class which handles the ordering of degrees of freedom (dof) and mappings between local and global dofs.
```
jbrezmorf: Update file header.

src/fem/discrete_space.hh
```diff
@@ -0,0 +1,111 @@
...
+ * This program is free software; you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License version 3 as published by the
+ * Free Software Foundation. (http://www.gnu.org/licenses/gpl-3.0.en.html)
+ * 
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+ *
+ * 
+ * @file    dofhandler.hh
```
jbrezmorf: Update file header.

jbrezmorf: Not clear what is the purpose of DiscreteSpace class. Just wrap FE for elements of different dimension? For p-refined or XFEM case it could be better to have elementwise additional data as part of the DofHandler itself.

src/fem/dofhandler.hh
```diff
@@ -284,47 +202,42 @@ public:
...
+     * @param sequential If true then each processor will have information about all dofs.
+     * @param offset     The offset.
      */
-    void distribute_dofs(FiniteElement<1> &fe1d,
-    		FiniteElement<2> &fe2d,
-    		FiniteElement<3> &fe3d,
-    		const unsigned int offset = 0);
+    void distribute_dofs(std::shared_ptr<DiscreteSpace> ds,
+                         bool sequential = false,
+                         const unsigned int offset = 0);
```
jbrezmorf: Do we need offset anymore?

src/fem/dofhandler.hh
```diff
@@ -346,6 +259,13 @@ public:
 	 * @param loc_nb Local index of neighbour.
 	 */
 	inline LongIdx nb_index(int loc_nb) const { return nb_4_loc[loc_nb]; }
+	
+	/**
+     * @brief Return number of dofs on given cell.
+     * 
+     * @param cell Cell accessor.
+     */
+	unsigned int n_dofs(ElementAccessor<3> cell) const;
```
jbrezmorf: What is CellIterator vs. ElementAccessor. Documentation is bit confusing.
Anyway, is it enough to provide iterating over the local cells with a dedicated DoFHandler iterator with automaticaly updated informations, then we can remove both get_loc_dof_indices and n_dofs and can support also getting particular dofs for faces.

... can be part of further development, possibly can be done by David.

src/fem/dofhandler.hh
jstebel: Probably not. This was originally meant for systems of finite elements, which are now possible through FESystem.

src/fem/dofhandler.hh
jstebel: CellIterator removed and replaced by ElementAccessor<3>.

PR #982, opened 2018-09-06
Simplify LinearODESolver, remove its derived classes (analytic, Pade).

jstebel:
Use arma::expmat to compute matrix exponential.
Update inaccurate test results (verified by using current method with smaller time step).


jstebel:
Should the unused files (linear_ode_analytic.*, pade_approximant.*) be deleted?

PR #983, opened 2018-09-07
Js seq dof handler

### LINKED ISSUE #981: Use parallel DoFHandler
New parallel DofHandler can not be used because of FieldFE is lack support for be based on parallel DoFHanlder and parallel outputs and not automatically used.
Related to:  #814, 

We need:
- [x] (JS) Have local numbering of Dofs incluing ghost dofs. Method DofHandler::get_loc_dof_indices working for ghost elements.

- [x]  (DF) Ranges (own_range, local_range - includs ghost cells), Accessor and Iterator for DofHandler Cells. FInish in #1014.

- [x] (DF) Replace VecSeqDouble with VecMPI. VecMPI constructor takes communicator as parameter, with default value PETSC_COMM_SELF. For this value we create just VecSeqWithArray. Replace VecSeqDouble constructors with a function caling VecMPI constructor with PETSC_COMM_SELF. ... for better readability.

- [x] (DF) VecMPI: move to la subdir

- [x] (DF) FieldFE (should know if it is based on serial or parallel DofHandler ??). For evaluation use 
DofHandler::get_loc_dof_indices. 

- [x] (DF) #1015 Use parallel DH in TransportDG:
- [x] (JS) #1012 Support for ghost values in VecMPI.

- [x] (DF, JB) FieldFE, method scatter to ghost dofs (must be called after values are changed and before evaluation). This calls scatter of VecMPI which in turn calls scatter for the VecGhost.

- [x] (JS) Support to create a serial DofHandler from a parallel DofHandler and associated Petsc scatter.

- [x] (DF) #1016 Output of parallel fields through serial output streams.

- [x] Apply parallel DH and output fields to Reactions.
- [x] Apply parallel DH and output fields to Transport. Need assembly based on DH. #1036

- [x]  #1013 Apply parallel DH and output fields to DracyFlow.

- [ ] Remove remaining (temporary) usages of cell_accessor_from_element (mainly in FieldFE)

jstebel:
Partially resolves #981:
- Create sequential dof handler from parallel one.
- Create corresponding scatter to create sequential vector of dofs from parallel one.
- Map of local+ghost dof indices to global indices.


src/fem/dofhandler.cc
```diff
@@ -542,14 +373,37 @@ void DOFHandlerMultiDim::distribute_dofs(std::shared_ptr<DiscreteSpace> ds,
...
 
 void DOFHandlerMultiDim::create_sequential()
 {
-  ASSERT(dof_indices.size() > 0).error("Attempt to create sequential dof handler from uninitialized or already sequential object!");
+  if (dh_seq_ != nullptr) return;
+  
+  if ( !is_parallel_ )
+  {
+      dh_seq_ = std::make_shared<DOFHandlerMultiDim>(*this);
+      return;
```
jbrezmorf: Missing scatter in the sequential case?

jstebel: Fixed in new commit.

PR #985, opened 2018-09-11
Df join field interpolated

dflanderka:
- merge FieldInterpolatedP0 with FieldFE


src/mesh/bih_tree.cc
```diff
@@ -41,6 +41,7 @@ BIHTree::~BIHTree() {
 
 
 void BIHTree::add_boxes(const std::vector<BoundingBox> &boxes) {
+	main_box_ = BoundingBox( boxes[0].min() ); // fix problem with undefined main_box_
```
jbrezmorf: Initialize only if elements_ are empty.

jbrezmorf: Add explicit constructor call for main_box_ in the BIH_Tree constructor.
Initialize only if elements_ are empty.

Modify BoundingBox default constructor to initialize to NaNs ... in the parallel DH branch.

jbrezmorf: Add explicit constructor call for main_box_ in the BIH_Tree constructor.
Initialize only if elements_ are empty.

Modify BoundingBox default constructor to initialize to NaNs ... in the parallel DH branch.

PR #999, opened 2018-10-12
Change tag specification syntax

janhybs:
- build user image under different name based on <tag> value
 - add image_tag file, which is the default value for the <tag> value
 - auto-personalize image if not exists


bin/fterm
```diff
@@ -2,27 +2,27 @@
 # This script can personalize flow123d images (adding user/group, adding keys, setting up git)
 # usage ./fterm [options] [dbg|rel] [action] [arguments]
 #   options:
-#     -v|--verbose
-#     |-d|--debug    - turn on debug mode (set -x)
-#     
+#     -v|--verbose    turn on debug mode
+#     --set-x         turn on debug mode (set -x)
```
jbrezmorf: I prefer original -d/--debug flag, "-set x" is BASH mess

PR #1011, opened 2018-11-01
Df remove create field

dflanderka:
- replace methods create_field and fill_output_data in DarcyFlow


jbrezmorf:
Motivation of these changes is not clear to me. We should discus that at Friday.

jbrezmorf:
Motivation was removal of fill_output_data usage from Darcy flow output.
Need to be daone together with direct output of computational fields and simplification of computation of postprocessed fields.

dflanderka:
Replace of methods create_field and fill_output_data is solved in branch DF_set_velocity_field. Thios branch is obsolete and can be closed now.

PR #1027, opened 2018-11-23
JHy improve runtest

janhybs:
Improve runtest, auto detect tty mode

 - add option --show-output controlling when to display output


janhybs:
outdated

PR #1029, opened 2018-12-10
Js simplify dh

### LINKED ISSUE #1009: Simplify DofHandler
- [x] Introduce unordered_map : global_to_local, mapping global element indices after partitioning to local own and ghost indices. Nevertheless mark this variable as **obsolete**. It is just temporary solution how to map ElementAccessors to DHCellAccessors until we have parallel mesh. So this mapping should be used only in 
corresponding method dh ->DHCellAccessor(ElementAccessors).
- [x] Replace/remove row_4_el, ghost_4_loc, el_4_loc.
- [x] Are other vectors in DH necessary? In particular edg_4_loc, node_4_loc, ghost_proc, ghost_proc_el (private?), cell_starts_seq, dof_indices_seq
- [x] cell_starts should map only local =own and ghost elements

jstebel:
Resolves #1009.


jstebel:
Reduced usage of global_to_local_el_idx_. Ready to merge.

PR #1034, opened 2019-01-28
Js default solver amg

jstebel:
Set default options of PETSc solver for positive definite matrix.
Use Hypre/BoomerAMG in both sequential and parallel mode.


jbrezmorf:
We should make two more changes:
1. Move predefined PETSC options to individual equations. Create a specific Selection of predefined options in each  equation (or possibly in some base classes).
2. Make LynSys input parametrized by the selection type and introduce a key to select predefined  PETSc options.

PR #1038, opened 2019-02-01
Df parallel reaction output

dflanderka:
- apply parallel DH and output fields to Sorption and DualPorosity
- remove gathering of data


src/reaction/dual_porosity.cc
```diff
@@ -213,16 +206,32 @@ void DualPorosity::initialize_fields()
   //setting fields in data
   data_.set_mesh(*mesh_);
   
+  //initialization of DOF handler
+  static FE_P_disc<0> fe0(0);
+  static FE_P_disc<1> fe1(0);
+  static FE_P_disc<2> fe2(0);
+  static FE_P_disc<3> fe3(0);
+  shared_ptr<DOFHandlerMultiDim> dh = make_shared<DOFHandlerMultiDim>(*mesh_);
```
dflanderka: Maybe use better initialization of DH.

jbrezmorf: Create sngle common dof handler for reactions in OperatorSplitting constructor.
Implement ReactionTerm::set_dh(DH) and use it to pass the common DH to the top reaction term instance.
Modify all reaction subclasses to pass DH down through the reaction tree in their initialize() method.

src/reaction/dual_porosity.cc
```diff
@@ -213,16 +206,32 @@ void DualPorosity::initialize_fields()
...
+  shared_ptr<DiscreteSpace> ds = make_shared<EqualOrderDiscreteSpace>( mesh_, &fe0, &fe1, &fe2, &fe3);
+  dh->distribute_dofs(ds);
+
   //initialization of output
   data_.output_fields.set_components(substances_.names());
   data_.output_fields.set_mesh(*mesh_);
   data_.output_fields.output_type(OutputTime::ELEM_DATA);
   data_.conc_immobile.setup_components();
+  vconc_immobile = new Vec [substances_.size()];
```
jbrezmorf: This was used only for creating the scatter and can be removed together with its initialization.

PR #1051, opened 2019-03-14
Js fix loc dof indices

### LINKED ISSUE #1049: Bug in DOFHandlerMultiDim::get_loc_dof_indices
- [ ] merge to master partial fix using subtraction of loffset
- [x] change get_loc_dof_indices and get_dof_indices to be protected and taking local element index as the parameter. Reimplement corrsponding methods in the DHCellAccessor to call these methods.
- [x] Change distribution of DOFs to produce vector dof_indices containing local dof indices instead of global ones, build vector local_to_global_dof_idx instead. Change implementation of  get_loc_dof_indices and get_dof_indices to deal correctly with ghost dofs.

jstebel:
Resolves #1049.

Some comments/questions added in the diff.

### DIFF
--- a/src/fem/dh_cell_accessor.hh
+++ b/src/fem/dh_cell_accessor.hh
@@ -53,7 +53,7 @@ public:
 
     /// Return local index to element (index of DOF handler).
     inline unsigned int local_idx() const {
-    	ASSERT_LT_DBG(loc_ele_idx_, dof_handler_->el_ds_->lsize()+dof_handler_->ghost_4_loc.size()).error("Local element index is out of range!\n");
+        ASSERT_LT_DBG(loc_ele_idx_, dof_handler_->global_to_local_el_idx_.size()).error("Local element index is out of range!\n");
         return loc_ele_idx_;
     }
 
--- a/src/fem/dh_cell_accessor.hh
+++ b/src/fem/dh_cell_accessor.hh
@@ -74,14 +74,16 @@ public:
      *
      * @param indices Vector of dof indices on the cell.
      */
-    unsigned int get_dof_indices(std::vector<int> &indices) const;
+    unsigned int get_dof_indices(std::vector<int> &indices) const
+    { return dof_handler_->get_dof_indices( *this, indices ); }
 
     /**
      * @brief Returns the indices of dofs associated to the cell on the local process.
      *
      * @param indices Array of dof indices on the cell.
      */
-    unsigned int get_loc_dof_indices(std::vector<LongIdx> &indices) const;
+    unsigned int get_loc_dof_indices(std::vector<LongIdx> &indices) const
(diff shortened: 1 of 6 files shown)


src/fem/dh_cell_accessor.hh
```diff
@@ -53,7 +53,7 @@ public:
 
     /// Return local index to element (index of DOF handler).
     inline unsigned int local_idx() const {
-    	ASSERT_LT_DBG(loc_ele_idx_, dof_handler_->el_ds_->lsize()+dof_handler_->ghost_4_loc.size()).error("Local element index is out of range!\n");
+        ASSERT_LT_DBG(loc_ele_idx_, dof_handler_->global_to_local_el_idx_.size()).error("Local element index is out of range!\n");
```
jstebel: If we want to use DHCellAccessor::get_dof_indices() for sequential DH, then we need to define local_idx() for all elements. So now we take local_idx() as the index in the map global_to_local_el_idx_.

src/fields/field_fe.cc
```diff
@@ -646,7 +646,7 @@ void FieldFE<spacedim, Value>::calculate_native_values(ElementDataCache<double>:
 	else mesh = dh_->mesh();
 	for (auto ele : mesh->elements_range()) { // remove special case for rank == 0 - necessary for correct output
 		if (this->boundary_domain_) dof_size = value_handler1_.get_dof_indices( ele, dof_indices_ );
-		else dof_size = dh_->get_loc_dof_indices( ele, dof_indices_ );
+		else dof_size = dh_->cell_accessor_from_element(ele.idx()).get_loc_dof_indices( dof_indices_ );
```
jstebel: I guess using cell_accessor_from_element() is a temporary solution. In fact, we should iterate using dh_->local_range(), which in the case of sequential DH includes all elements. See also other places in the pull request.

src/fem/dh_cell_accessor.hh
jbrezmorf: Do you mean sequential DH but parallel mesh? Then yes. But that should be communicated by sequential() method isn't it? Anyway I hope that finally the sequential DH will not used.

jbrezmorf: There should be local_to_global_el_idx, isn't it?

src/fields/field_fe.cc
jbrezmorf: Yes this is temporary. Reason are fields with boundary domain. One possible solution is to have sort of restriction of a DH to the boundary. Then FieldFE could be agnostic about its domain.

src/fem/dh_cell_accessor.hh
jstebel: There is no local_to_global_element map. We need just the number of elements on which the DH has information about dofs.

Issue #1048, opened 2019-03-11
Terminate all MPI processes on  error


jbrezmorf:
- remove parameters of Application and ApplicationBase constructors
- remove call of run form  ApplicationBase::init 
- explicit call of run from main method
- implement Application::terminate -> call MPI_Abort
- create instance of Application out of the try block in the main method
- call terminate in the case of exception

dflanderka:
Example of Error output:

===
------------------------------------------------------------
Found 1 yaml file/s
Running in LOCAL mode
Running 2 cases (excluding all tags in set ['disabled'])
------------------------------------------------------------
Case 01 of 02
Running: 1 x 02_generic_input/03_field_descriptors
bin/flow123d -s tests/02_generic_input/03_field_descriptors.yaml -o tests/02_generic_input/test_results/03_field_descriptors.1
Done    | elapsed time 0:00:01:562, memory used  142.87MB
[ERROR] |   Command ended with 1! (pid=610)
            Full command:
            /flow123d/bin/flow123d -s /flow123d/tests/02_generic_input/03_field_descriptors.yaml -o /flow123d/tests/02_generic_input/test_results/03_field_descriptors.1
    ------------------------------------------------------------
    Output from file /flow123d/tests/02_generic_input/test_results/03_field_descriptors.1/job_output.log
    ############################################################
    ##
    ##                   [EI_MPI_Rank_TAG*] = -1
    ##
    ##                   ** Stacktrace **
    ##                     0  SchurComplement::create_inversion_matrix()
    ##                     1  SchurComplement::form_schur()
    ##                     2  SchurComplement::resolve()
    ##                     3  SchurComplement::compute_residual()
    ##                     4  DarcyMH::solve_nonlinear()
    ##                     5  DarcyMH::zero_time_step()
    ##                     6  HC_ExplicitSequential::run_simulation()
    ##                     7  Application::run()
    ##                     8  ApplicationBase::init(int, char**)
    ##                     9  main
    ##
    ##                   --------------------------------------------------------
    ## application called MPI_Abort(MPI_COMM_WORLD, 1) - process 0
    ## [unset]: write_line error; fd=-1 buf=:cmd=abort exitcode=1
    ## :
    ## system msg for write_line failure : Bad file descriptor
    ############################################################
------------------------------------------------------------
Case 02 of 02
Running: 2 x 02_generic_input/03_field_descriptors
[...]

PR #1052, opened 2019-03-14
Df error kill processes

dflanderka:
- solve #1048

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -349,6 +349,8 @@ void Application::run() {
         }
 
     }
+
+    this->after_run();
 }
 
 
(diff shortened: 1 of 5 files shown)


src/main.cc
```diff
@@ -74,8 +74,8 @@ it::Record & Application::get_input_type() {
 
 
 
-Application::Application( int argc,  char ** argv)
-: ApplicationBase(argc, argv),
+Application::Application( char ** argv)
```
jbrezmorf: No need for argv.

jbrezmorf:
Remove argv.
Otherwise good, can merge.

src/main.cc
dflanderka: We can't remove it, argv[0] is used as argument of PythonLoader::initialize method.

jbrezmorf: OK so just pass argv[0], that's the program path.

PR #1054, opened 2019-03-15
Js hm coupling

jstebel:
Basis for integration of mechanics/poroelasticity.


jstebel:
Can be merged in this form? (Already checked some time ago.)

Issue #1047, opened 2019-03-04
Simplification of integration test output


jbrezmorf:
@janhybs:

IMO  our current approach with testing generates overwhelmingly long console  output which complicates things I suggest to display short logs and  offer to view full log in form of a artifacts, example here: https://stackoverflow.com/questions/41974070/how-to-download-build-output-files-from-jenkins-ui-console-itself

jbrezmorf:
Good idea, only requirement is that artifacts are on fingertips from the test output page.

janhybs:
From the way i see it, there are 3 alterations to be done:
 - [ ] **update runtest** so it only displays output on error (or when forced to)
 - [ ] **reconfigure Jenkins** server so it offers artifacts
 - [ ] **notify developers** so they can get used to it

I propose to suppress output for other parts of the build system as well (use `make -s` more, tune down output of unit tests). At a glance, developer should clearly see the most important stuff (success, failure) but developers also **must** have access to logs, allowing him to look into the things deeper.

jbrezmorf:
Absolutely agree. Only review to stdout, and full logs into atrefacts if they are available to be associated to individual test jobs.

janhybs:
I've reconfigured Jenkins and updated runtest script.
Console now looks like [this](http://ciflow.nti.tul.cz:8080/job/Flow123d-ci2runner-debug-test-20_solute_fv/10/console)

And artifacts looks like [this](http://ciflow.nti.tul.cz:8080/job/Flow123d-ci2runner-debug-test-20_solute_fv/10/)

PR #1068, opened 2019-04-03
JHy runtest update

janhybs:
Implementing #1047, contains Jenkins job builder configuration update and python runtest script update


jbrezmorf:
- Do we still get full output from the runtest when running manually?
- Should be merged into master.

janhybs:
Yes, nothing has changed, output is in the `job_output.log` in the test output directory file as before.

PR #1069, opened 2019-04-08
Pe balance

Paulie14:
- use sequential matrices for sources balance
- replace 'MatMultTransposeAdd' calls with own direct computations
- unify 'add_source_xxx' functions to a single one


src/coupling/balance.cc
```diff
@@ -474,7 +457,6 @@ void Balance::finish_source_assembly(unsigned int quantity_idx)
 	chkerr(MatAssemblyEnd(region_source_matrix_[quantity_idx], MAT_FINAL_ASSEMBLY));
 	chkerr(MatAssemblyBegin(region_source_rhs_[quantity_idx], MAT_FINAL_ASSEMBLY));
 	chkerr(MatAssemblyEnd(region_source_rhs_[quantity_idx], MAT_FINAL_ASSEMBLY));
-	chkerr(MatMultTranspose(region_source_rhs_[quantity_idx], ones_, region_source_vec_[quantity_idx]));
```
jstebel: I think ones_ is no more used and can be removed.

src/coupling/balance.hh
```diff
@@ -506,9 +493,9 @@ private:
     std::vector<std::vector<double> > fluxes_in_;
     std::vector<std::vector<double> > fluxes_out_;
     std::vector<std::vector<double> > masses_;
-    std::vector<std::vector<double> > sources_;
     std::vector<std::vector<double> > sources_in_;
     std::vector<std::vector<double> > sources_out_;
+    std::vector<double> temp_sources_;
```
jstebel: Is temp_sources_ used elsewhere than in calculate_cumulative()? If not, it should be a local variable in this method.

Paulie14: done

src/coupling/balance.cc
Paulie14: done, removed also ones_be_.

src/coupling/balance.cc
```diff
@@ -637,8 +590,13 @@ void Balance::calculate_cumulative(unsigned int quantity_idx,
...
 	chkerr(VecZeroEntries(boundary_vec));
+    
 	Vec temp;
-	chkerr(VecDuplicate(ones_be_, &temp));
+    chkerr(VecCreateMPI(PETSC_COMM_WORLD,
+			be_regions_.size(),
+			PETSC_DECIDE,
+			&temp));
+    
 	chkerr(MatMultAdd(be_flux_matrix_[quantity_idx], solution, be_flux_vec_[quantity_idx], temp));
```
Paulie14: to JB: this might be also implemented directly, without creating MPI objects and ```MatMultTranspose```.
```increment_fluxes_``` would be then computed locally and reduced in the output function similarly to ```increment_sources_```. Should be done now or be left for future?

Paulie14: I did minor simplifications.
The larger changes, including boundary edge indexing, are described in #1070.

PR #1077, opened 2019-05-20
Jb no term

jbrezmorf:
A simple way to use 'run' or 'exec' without a terminal, e.g. when used as a subprocess from python.


src/mesh/mesh.cc
```diff
@@ -1201,40 +1201,44 @@ void Mesh::distribute_nodes() {
...
-                n_ghost_nodes++;
-                ghost_node_flag[i_node] = true;
-            }
+            local_node_flag[i_node] = true;
+            if (i_proc < node_proc[i_node]) node_proc[i_node] = i_proc;
         }
     }
 
+    unsigned int n_own_nodes=0, n_local_nodes=0; // number of own and ghost nodes
+    for(uint i_proc : node_proc) if (i_proc == my_proc) n_own_nodes++;
```
jbrezmorf: Check for fix here.

Issue #628, opened 2016-07-15
Output observe values for every timestep


jbrezmorf:
Currently the observe values are output only in output times since e need a support in equations to 
update output fields only on observe elements. Necessary steps:
- support for parallel (fe) fields
- parallel output, in particular observe need to evaluate only local part of observe points and reduce results to the zero proc.
- derived/postrpocesed fields need support for update only on a given subset of elements (in order to do cheep observe output in every time)

jbrezmorf:
- [ ] For every input observe point get its processor and assign its local index on that processor. Store both local and global index of point. Create map from local to global numbering of points 
- [ ] Create parallel distribution of observe points. (global numbering is original numbering at input).
- [ ] Point accessor and range to iterate over local points. Point accessor provides loc_point_time_index ... 
      index into local par of the element data cache.
- [ ] Modify Field<>::observe_output to use loc_point_time_index instead of i_data.
- [ ] Observe::prepare_compute_data ... return local element data cache (alternative is that ObserPointAccessor provides directly method for storing the value, this way we completely hide data structures)
- [ ] Observe::output_time_frame -
	- increase i_time (time row in element data cache), save observe_values_time_ to separate vector.
        - if the table is full, perform gather and output to file

dflanderka:
Partly solved in #1078

jbrezmorf:
Make integration test:
- check boxes in the last comment if it match current implementation
- take some of ConvectiveTransport problems with larger mesh
- add support for observe output in all time steps
- compare times

PR #1078, opened 2019-05-29
Df parallel observe output

dflanderka:
- part of #628


src/io/observe.cc
```diff
@@ -257,6 +258,7 @@ void ObservePoint::find_observe_point(Mesh &mesh) {
 			observe_data_.element_idx_ = candidate_data.element_idx_;
 			observe_data_.local_coords_ = candidate_data.local_coords_;
 			observe_data_.global_coords_ = candidate_data.global_coords_;
+			observe_data_.proc_ = candidate_data.proc_;
```
dflanderka: Implement and use copy constructor of ObservePointData.

src/io/observe.cc
```diff
@@ -417,34 +438,60 @@ void Observe::output_header() {
 
 }
 
-void Observe::output_time_frame(double time) {
+void Observe::output_time_frame(double time, bool write_automaticcaly) {
```
jbrezmorf: rename 'write_automaticacly' to 'flush' (similar to stream)

PR #1083, opened 2019-06-11
Df simplify output times

dflanderka:


jbrezmorf:
ElementDataCache use internal time attribute if used for the input data, but this attribute is not used in the case of output data, the OutputTime attribute is used instead. This pull tries to unify the usage.

Problem is that in the case of input the times can be different, but in the output whole set of eval data caches must have the same output time corresponding to the time of the output frame.

jbrezmorf:
Won't merge. Not clear what was the intention of the more complicated time passing. to the ElementDataCache.

PR #1089, opened 2019-06-18
Js fix quadrature permutations

### LINKED ISSUE #1088: Wrong permutation of points in constructor of Quadrature.
Quadrature created from existing lower-dimensional subquadrature does not produce correct point coordinates. When used in FESideValues, the quadrature points differ from points in FEValues using the subquadrature.

jstebel:
Resolves #1088.

### DIFF
--- a/src/quadrature/quadrature.hh
+++ b/src/quadrature/quadrature.hh
@@ -224,7 +224,7 @@ Quadrature<dim>::Quadrature(const Quadrature<dim-1> &subq, unsigned int sid, uns
         
         //permute
         for (unsigned int i=0; i<RefElement<dim>::n_nodes_per_side; i++) {
-            pp(i) = p(RefElement<dim>::side_permutations[pid][i]);
+            pp(RefElement<dim>::side_permutations[pid][i]) = p(i);
         }
         
         el_bar_coords = RefElement<dim>::template interpolate<dim-1>(pp,sid);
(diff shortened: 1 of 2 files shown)


unit_tests/fem/quadrature_test.cpp
```diff
@@ -85,36 +85,70 @@ TEST(Quadrature, midpoint){
 }
 
 
+/// Map lower dimensional quadrature to element quadrature to element quadrature
```
Paulie14: fix comment :)

jbrezmorf:
Is this still related to the problem we arrive at during changes in the velocity calculation?

jstebel:
No. The problem arises only in Elasticity.

Od: Jan Brezina
Odesláno:úterý 18. června 2019 15:02
Komu: flow123d/flow123d
Kopie: Jan Stebel; Assign
Předmět: Re: [flow123d/flow123d] Js fix quadrature permutations (#1089)

Is this still related to the problem we arrive at during changes in the velocity calculation?

src/quadrature/quadrature.hh
```diff
@@ -224,7 +224,7 @@ Quadrature<dim>::Quadrature(const Quadrature<dim-1> &subq, unsigned int sid, uns
         
         //permute
         for (unsigned int i=0; i<RefElement<dim>::n_nodes_per_side; i++) {
```
jbrezmorf: Why there is the upper bound 'n_nodes_per_side' while the 'p' and 'pp' vectors have size 'dim'? Possibly improve readability by introducing a common constant within this method.

jstebel: For simplex it holds:

RefElement<dim>::n_nodes_per_side = dim

While side_permutations have dimension n_permutations x n_nodes_per_side, points of course have dimension dim. But I do not see which of these 2 labels for the same value should be preferred.

jstebel: Added assert and explanation in the beginning of the constructor.

PR #1093, opened 2019-06-19
Df fix field fe speed

dflanderka:
- change evaluation of FieldFE


dflanderka:
- dead-end development branch
- replaced with #1126 and #1164

PR #1105, opened 2019-07-10
Df simplify dg assembly

dflanderka:
- implement AssemblyDG class
- move methods of TransportDG (templated with dim) to this class


src/transport/assembly_dg.hh
```diff
@@ -0,0 +1,928 @@
...
+        if (cell.elm()->boundary_idx_ == nullptr) return;
+
+        for (unsigned int si=0; si<cell.elm()->n_sides(); si++)
+        {
+            const Edge *edg = cell.elm().side(si)->edge();
+            if (edg->n_sides > 1) continue;
+            // skip edges lying not on the boundary
+            if (edg->side(0)->cond() == NULL) continue;
+
+            if (edg->side(0)->dim() != dim-1)
```
jbrezmorf: This can be removed. Must test it on a problem with DG on mixed dimension (26/01).

src/transport/heat_model.hh
```diff
@@ -234,13 +234,32 @@ public:
...
     { return substances_; }
 
+    const vector<unsigned int> &get_subst_idx() override
+	{ return subst_idx; }
+
+    std::shared_ptr<Balance> balance() const override {
+        return this->balance_;
+    }
+
+    std::shared_ptr<FieldFE<3, FieldValue<3>::VectorFixed>> velocity_field_ptr() const override {
```
jbrezmorf: Try to get rid of virtual getters, some are already implemented in parent classes, others can be done in similar way.

Issue #1046, opened 2019-03-01
Simplify usage of FiniteElement


jbrezmorf:
Since FunctionSpace and FiniteElement are small objects and used in non-critical code (few initializations) we should rather pass them by value and get rid of shared_ptr<FunctionSpace> in FiniteElement and
shared_ptr<FiniteElement> in FeSystem, FEValues, DiscreteSpace, etc.

We should also introduce a shortcut for types templated by dimension, eg.

typedef std::tuple<FiniteElement<0>,  FiniteElement<1>, FiniteElement<2>, FiniteElement<3>> FiniteElement;

or more generaly

template <T>
use std::tuple<T<0>,  T<1>, T<2>, T<3>> Multidim<T>;

Then we can introduce multidim FeSystem wrapper etc.

jbrezmorf:
- passing by value currently not possible as FiniteElement as well as FunctionSpace are abstract classes
- rather encapsulate these into a MixedPtr<T> class that use shared pointers internally
- Mixed<T> would be used to store T<dim> directly into the tuple, useful for e.g. for FEValues

- idea: pass FeView directy to FieldFE::set_fe_data() instead of component index, seems too low level but we have to create FeValues and Views for the assembly anyway. However that probably implies to have these in data_ instead of Assembler instance.

jbrezmorf:
- we can merge FiniteElement and FESystem (other FiniteElements may be just functions) this allows passing FinteElement instances by value.
- DiscreteSpace should remain abstract (shared pointers used) but is not used directly.
- then dicontinuous finite elements may be created by a function instead of a template.

jbrezmorf:
Possible relics:
Can we remove FiniteElement:: update_each ?
Where is friend FESystem used in FiniteElement?

jstebel:
One FiniteElement class for all types
- Constructor takes parameters that uniquely define the FE: mapping type (now FEType), dofs_, function_space_, nonzero_components_, ...? Other internal variables will be automatically computed.
- rename FEType to FEMappingType, number of components is given by function space
- particular FE types will be generated by specific initialization functions

Merging FESystem with FiniteElement
- instead of several constructors use static methods to initialize vector, tensor, mixed system etc.

dflanderka:
Places that makes separate objects for different dimensions

Flow:
 - AssemblyMH class
 - DarcyFlowMHOutput::FEData struct

Transport:
 - FEObjects of DG transport
 - assemble_xxx methods of DG transport
[...]

Issue #1003, opened 2018-10-22
FieldFE for elements of dimension zero


jbrezmorf:
In order to allow boundary FieldFE we need support for evaluation on elements of dimension zero.
Currently it is handled by explicit specialization with leads to code duplicity.
1. Use conditions in the code instead of explicit specialization of the class
2. Try to get support for dimension 0 elements in : RefElement, Mapping, Quadrature, FEValues.

jbrezmorf:
- [ ] Replace explicit specialization of FEValueHandler by conditions in generic implementation.
Try: `MappingP1< (elemdim>0?elemdim:1), 3> *map_;` in header file.

- [ ] support for element of dimension zero in: RefElement, Mapping, Quadrature, FEValues.

jbrezmorf:
Is this problem only of current implementation of boundary FieldFE? 
It seems, that value method should possibly be different for evaluation on the boundary. 
Consider usage of a bulk field as a boundary condition.

dflanderka:
Wee need to fix specialized constructors of FE_RT0<0> and FE_RT0_disc<0>. Existing solution was created in #1108 (implementation of Mixed class), but it needs to improve for further development.

PR #1108, opened 2019-07-15
Jb simple fe

jbrezmorf:
Solve issue: #1046


src/tools/mixed.hh
```diff
@@ -0,0 +1,200 @@
...
+ * fe_values = FEValues<dim>(mapping.get<dim>(), gauss.get<dim>(), ...)
+ *
+ * All parameters of the Mixed<T>(...) constructor are forwarder (by value) to the T(...) constructor for every dimension.
+ * Only forwarding by value is supported due to problems with MixedPtr.
+ * Can not resolve copy constructor correctly.
+ *
+ */
+public:
+    template < template<Dim...> class TT>
+    static Mixed<T> cast_to_parent_template(const Mixed<TT> &other)
```
jbrezmorf: Try to introduce user conversion operator instead using template like:
```
template< template<DimParent ...> class Parent, is_convertible<TParent, T>>
operator Mixed<Parent>() const { ... }
```
just conversion like `Parent(other.get<0>())` should be used instead of the dynamic cast.

Test in unit test.

src/fem/dofhandler.cc
```diff
@@ -749,19 +749,19 @@ SubDOFHandlerMultiDim::SubDOFHandlerMultiDim(std::shared_ptr<DOFHandlerMultiDim>
...
     ASSERT_DBG( fe_sys0 != nullptr ).error("sub_handler assumes that dof handler uses FESystem<0>!");
     ASSERT_DBG( fe_sys1 != nullptr ).error("sub_handler assumes that dof handler uses FESystem<1>!");
     ASSERT_DBG( fe_sys2 != nullptr ).error("sub_handler assumes that dof handler uses FESystem<2>!");
     ASSERT_DBG( fe_sys3 != nullptr ).error("sub_handler assumes that dof handler uses FESystem<3>!");
-    ds_ = std::make_shared<EqualOrderDiscreteSpace>( mesh_,
-                                                     fe_sys0->fe()[component_idx].get(),
-                                                     fe_sys1->fe()[component_idx].get(),
-                                                     fe_sys2->fe()[component_idx].get(),
-                                                     fe_sys3->fe()[component_idx].get() );
+    MixedPtr<FiniteElement> sub_mixed(fe_sys0->fe()[component_idx],
```
jbrezmorf: Simplify by, changing ds.fe() to return MixedPtr<FESystem>.

src/fem/fe_values.cc
```diff
@@ -552,6 +554,20 @@ void FEValues<dim,spacedim>::reinit(ElementAccessor<3> & cell)
...
 
+MixedPtr<FEValues> mixed_fe_values(
+        Mixed<MappingP1> &mapping,
+        Mixed<Quadrature> &quadrature,
+        MixedPtr<FiniteElement> fe,
+        UpdateFlags flags)
+{
+	std::shared_ptr< FiniteElement<0> > fe0 = std::make_shared< FE_P<0> >(0);
+    return MixedPtr<FEValues>(
+      std::make_shared<FEValues<0>>(mapping.get<0>(), quadrature.get<0>(), *fe0, flags),
```
jbrezmorf: Try to resolve FeRT0<0> issue and make this code consistent for dim 0.
If can not be simply resolved make a note in the issue #1003.

src/flow/darcy_flow_mh.cc
```diff
@@ -400,17 +400,19 @@ void DarcyMH::initialize() {
 		std::shared_ptr< FiniteElement<2> > fe2_cr = std::make_shared<FE_CR<2>>();
 		std::shared_ptr< FiniteElement<3> > fe3_cr = std::make_shared<FE_CR<3>>();
 // 	    static FiniteElement<0> fe0_sys = FE_P_disc<0>(0); //TODO fix and use solution with FESystem<0>( {fe0_rt, fe0_disc, fe0_cr} )
-        static FESystem<0> fe0_sys( {fe0_disc, fe0_disc, fe0_cr} );
-		static FESystem<1> fe1_sys( {fe1_rt, fe1_disc, fe1_cr} );
-		static FESystem<2> fe2_sys( {fe2_rt, fe2_disc, fe2_cr} );
-		static FESystem<3> fe3_sys( {fe3_rt, fe3_disc, fe3_cr} );
-		std::shared_ptr<DiscreteSpace> ds = std::make_shared<EqualOrderDiscreteSpace>( mesh_, &fe0_sys, &fe1_sys, &fe2_sys, &fe3_sys);
+		FESystem<0> fe0_sys( {fe0_disc, fe0_disc, fe0_cr} );
```
jbrezmorf: Resolve RT<0>.

src/fem/dofhandler.cc
dflanderka: Method ds.fe() changed, but we can't simplify using of 'sub_mixed' variable because this object is create above one of components of FESystem (stored in other Mixed).

src/flow/darcy_flow_mh_output.cc
```diff
@@ -452,11 +451,31 @@ void DarcyFlowMHOutput::l2_diff_local(DHCellAccessor dh_cell,
...
+    //Mixed<Quadrature> q_base = quad;
+    UpdateFlags flags = update_values | update_JxW_values | update_quadrature_points;
+    fe_values = MixedPtr<FEValues>(
+    	      std::make_shared<FEValues<0>>(mapp.get<0>(), quad.get<0>(), *fe_p0.get<0>(), flags),
+    		  std::make_shared<FEValues<1>>(mapp.get<1>(), quad.get<1>(), *fe_p0.get<1>(), flags),
+    		  std::make_shared<FEValues<2>>(mapp.get<2>(), quad.get<2>(), *fe_p0.get<2>(), flags),
+    		  std::make_shared<FEValues<3>>(mapp.get<3>(), quad.get<3>(), *fe_p0.get<3>(), flags)
+    	      );
+    //fe_values = mixed_fe_values(mapp, q_base, fe_p0, flags);
+    fv_rt = MixedPtr<FEValues>(
```
jbrezmorf: Try to debug this particular problem, since it seems, that copy constructor for Quadrature is OK. So there is possibly still some problem in the Mixed conversion constructor.

src/tools/mixed.hh
```diff
@@ -0,0 +1,221 @@
...
+        		T<0>(other.get<0>()),
+				T<1>(other.get<1>()),
+				T<2>(other.get<2>()),
+				T<3>(other.get<3>()));
+    }
+
+    template < template<Dim...> class TT>
+    Mixed(const Mixed<TT> &other)
+    : _MixedBase<T>(
+            other.get<0>(),
```
jbrezmorf: Possibly try explicit call of the copy constructor here:
`T<0>(other.get<0>())` as in the cast method

PR #1113, opened 2019-07-24
Df consistent time unit

jbrezmorf:


src/input/type_tuple.cc
```diff
@@ -69,6 +69,19 @@ bool Tuple::operator==(const TypeBase &other) const
 
 const Tuple &Tuple::close() const {
     data_->closed_=true;
+
```
jbrezmorf: Add assert here and in Record to throw when closing an empty Tuple or Record.

jbrezmorf: ... need a change in unit test.

PR #1116, opened 2019-08-08
Pe fix balance tomaster

Paulie14:
Fix balance of fluxes.
some tests fail on 'time limit exceeded', otherwise it is OK

### DIFF
--- a/src/coupling/balance.cc
+++ b/src/coupling/balance.cc
@@ -222,7 +222,7 @@ void Balance::lazy_initialize()
             {
                 Boundary *b = elm.side(si)->cond();
                 if (b != nullptr){
-                    unsigned int ele_side_uid = get_boundary_edge_uid(elm.side(si));
+                    LongIdx ele_side_uid = get_boundary_edge_uid(elm.side(si));
                     be_id_map_[ele_side_uid] = be_id;
                     be_regions_.push_back(b->region().boundary_idx());
                     be_id++;
--- a/src/coupling/balance.cc
+++ b/src/coupling/balance.cc
@@ -18,6 +18,7 @@
 
 #include <iostream>
 #include <iomanip>
+#include <unordered_map>
 
 #include "system/system.hh"
 #include "system/sys_profiler.hh"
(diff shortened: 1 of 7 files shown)


src/coupling/balance.cc
```diff
@@ -18,6 +18,7 @@
 
 #include <iostream>
 #include <iomanip>
+#include <unordered_map>
```
jstebel: Already included in balance.hh.

src/coupling/balance.hh
```diff
@@ -419,6 +421,9 @@ private:
 	/// Format double value of csv output. If delimiter is space, align text to column.
 	std::string format_csv_val(double val, char delimiter, bool initial = false);
 
+    /// Computes unique id of local boundary edge from local element index and element side index
+    inline unsigned int get_boundary_edge_uid(SideIter side)
```
jstebel: Return type should be LongIdx.

src/coupling/balance.hh
```diff
@@ -471,7 +476,10 @@ private:
     /// Vectors for calculation of mass (n_bulk_regions).
     Vec *region_mass_vec_;
 
-    /// Number of boundary region for each local boundary edge.
+    /// Maps unique identifier of (local bulk element idx, side idx) to local boundary edge.
```
jstebel: Improve documentation so that it is clear, that the LongIdx key is the result of get_boundary_edge_uid(s), where s is a side of a locally owned element.

src/coupling/balance.cc
jbrezmorf: Never mind its better practice to include everything used in the actual source file.

src/coupling/balance.hh
```diff
@@ -277,7 +279,7 @@ public:
      * possibly done when we will have a boundary mesh.
 	 */
 	void add_flux_matrix_values(unsigned int quantity_idx,
-			unsigned int boundary_idx,
+			SideIter side,
```
jbrezmorf: Would be better to pass DHCellSide which also know the dofs, so we can avoid passing them explicitly in separate parameter.

src/coupling/balance.hh
```diff
@@ -277,7 +279,7 @@ public:
      * possibly done when we will have a boundary mesh.
```
jbrezmorf: Fix doc. Remove TODO.

Issue #1114, opened 2019-07-26
Bug in side.cond() when using it in 'for DHCellSide : cell.side_range()', (Release).


Paulie14:
In the particular place in transport DG, the function ```side.cond()``` always returns null,
when using ```side``` directly from ```for (DHCellSide side : cell.side_range())```.
When the side is constructed explicitly, or when the inline methods for ```side``` and ```cond``` are enrolled, it returns the BC correctly.

This happens only in release version, so it seems that the compiler somehow optimizes the inlined chain of calls in a wrong way. We do not know whether it happens also somewhere else.

Tested in branch JS_hm_coupling, commit e6ea3cfeb8e9e7afa9e274b443b81e848b988f06.

testing code:
transport_dg.cc: ```void TransportDG<Model>::output_region_statistics()```
```
for (DHCellSide side : cell.side_range())
        {
//             Side s_side = side.side();
            
            unsigned int cond_idx;
            unsigned int elem_idx = cell.elm_idx();
            unsigned int side_idx = side.side_idx();
            Mesh* mesh1 = cell.dof_handler_->mesh();
            Element* ele = &(mesh1->element_vec_[elem_idx]);
            if (ele->boundary_idx_ == NULL) cond_idx = Mesh::undef_idx;
            else cond_idx = ele->boundary_idx_[side_idx];
            
//             LogOut() << "elem_idx: " << elem_idx << "side_idx: " << side_idx << " cond_idx: " << cond_idx;
            
            Boundary* cptr;
            if (cond_idx == Mesh::undef_idx) cptr = NULL;
            else cptr = &( mesh1->boundary_[ cond_idx ] );
            
            LogOut() << "el id: " << side.elem_idx() << " i_side: " << side.side_idx() << " cond_idx: " << side.side().cond_idx() << " cond: " << cptr;
            
        	if (side.cond() == nullptr) continue;
```

jbrezmorf:
We have temporary fix in 42d6dfc0.

jbrezmorf:
need to apply this commit to remove_mhdh

jbrezmorf:
reproduced with test 26/01
!! happens only with release build due to compiler inlining and optimizations

dflanderka:
Fix in d35b574.

PR #1123, opened 2019-08-15
Df memory leaks

dflanderka:


src/fem/dh_cell_accessor.hh
```diff
@@ -190,6 +192,36 @@ public:
...
+
+    inline ElementAccessor<3> element() const {
+    	return side().element();
+    }
+
+    inline unsigned int elem_idx() const {
+    	return side().elem_idx();
+    }
+
+    inline Boundary * cond() const {
```
jbrezmorf: Is this the problematic implementation in the release build? If it is the case, we should rather use the temporary fixed implementation until the issue #1114  is resolved.

dflanderka: Solved by cherry-pick of appropriate commit.

Issue #995, opened 2018-10-04
Remove functions create_field and fill_output_data.


dflanderka:
We need to remove temporary functions in vector_seq_double.hh and replace with directly created output fields as FieldFE with own P0 DOF handler. Current approach used on several places creates its own Vector with implicit distribution of dofs based on Mesh indexing and then FieldFE is created for output. This is done by crating an independent vector with distribution according to a P0 dof handler, then we must synchronize tha data between vectors using the fill_output_data method.

- [ ] Use appropriate iterator over cell range of DOF handler for fill in all occurences in DarcyFlow.
	- simplify FieldFE::set_fe_data: create own maps (remove these parameters), data is optional (create own vector according to dofhandler), return vector.
	- use DHCell in methods: make_element_scalar, make_element_vector, make_corner_scalar
	- DarcyFlowOutput do not access output vectors directly.
- [ ] Replace all cycles iterating over mesh range with cycles iterating over DOF handler range in reactions and transport. Use same process of filling as in DarcyFlow.

dflanderka:
Merged in pull #1124

PR #1124, opened 2019-08-16
Df remove create fe

dflanderka:
- final replace of create_field and fill_output_data functions
- solve #995


src/transport/transport.cc
```diff
@@ -937,8 +938,14 @@ void ConvectionTransport::output_data() {
...
     //}
 
+	unsigned int ndofs = dh_->max_elem_dofs();
+	unsigned int idof; // iterate over indices
+	std::vector<LongIdx> indices(ndofs);
     for (unsigned int sbi = 0; sbi < n_substances(); sbi++) {
-    	fill_output_data(out_conc[sbi], output_field_ptr[sbi]);
+        for (auto cell : dh_->own_range()) {
+            cell.get_loc_dof_indices(indices);
+            for(idof=0; idof<ndofs; idof++) output_field_ptr[sbi]->get_data_vec()[ indices[idof] ] = (*out_conc[sbi].data_ptr())[ ndofs*cell.elm_idx()+idof ];
```
jbrezmorf: There should be no such copy. We should have a field for the mobile concentration and get  the  `out_conc` vector or rather VecMPI from it. A lot of code can possibly be simplified using the VecMPI wrapper in the  ConvectionTransport consistently.

Issue #1053, opened 2019-03-15
Native IO tests


jbrezmorf:
Issues:
- After change in dof distribution and fix of get_loc_dof_indices the tests for native input fail.

- tests:  02_generic_input/06, 21 - stored input data are incompatible with current dof handler and there is not mentioned how they were created

- seems that we have no test for native-data output 

TODO:
- create tests in 03_generic_output for native_data output
- use these output data (copy output files) in 02_generic_input  native input tests, mention in comment which test to use for generating native input data

jbrezmorf:
- move part of OutputTime::prepare_compute_data into Field::compute_field_data
- Need to rethink how to set native DH for the native input fields. Currently there is no method in the abstract Field API through which one can pass the native DH. Not clear if we want to store the DH along with the mesh.

jbrezmorf:
- seems that output of the Native data works fine (need tests)
- we must fix the input of the native data. Currently we want to support only input of serial output and parallel input when number of processors much (and dof distribution match as well)
- this issue depends on resolution of #1013

dflanderka:
Done in pull #1125

PR #1125, opened 2019-08-21
Df native data

dflanderka:
- solve #1053


jbrezmorf:
DF: This PR fix:
- native data IO (and its tests)
- effective storage of values for FieldFE (elementwise input) in parralel run (store just own values)

with the following gotchas:
- breaks reading boundary FieldFE (elementwise and intrpolationP0) in connection with parallel calculation
- native output does not work for DH with system fe
- no support for parallel VTK IO

Issue #1136, opened 2019-10-01
Remove template parameter "dim" in Quadrature, Mapping, FEValues


jstebel:
For easier passing of data between classes and methods working on cells of all dimensions, especially fields. Related to #1121.
- Create mechanism of construction of arma::vec<fixed> and arma::mat<fixed> from 2d array of double.
- In Quadrature, store points as vector of doubles, and save dim as runtime parameter. Make point(), get_points() and set_point() templated by dimension. In debug mode check dimension match.
- Remove abstract class Mapping and use only MappingP1.

jstebel:
Preliminary test of inline allocation of arma::vec3 from vector<double> shows approx. 5x speed up in simple calculation of norm and almost 10x speed up in allocation in favor of vector<arma::vec3>.

jbrezmorf:
I have merged development of Peter Kral into master branch. Look at src/system/armor.hh for inspiration. 
I think we can just add a constructor to the armor::Mat class that accept pointer to external memory and remove its internal fixed space. We can provide Mat<...>::Array for allocation like:
```
auto mat_array = new Mat<...>::Array[10];
```

PR #1137, opened 2019-10-04
Js use armor

jstebel:
- Proposal of Armor::Array class providing a vector (array) of matrices without having the matrix size as template parameter.
- Application in Quadrature point coordinates. The aim is to make the class Quadrature non-templated.

Related to #1136.

### DIFF
--- a/src/system/armor.hh
+++ b/src/system/armor.hh
@@ -1,3 +1,6 @@
+#ifndef ARMOR_HH
+#define ARMOR_HH
+
 //#define ARMA_DONT_USE_WRAPPER
 //#define ARMA_NO_DEBUG
 #include <armadillo>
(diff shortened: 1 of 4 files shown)


src/system/armor.hh
```diff
@@ -130,6 +113,65 @@ public:
...
+        for (auto i : p) data.push_back( i );
+    }
+    
+    /**
+     * Return matrix at given position in array. The returned object is a Armor::Mat
+     * pointing to the respective data block in the Array's storage.
+     * @param i  Index of matrix.
+     */
+    template<uint nr, uint nc = 1>
+    inline Mat<Type,nr,nc> get(uint i) const
```
jbrezmorf: Cen we remove `const` specifier which is violated by the const_cast anyway?

jbrezmorf:
Do we need dynamic Armor::Array? There can be problems with std::vector reallocations.

PR #1139, opened 2019-10-16
JHy/build-improvements

janhybs:
1) Build directory is now within the main dir:
  - was: `../build-$GIT_BRANCH`
  - is: `build-$GIT_BRANCH`

2) Fix some issue with newer python libs
3) Add option to specify output path of a profiler via `--profiler-path foo.json`

### DIFF
--- a/src/python/scripts/yamlc/yaml_config.py
+++ b/src/python/scripts/yamlc/yaml_config.py
@@ -175,7 +175,10 @@ class ConfigBase(object):
 
     def _read_yaml(self):
         with open(self.yaml_config_file, 'r') as fp:
-            result = yaml.load(fp)
+            try:
+                result = yaml.safe_load(fp)
+            except:
+                result = yaml.load(fp)
         return result or dict()
 
     def _get_all_yamls(self):
--- a/src/system/sys_profiler.cc
+++ b/src/system/sys_profiler.cc
@@ -703,18 +703,7 @@ void Profiler::output(MPI_Comm comm, ostream &os) {
 }
 
 
-void Profiler::output(MPI_Comm comm) {
-    int mpi_rank;
-    chkerr( MPI_Comm_rank(MPI_COMM_WORLD, &mpi_rank) );
-    if (mpi_rank == 0) {
-        output(comm, *get_default_output_stream());
-    } else {
-        ostringstream os;
-        output(comm, os );
-    }
-}
-
-void Profiler::output(MPI_Comm comm, string profiler_path) {
+void Profiler::output(MPI_Comm comm, string profiler_path /* = "" */) {
     if(profiler_path == "") {
         output(comm);
     }else {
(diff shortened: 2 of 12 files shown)


src/system/sys_profiler.cc
```diff
@@ -714,6 +714,23 @@ void Profiler::output(MPI_Comm comm) {
     }
 }
 
+void Profiler::output(MPI_Comm comm, string profiler_path) {
```
jbrezmorf: Add the `profiler_path` parameter as optional to the existing methods.

src/python/scripts/yamlc/yaml_config.py
```diff
@@ -175,7 +175,10 @@ def get_one(self, yaml_case_file):
 
     def _read_yaml(self):
         with open(self.yaml_config_file, 'r') as fp:
-            result = yaml.load(fp)
+            try:
+                result = yaml.safe_load(fp)
```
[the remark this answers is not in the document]
janhybs: Simple `load` will be deprecated in the future due to security concerns.

src/system/sys_profiler.cc
janhybs: done

src/python/scripts/yamlc/yaml_config.py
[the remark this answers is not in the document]
janhybs: Simple `load` will be deprecated in the future due to security concerns.

janhybs:
Since `CMakeLists.txt` is related to the build process, this branch now includes default `config.default.cmake` which will not require any `config.cmake` when running inside docker `flow-libs-dev-rel` or `flow-libs-dev-rel`

src/python/scripts/yamlc/yaml_config.py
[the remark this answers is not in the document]
jbrezmorf: I know, so why not just use the safe variant.

src/system/sys_profiler.cc
```diff
@@ -703,14 +701,20 @@ void Profiler::output(MPI_Comm comm, ostream &os) {
...
-    int mpi_rank;
-    chkerr( MPI_Comm_rank(MPI_COMM_WORLD, &mpi_rank) );
-    if (mpi_rank == 0) {
-        output(comm, *get_default_output_stream());
-    } else {
-        ostringstream os;
-        output(comm, os );
+void Profiler::output(MPI_Comm comm, string profiler_path /* = "" */) {
+    if(profiler_path == "") {
+        output(comm);
```
jbrezmorf: `if (mpi_rank == 0)` should come first.
then use `*get_default_output_stream()` in the case of empty path.

jbrezmorf:
> Since `CMakeLists.txt` is related to the build process, this branch now includes default `config.default.cmake` which will not require any `config.cmake` when running inside docker `flow-libs-dev-rel` or `flow-libs-dev-dbg`

Any problem with that?

src/system/sys_profiler.cc
janhybs: Done

PR #1142, opened 2019-10-25
Js use armor

jstebel:
Remove template parameter "dim" from Quadrature.

### DIFF
--- a/src/quadrature/quadrature.cc
+++ b/src/quadrature/quadrature.cc
@@ -1,0 +1,128 @@
+/*!
+ *
+﻿ * Copyright (C) 2015 Technical University of Liberec.  All rights reserved.
+ * 
+ * This program is free software; you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License version 3 as published by the
+ * Free Software Foundation. (http://www.gnu.org/licenses/gpl-3.0.en.html)
+ * 
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+ *
+ * 
+ * @file    quadrature.hh
+ * @brief   Basic definitions of numerical quadrature rules.
+ * @author  Jan Stebel
+ */
+
+#include "quadrature/quadrature.hh"
+
+
+
+Quadrature& Quadrature::operator=(const Quadrature &q)
+{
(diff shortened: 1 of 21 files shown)


src/quadrature/quadrature.hh
```diff
@@ -45,60 +45,90 @@
...
 
     /// Returns number of quadrature points.
-    const unsigned int size() const;
+    inline const unsigned int size() const
+    { return weights.size(); }
 
     /// Returns the <tt>i</tt>th quadrature point.
-    arma::vec::fixed<dim> point(const unsigned int i) const;
+    template<unsigned int point_dim>
+    inline arma::vec::fixed<point_dim> point(const unsigned int i) const
```
jbrezmorf: We can simplify the interface by returning directly the Mat object returned by get<point_dim>(i).
This way we can assign to the point and we can drop the set_point method. Same for the weights.

src/quadrature/quadrature.cc
```diff
@@ -0,0 +1,128 @@
...
+{}
+
+
+
+
+
+
+
+template<unsigned int quad_dim>
+Quadrature quadrature_from_side(const Quadrature &subq, unsigned int sid, unsigned int pid)
```
jbrezmorf: I suppose this is a replacement for the constructor? However as the diff takes it as a completely new code it is hard to see the changes. Better discuss it personally.

src/system/armor.hh
```diff
@@ -162,7 +162,7 @@ public:
     inline Mat<Type,nr,nc> get(uint i) const
     {
         ASSERT_DBG( (nr == nRows) && (nc == nCols) );
-        return Mat<Type,nr,nc>( const_cast<Type*>(data.data()) + i*nRows*nCols );
+        return Mat<Type,nr,nc>( (Type *)data.data() + i*nRows*nCols );
```
jbrezmorf: We will probably need both const and non-const get method to allow assignenment.

jbrezmorf:
Great. Just few points to discuss, possibly on Tuesday.

src/quadrature/quadrature.cc
jstebel: Yes, this method was originally constructor. I was not able to declare a templated constructor, hence moved it to a templated function.

src/quadrature/quadrature.hh
jstebel: Fixed.

src/system/armor.hh
jstebel: This method already allows assignment. Do we need some other version?

PR #1145, opened 2019-11-04
Df formula field params

dflanderka:


dflanderka:
First commit edc60dc cherry-picked to DF_use_bparser, next to commits will be used (maybe) later.

PR #1149, opened 2019-11-06
Df formula parser

dflanderka:
- test possibly new parsers of FieldFormula: muparser and exprtk
- exprtk is faster
- add this parser to code and create speed unit test


dflanderka:
Test of different parsers for FieldFormula. Exprtk is best but doesn't allow parsing of vector and tensor fields. We will use own Bparser.

dflanderka:
Blind branch, tests of parsers are solved in parser_comparison repository.

PR #1159, opened 2019-11-15
Js reduce mapping

jstebel:
Resolve part of #1136 .

Remove abstract class Mapping.
Make from MappingP1 just an envelope of static methods (templated by dim and spacedim).
Move fill* methods to FEValues and FESideValues.


src/fields/fe_value_handler.cc
```diff
@@ -124,19 +120,18 @@ template <int elemdim, int spacedim, class Value>
...
 {
-    ASSERT_PTR(map_).error();
     ASSERT_EQ( point_list.size(), value_list.size() ).error();
 
     const DHCellAccessor cell = dh_->cell_accessor_from_element( elm.idx() );
     if (boundary_dofs_) this->get_dof_indices( elm, dof_indices);
     else cell.get_loc_dof_indices( dof_indices );
 
-    arma::mat map_mat = map_->element_map(elm);
+    arma::mat map_mat = MappingP1<elemdim,spacedim>::element_map(elm);
```
jstebel: I am not sure if the second template argument should be spacedim or 3 (as in the original object map_, see l. 106).

jstebel:
Updated structure of FEValues:
- Data and functionality not related to FE moved to new classes ElementValues, ElemSideValues
- Removed class FEValuesData, moved FEInternalData inside FEValuesBase.

src/fields/fe_value_handler.hh
```diff
@@ -20,12 +20,12 @@
...
 #include "system/index_types.hh"
 #include "fields/field_values.hh"
-#include "fem/mapping_p1.hh"
 #include "fem/finite_element.hh"
 #include "mesh/point.hh"
+#include "la/vector_mpi.hh"
 #include <armadillo>
 
-class VectorMPI;
+class DOFHandlerMultiDim;
```
jbrezmorf: Why these changes in the includes and forward declaration?

jbrezmorf:
What is the reason to keep ElementValues and ElementSideValues derived from a common base class? Can we have them completely separated? Or the other way, merged into single class with separate methods for bulk and side data? This second approach would be possibly better for further usage in specialized fields.

src/fields/fe_value_handler.hh
jstebel: VectorMPI is no more necessary. DOFHandlerMultiDim was declared via fem/mapping_p1.hh, which is now not included.

PR #1162, opened 2019-11-25
Df js use armor

jbrezmorf:


dflanderka:
Armor code is reimplemented, Branch can be closed.

PR #1175, opened 2019-12-18
Js fe values without dim

### LINKED ISSUE #1136: Remove template parameter "dim" in Quadrature, Mapping, FEValues

jstebel:
Simplified ElementValues. Partially resolves #1136 
- removed template parameter "dim"
- unified class for sides and element


jbrezmorf:
General topics for discussion:
Interface points:
Implementation points:
- The assembly process will be performed over "patches of elements". For each patch:
   1. collect quadrature points on its elements
   2. precompute fevalues and field values in quadrature points
   3. call actual assembly functions
  Therefore we need to store values in quadrature points in a consistent way, 
  that is values in all points on the patch in the single vector

(To be continued.)

PR #1178, opened 2020-01-23
Pe balance

Paulie14:
- change interface in balance to use local dofs
- update transport.cc assemble routines with DHCellAccessor
- change type Idx to int
- improve a bit code structure of BC assembly in LMH

### DIFF
--- a/src/coupling/balance.hh
+++ b/src/coupling/balance.hh
@@ -36,6 +36,7 @@ class Mesh;
 class TimeGovernor;
 class DOFHandlerMultiDim;
 class DHCellSide;
+class DHCellAccessor;
 namespace Input {
 	namespace Type {
 		class Record;
@@ -262,6 +263,7 @@ public:
 	}
 
 	/**
+	 * DEPRECATED, use add_mass_values instead.
 	 * Adds elements into matrix for computing mass.
 	 * @param quantity_idx  Index of quantity.
 	 * @param region_idx    Index of bulk region.
@@ -298,6 +300,20 @@ public:
             unsigned int region_idx,
             double value);
 	
+	/**
+	 * Adds elements into matrix for computing (outgoing) flux.
+	 * @param quantity_idx  Index of quantity.
+	 * @param side          DHCellSide iterator.
+	 * @param loc_dof_indices   Local dof indices (to the solution vector) to be added.
+	 * @param mat_values    Values to be added into matrix.
+     * @param vec_value     Value to be added into vector.
+	 */
+	void add_mass_values(unsigned int quantity_idx,
+			const DHCellAccessor &dh_cell,
+			const vector<Idx> &loc_dof_indices,
+			const std::vector<double> &mat_values,
+			const double &vec_value);
+
 	/**
 	 * Adds elements into matrix for computing (outgoing) flux.
 	 * @param quantity_idx  Index of quantity.
--- a/src/coupling/balance.hh
+++ b/src/coupling/balance.hh
@@ -264,46 +264,57 @@ public:
 
 	
 	/**
-	 * Adds elements into matrix for computing (outgoing) flux.
+	 * Adds elements into matrix for computing mass.
+	 * The mass matrix M is in format [n_dofs x n_bulk_regions] for each quantity.
+	 * The mass vector mv is in format [n_bulk_regions] for each quantity.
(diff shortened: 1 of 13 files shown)


src/coupling/balance.hh
```diff
@@ -260,42 +262,34 @@ public:
 			start_source_assembly(idx);
 	}
 
+	
 	/**
-	 * Adds elements into matrix for computing mass.
+	 * Adds elements into matrix for computing (outgoing) flux.
```
jbrezmorf: for computing mass.
... document shape of matrix and vector for every kind of the add_* routine.

src/coupling/balance.hh
```diff
@@ -260,42 +262,34 @@ public:
...
 	 */
-	void add_mass_matrix_values(unsigned int quantity_idx,
-			unsigned int region_idx,
-			const std::vector<LongIdx> &dof_indices,
-			const std::vector<double> &values);
+	void add_mass_values(unsigned int quantity_idx,
+			const DHCellAccessor &dh_cell,
+			const LocDofVec &loc_dof_indices,
+			const std::vector<double> &mat_values,
+			const double &vec_value);
```
jbrezmorf: vec_value could be rather passed by value.

src/coupling/balance.hh
```diff
@@ -260,42 +262,34 @@ public:
...
-     * @param value         Value to be added.
-     */
-    void add_mass_vec_value(unsigned int quantity_idx,
-            unsigned int region_idx,
-            double value);
+	void add_flux_values(unsigned int quantity_idx,
+			const DHCellSide &side,
+			const LocDofVec &loc_dof_indices,
+			const std::vector<double> &mat_values,
+			const double &vec_value);
```
jbrezmorf: vec_value could be rather passed by value.

src/coupling/balance.hh
```diff
@@ -307,20 +301,10 @@ public:
 	 */
 	void add_source_values(unsigned int quantity_idx,
 			unsigned int region_idx,
-			const std::vector<LongIdx> &loc_dof_indices,
+			const LocDofVec &loc_dof_indices,
 			const std::vector<double> &mat_values,
             const std::vector<double> &vec_values);
```
jbrezmorf: why we have vector of vec_values here?

src/flow/assembly_lmh.hh
```diff
@@ -235,139 +232,144 @@ protected:
...
+            // side-edge (flux-lambda) terms
+            loc_system_.add_value(loc_side_dofs[sidx], loc_edge_dofs[sidx], 1.0);
+            loc_system_.add_value(loc_edge_dofs[sidx], loc_side_dofs[sidx], 1.0);
+        }
+    }
+
+
+    void assemble_side_bc(const DHCellSide& side, double cross_section, bool use_dirichlet_switch)
+    {
+        const unsigned int sidx = side.side_idx();
```
jbrezmorf: I hope this doesn't change as the diff is a mess.

src/flow/assembly_lmh.hh
```diff
@@ -235,139 +232,144 @@ protected:
...
+
+            double bc_pressure = ad_->bc_pressure.value(b_ele.centre(), b_ele);
+            double bc_switch_pressure = ad_->bc_switch_pressure.value(b_ele.centre(), b_ele);
+            double bc_flux = -ad_->bc_flux.value(b_ele.centre(), b_ele);
+            double bc_sigma = ad_->bc_robin_sigma.value(b_ele.centre(), b_ele);
+
+            double solution_head = ad_->p_edge_solution[loc_schur_.row_dofs[sidx]];
+
+            // Force Robin type during the first iteration of the unsteady case.
+            if (solution_head > bc_switch_pressure  || ad_->force_bc_switch) {
```
jbrezmorf: 'force_bc_switch' rename to 'force_no_neuman_bc' ?

src/flow/assembly_lmh.hh
Paulie14: I did this with intention not to do any changes in the code logic, but to separate the switch part of the BC code and to reduce the tabs of all the "embeddings".

src/flow/assembly_lmh.hh
Paulie14: Agreed.

PR #1203, opened 2020-04-08
Df evaluate fe

dflanderka:


src/fields/field_fe.cc
```diff
@@ -223,6 +284,59 @@ void FieldFE<spacedim, Value>::value_list (const Armor::array &point_list, const
...
+        this->init_fe_value<1>(eval_points);
+        this->init_fe_value<2>(eval_points);
+        this->init_fe_value<3>(eval_points);
+    }
+
+    auto update_cache_data = cache_map.update_cache_data();
+    std::unordered_map<unsigned int, typename ElementCacheMap::RegionData>::iterator reg_elm_it =
+            update_cache_data.region_cache_indices_map_.find(region_idx);
+
+    for (unsigned int i_elm=0; i_elm<reg_elm_it->second.element_set_.size(); ++i_elm) {
```
jbrezmorf, line 306: Over-complicated ElementCacheMap.
Why there is ElementSet in the RegionData? I think there is no need for RegionData at all. As cached elements of the single region form a continuous sequence (range) then 
`elm_idx_`  + `UpdateCacheHelper::region_cache_indices_range_`

contains whole information.

Would be better to pass the region_chunk index (`UpdateCacheHelper::region_cache_indices_range_`)
 instead of the region_idx in order to avoid unecessary `find(region_idx)`. 
Then this loop should probably run over the region chunk using `region_cache_indices_range_`, similarly to FieldModel<..>::cache_update

src/fields/field_fe.cc
```diff
@@ -223,6 +284,59 @@ void FieldFE<spacedim, Value>::value_list (const Armor::array &point_list, const
...
+        DHCellAccessor cell = dh_->cell_accessor_from_element( elm_idx );
+        LocDofVec loc_dofs = cell.get_loc_dof_indices();
+
+        for (unsigned int i_ep=0; i_ep<eval_points->max_size(); ++i_ep) { // i_eval_point
+            //DHCellAccessor cache_cell = cache_map(cell);
+            int field_cache_idx = cache_map.get_field_value_cache_index(cache_map(cell).element_cache_index(), i_ep);
+            if (field_cache_idx < 0) continue; // skip
+            mat_value.fill(0.0);
+    		for (unsigned int i_dof=0; i_dof<loc_dofs.n_elem; i_dof++) {
+    		    mat_value += data_vec_[loc_dofs[i_dof]]
```
jbrezmorf, line 320: This step is basically matrix-vector multiplication. with FeValues matrix of shape n_values * n_dofs, multiplied by
the dof value vector. See implementation of FeValues::xyz_view and try use here shape_value_component directly.
That way we can avoid EvalShapeHandler and open space for more optimization later on.

jbrezmorf: consult with @jstebel  transpose in tensor case

src/fields/field_fe.cc
jbrezmorf: Mark as an auxiliary internal structure. We can probably replace it by a map:
region_index -> region_cache_range_index. Postpone further optimization after the FeildFE will be actually used.

jbrezmorf:
Try to profile the 22/04 test using oprifile:

1. run with `operf`, [example](https://www.nsnam.org/wiki/HOWTO_use_oprofile)
2. generate the report using `opreport`, [examples](https://oprofile.sourceforge.io/examples/)

src/fields/field_fe.cc
jbrezmorf: 1. @jstebel Check how tensor elements are indexed (column-wise/row-wise)
2. @dflanderka Check how we treat GMSH tensors in IO operations.

jstebel: GMSH treats tensors in "row-wise" order. I.e. values "1 2 3 4 5 6 7 8 9" correspond to the tensor:
1 2 3
4 5 6
7 8 9

Simple example: In the following MSH file, plugin Eigenvectors produces vectors in X and Z axis, which means that the values " 1 1 0 0 1 0 0 0 1" in the last but one line represent the tensor:
1 1 0
0 1 0
0 0 1

$MeshFormat
2.2 0 8
$EndMeshFormat
$PhysicalNames
1
3       100     "tetra"
$EndPhysicalNames
$Nodes
4
1 0 0 0
2 1 0 0
3 0 1 0
4 0 0 1
$EndNodes
$Elements
1
1 4 2 100 100 1 2 3 4
$EndElements

$ElementData
1
"tensor"
1
0.0
3
0
9
1
1 1 1 0 0 1 0 0 0 1
$EndElementData

dflanderka: Data object in FieldFE (VectorMPI) is in "row-wise" order too. That is for tensor:
1 2 3
4 5 6
7 8 9
Values are stored in order "1 2 3 4 5 6 7 8 9".

Only difference is in armadillo "inner" data represantation  that is stored in "col-wise". If we get mat.memptr() and iterate through items, result is "1 4 7 2 5 8 3 6 9". But all other (print to output, item on position i,j) result correct values of original tensor.

Both - code in FieldFE and unit test (field_evaluate_fe_test) - is correct. There si two ways how to initialize matrix with initializer list:
arma::mat33 A = {1, 4, 7, 2, 5, 8, 3, 6, 9};
arma::mat33 B = { {1, 2, 3}, {4, 5, 6}, {7, 8, 9} }; // A == B
Unit test uses first initialization and therefore it looks like an error.

dflanderka: Proposal of replace of EvalShapeHandler is in 584d1a6.

PR #1211, opened 2020-04-23
Template record in EquationBase with key 'time'.

jstebel:
Equations which use time key now inherit their input record from EquationBase and thus use the same settings for time governor.
Now time key is obligatory with empty default value.

Question: How about HC_ExplicitSequential? It has key "time" but is not descendant of EquationBase. Should its input record also copy keys from EquationBase::record_template()?

### DIFF
--- a/src/coupling/equation.cc
+++ b/src/coupling/equation.cc
@@ -33,6 +33,13 @@
  * Implementation of EqBase
  */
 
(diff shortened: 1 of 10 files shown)


jstebel:
Inheritance of HC_ExplicitSequential resolved.

Issue #779, opened 2017-01-17
Class for iterative coupling


jstebel:
Similarly as CouplingSequentialBase, create a base class CouplingIterativeBase to allow iterations between processes within one computational time step. Required for poroelasticity.

jstebel:
Attributes:
- equations (flow, mechanics, ...)
- input record template

Virtual methods:
- after solve step of one equation, update data in other equations
- compute residuum
- after converged, update data from "old" time

jstebel:
Iterative HM coupling acts as a preconditioned Richardson method:
x_{n+1} = x_n + P^{-1}(b - A*x_n), 
where x_n = [ u_n, p_n ]^T and A is block-wise matrix. Hence to apply it we need to compute the residual and solve the preconditioned system. Within PETSc we could use the MATNEST matrix type.

jstebel:
Partially implemented in the class IterativeCoupling. Interface to PETSC SNES solvers should be moved to next release.

PR #1217, opened 2020-05-05
Js coupling

jstebel:
Proposal of general class for iterative couplings. Related to #779 .

There is also a bug fix resolving wrong update of fluid source in HM iterations.


jbrezmorf:
Merge this to master. It works and makes some valuable improvements. After we can continue to find a unified approach for hierarchical coupling.

PR #1224, opened 2020-05-14
operator [] for Mixed* templates

jbrezmorf:
- fix build under clang (still many warnings)
- rename Idx to IntIdx to diminish possible conflicts

### DIFF
--- a/src/fem/dh_cell_accessor.hh
+++ b/src/fem/dh_cell_accessor.hh
@@ -113,7 +113,7 @@ public:
     template<unsigned int dim>
     FEPtr<dim> fe() const {
         ElementAccessor<3> elm_acc = this->elm();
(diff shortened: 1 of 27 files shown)


jbrezmorf:
@dflanderka , @Paulie14, @jstebel 

This commit introduces access of the components of Mixed and MixedPtr tuples using the operator[]. This replaces problematic 'get' method where explicit specification of the dimesion requires usage of the 'template' keyword which leads to ugly code on lot of places.

The operator[] takes Dim<dim> as the index instead of an int. This can be used if we use inside of other template. Further the custom literal is defined so one can use indeces: 0_d, 1_d ,...

See the commit to figure out how it is used.

Next I plan to implement kind of 'for'.

PR #1230, opened 2020-05-25
Pe pass vel

Paulie14:
- pass flux from Darcy flow by data()
- clean up in DarcyFlowInterface
- simplify creation of FieldFE where applicable

question:
In transport, we have an auxiliary flag which is set when the flux is changed. I was thinking about using the property of the field `flow_flux->changed()` but that did not work for me (did some tests on 27/05).
Is this the way or not?

### DIFF
--- a/src/flow/darcy_flow_lmh.cc
+++ b/src/flow/darcy_flow_lmh.cc
@@ -289,12 +289,12 @@ void DarcyLMH::initialize() {
 
     { // construct pressure, velocity and piezo head fields
 		uint rt_component = 0;
-        ele_flux_ptr = create_field_fe<3, FieldValue<3>::VectorFixed>(data_->dh_, rt_component);
+        auto ele_flux_ptr = create_field_fe<3, FieldValue<3>::VectorFixed>(data_->dh_, rt_component);
+        data_->full_solution = ele_flux_ptr->get_data_vec();
+        data_->flux.set_field(mesh_->region_db().get_region_set("ALL"), ele_flux_ptr);
+
 		auto ele_velocity_ptr = std::make_shared< FieldDivide<3, FieldValue<3>::VectorFixed> >(ele_flux_ptr, data_->cross_section);
 		data_->field_ele_velocity.set_field(mesh_->region_db().get_region_set("ALL"), ele_velocity_ptr);
-		data_->full_solution = ele_flux_ptr->get_data_vec();
-
-        data_->flux.set_field(mesh_->region_db().get_region_set("ALL"), ele_flux_ptr);
 
 		uint p_ele_component = 0;
         auto ele_pressure_ptr = create_field_fe<3, FieldValue<3>::Scalar>(data_->dh_, p_ele_component, &data_->full_solution);
@@ -487,8 +487,8 @@ void DarcyLMH::update_solution()
 
     solve_time_step();
     
-    ele_flux_ptr->local_to_ghost_data_scatter_begin();
-    ele_flux_ptr->local_to_ghost_data_scatter_end();
+    data_->full_solution.local_to_ghost_begin();
+    data_->full_solution.local_to_ghost_end();
 }
 
 void DarcyLMH::solve_time_step(bool output)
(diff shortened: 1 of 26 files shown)


jstebel:
> In transport, we have an auxiliary flag which is set when the flux is changed. I was thinking about using the property of the field `flow_flux->changed()` but that did not work for me (did some tests on 27/05).
> Is this the way or not?

It should be possible similarly as in HM_Iterative::update_potential(), where we manually set the changed flag of certain field by calling set_time_result_changed().

Paulie14:
> > In transport, we have an auxiliary flag which is set when the flux is changed. I was thinking about using the property of the field `flow_flux->changed()` but that did not work for me (did some tests on 27/05).
> > Is this the way or not?
> 
> It should be possible similarly as in HM_Iterative::update_potential(), where we manually set the changed flag of certain field by calling set_time_result_changed().

Works well in DG, but I am not sure about `transport.cc`. I had to read the `changed` state of the flux field before setting time to all the fields with `LimitSide:right`. Honestly, I am not sure why it works...

src/flow/darcy_flow_lmh.cc
```diff
@@ -484,6 +486,9 @@ void DarcyLMH::update_solution()
     time_->view("DARCY"); //time governor information output
 
     solve_time_step();
+    
+    data_->full_solution.local_to_ghost_begin();
```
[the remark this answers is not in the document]
jbrezmorf: Same in Darcy MH.

src/flow/darcy_flow_mh.cc
```diff
@@ -950,7 +960,7 @@ void DarcyMH::create_linear_system(Input::AbstractRecord in_rec) {
             LinSys_BDDC *ls = new LinSys_BDDC(&(*data_->dh_->distr()),
                     true); // swap signs of matrix and rhs to make the matrix SPD
             ls->set_from_input(in_rec);
-            ls->set_solution( ele_flux_ptr->get_data_vec().petsc_vec() );
+            ls->set_solution( data_->full_solution.petsc_vec() );
```
jbrezmorf, line 963: Rather keep previous code.

src/flow/richards_lmh.cc
```diff
@@ -148,14 +148,10 @@ void RichardsLMH::initialize_specific() {
     data_->mesh = mesh_;
     data_->set_mesh(*mesh_);
 
-    data_->water_content_ptr = std::make_shared< FieldFE<3, FieldValue<3>::Scalar> >();
-    data_->water_content_ptr->set_fe_data(data_->dh_cr_disc_, 0);
-    data_->water_content.set_mesh(*mesh_);
+    data_->water_content_ptr = create_field_fe< 3, FieldValue<3>::Scalar >(data_->dh_cr_disc_);
```
jbrezmorf, line 151: Can we omit set_mesh?

src/transport/assembly_dg.hh
```diff
@@ -503,7 +503,7 @@ public:
                                 const Armor::array &point_list)
         {
             velocity.resize(point_list.size());
-            model_->velocity_field_ptr()->value_list(point_list, cell, velocity);
+            model_->data().flow_flux->value_list(point_list, cell, velocity);
```
jbrezmorf, line 506: In order to perform scatter we can use:

`model_->data().flow_flux->get_field_fe()->local_to_ghost_data_scatter_begin()`

src/transport/transport_dg.cc
```diff
@@ -505,7 +505,7 @@ void TransportDG<Model>::update_solution()
     // assemble stiffness matrix
     if (stiffness_matrix[0] == NULL
             || data_->subset(FieldFlag::in_main_matrix).changed()
-            || Model::flux_changed)
+            || data_->flow_flux.changed())
```
jbrezmorf, line 508: These various changed flags are problematic since they have to be switched on and off. We should probably introduce the state counters. There is a stub `system/lazy_dependency.py` for that.

Issue #1236, opened 2020-06-08
Adding ElementDataCache bug (DummyOutputData).


Paulie14:
Blocker for @ales-balvin-tul @milanhokr.

The following combination:
- DG model `transport: !Solute_AdvectionDiffusion_DG`
- with `reaction_term`
- with output (sorption/dual por.) to `gmsh` format

is never tested and fails with `ERRORstd::bad_cast`
in `output_time.impl.hh: prepare_compute_data()` in the cast of `DummyOutputData` in
```return dynamic_cast<ElementDataCache<T> &>(*(*it));```

E.g. in test 27_solute_dg_time/02_dg_dp_sorp_small.yaml - change solute output to `gmsh`
and output any field of `SorptionMobile`

Further survey:
- class `DummyOutputData` does not inherit from `ElementDataCache<T>` but only from `ElementDataCacheBase` so this cast would always fail for `DummyOutputData` (it has not happened until now)
- it seems that class `DummyOutputData` is not tested anywhere
- the row ```return dynamic_cast<ElementDataCache<T> &>(*(*it));``` appeared along time ago,
without concerning `DummyOutputData`
- we do not test `DummyOutputData` in any test for its original purpose (no output time for a particular field in gmsh format)
- not sure why this is happening only in combination with DG

Paulie14:
Fixed by e3e620c1a6627b23c5aac8cc5583360b7f2dbff5.

PR #1238, opened 2020-06-11
Pe dummycache bug

Paulie14:
- fix adding output fields data caches into final vector
- simplify usage of DummyElementDataCache


Paulie14:
fixes #1236

Issue #1234, opened 2020-06-08
boundary FieldFE


jbrezmorf:
**Problem 1**

We can not have boundary FieldFE since we need boundary DofHandler and boundary Mesh.
This is sort of optimization as we should be able to evaluate the bulk fields on the boundary as well.
Current concept of FieldBC only means that this kind of optimization can be applied.

**Problem 2**
Support both bulk and boundary FieldFE for boundary fields. Or have some adaptor of the bulk fields.

**Problem 3**
Special fields for boundary elements: X, Y, Z, D, normal

jbrezmorf:
@jstebel Make the `MeshBase` class with members and methods used by DofHandler, 
mark separately those used by the `duplicate_nodes`.

jbrezmorf:
TODO: 
- [ ] Finish common mesh interface, PR #1246. 
- [ ] Make BC Mesh.
- [ ] Make an issue for FieldFE on the BC Mesh (have to test DofHandler etc.)

jbrezmorf:
TODO @jbrezmorf :
How to work with boundary elements in the patch for two source meshes (bulk and boundary)?
Are explicit boundary elements necessary?

jstebel:
After merging master to JS_mesh_interface, the code fails on all tests excluding flow problems. After some attempts I am still not able to resolve the problem. It fails on segmentation fault, probably due to access outside of some array. Snippet of log message:

          ##                   region_idx 3: A_init_conc, 
          ##                   region_idx 4: A_init_conc, 
          ##                   region_idx 5: A_init_conc, 
          ## 00:00:04.263      DEBUG.(/src/coupling/generic_assembly.hh:429)Order of evaluated fields (InitConditionAssemblyDG):
          ##                   region_idx 0: A_init_conc, 
          ##                   region_idx 1: A_init_conc, 
          ##                   region_idx 2: A_init_conc, 
          ##                   region_idx 3: A_init_conc, 
          ##                   region_idx 4: A_init_conc, 
          ##                   region_idx 5: A_init_conc, 
          ## 00:00:04.831      DEBUG.(/src/coupling/generic_assembly.hh:429)Order of evaluated fields (StiffnessAssemblyDG):
          ##                   region_idx 0: cross_section, A_bc_type, A_bc_robin_sigma, sources_sigma_out, advection_coef, diffusion_coef, A_fracture_sigma, A_dg_penalty, 
[...]

PR #1246, opened 2020-06-17
Js mesh interface

jstebel:
MeshBase class (see #1234).


src/mesh/accessors_impl.hh
```diff
@@ -225,7 +224,7 @@ inline bool Side::is_boundary() const {
 }
 
 inline NodeAccessor<3> Side::node(unsigned int i) const {
```
jbrezmorf, line 226: Do we need `side_nodes` data structure and global numbering of sides? That was necessary for MH, but it is now replaced by universal dofhandler. We can possibly use element_idx + local side number in the Side accessor, then
the element accessor can be used to obtain nodes.

src/mesh/accessors_impl.hh
```diff
@@ -274,16 +273,16 @@ inline Edge Boundary::edge()
 inline ElementAccessor<3> Boundary::element_accessor()
 {
     ASSERT_DBG(is_valid());
-    return boundary_data_->mesh_->element_accessor(boundary_data_->bc_ele_idx_);
+    return boundary_data_->mesh_->get_bc_mesh()->element_accessor(boundary_data_->bc_ele_idx_);
```
jbrezmorf, line 276: Please, make a separate issue to remove Bounady and BoundaryData classes, replacing them by the accessors.

jbrezmorf: Could be the next step to make Mesh simpler and more consistent.

src/mesh/accessors_impl.hh
```diff
@@ -274,16 +273,16 @@ inline Edge Boundary::edge()
...
 { 
     return element_accessor().region();
 }
 
-inline Element * Boundary::element()
+inline const Element * Boundary::element()
 {
     ASSERT_DBG(is_valid());
-    return &( boundary_data_->mesh_->element_vec_[boundary_data_->bc_ele_idx_] );
+    return &( boundary_data_->mesh_->get_bc_mesh()->element(boundary_data_->bc_ele_idx_) );
```
jbrezmorf, line 287: Please, rename `geT_bc_mesh` to `bc_mesh`. Same for other getters, reduce boilerplate.

src/mesh/bc_mesh.cc
```diff
@@ -31,7 +31,7 @@ BCMesh::BCMesh(Mesh* parent_mesh)
 : parent_mesh_(parent_mesh),
   local_part_(nullptr)
 {
-	this->init_element_vector(0);
+	this->init_element_vector(1);
```
jbrezmorf, line 34: Whot is the reason to reserve 1 element?

src/mesh/bc_mesh.cc
```diff
@@ -86,7 +81,58 @@ unsigned int BCMesh::n_nodes() const {
 
 
 ElementAccessor<3> BCMesh::element_accessor(unsigned int idx) const {
```
jbrezmorf, line 83: Rather duplicate some data structures in BCMesh instance and move these intensively used getters and accessors to the MeshBase.

src/mesh/bc_mesh.cc
```diff
@@ -86,7 +81,58 @@ unsigned int BCMesh::n_nodes() const {
...
+	return Edge();
+}
+
+Boundary BCMesh::boundary(unsigned int) const
+{
+	ASSERT( false );
+	return Boundary();
 }
 
+const Neighbour &BCMesh::vb_neighbour(unsigned int) const
```
jbrezmorf, line 106: Should return a temporary accessor.

src/mesh/mesh.cc
```diff
@@ -648,14 +638,14 @@ void Mesh::make_neighbours_and_edges()
                     elm.boundary_idx_[s] = bdr_idx;
 
                     // fill boundary element
-                    Element * bc_ele = add_element_to_vector(-bdr_idx);
+                    Element * bc_ele = add_element_to_vector(-bdr_idx, true);
```
jbrezmorf, line 641: The implicit boundary elements should be added to the BCMesh.

src/mesh/mesh.cc
```diff
@@ -791,14 +781,14 @@ void Mesh::element_to_neigh_vb()
 
 	//MessageOut() << "Element to neighbours of vb2 type... "/*orig verb 5*/;
 
-	for (vector<Element>::iterator ele = element_vec_.begin(); ele!= element_vec_.begin()+bulk_size_; ++ele)
+	for (vector<Element>::iterator ele = element_vec_.begin(); ele!= element_vec_.end(); ++ele)
```
jbrezmorf, line 784: We should rather use an element range.

src/mesh/mesh.cc
```diff
@@ -791,14 +781,14 @@ void Mesh::element_to_neigh_vb()
...
 		ele->n_neighs_vb_ =0;
 
     // count vb neighs per element
     for (auto & ngh : this->vb_neighbours_)  ngh.element()->n_neighs_vb_++;
 
     // Allocation of the array per element
-    for (vector<Element>::iterator ele = element_vec_.begin(); ele!= element_vec_.begin()+bulk_size_; ++ele)
+    for (vector<Element>::iterator ele = element_vec_.begin(); ele!= element_vec_.end(); ++ele)
```
jbrezmorf, line 791: Same here.

src/mesh/mesh.cc
```diff
@@ -1074,17 +1064,11 @@ void Mesh::add_element(unsigned int elm_id, unsigned int dim, unsigned int regio
...
-	} else {
-		if(dim == 0 ) {
-			WarningOut().fmt("Bulk elements of zero size(dim=0) are not supported. Element ID: {}.\n", elm_id);
-		}
-		else {
-			Element *ele = add_element_to_vector(elm_id);
-			bulk_size_++;
-			this->init_element(ele, elm_id, dim, region_idx, partition_id, node_ids);
-		}
+	if (!region_idx.is_boundary() && dim == 0) {
```
jbrezmorf, line 1067: Do we still need separate `add_element_to_vector` and `add_element`? 
The reason was that `add_element` has created a temporary vector of boundary elements.

src/mesh/mesh.cc
```diff
@@ -1118,35 +1102,36 @@ vector<vector<unsigned int> > const & Mesh::node_elements() {
...
 	node_ids_.clear();
 	node_ids_.reserve(size);
 }
 
 
-Element * Mesh::add_element_to_vector(int id) {
-    element_vec_.push_back( Element() );
-    Element * elem = &element_vec_.back(); //[element_vec_.size()-1];
-    element_ids_.add_item((unsigned int)(id));
+Element * MeshBase::add_element_to_vector(int id, bool is_boundary) {
```
jbrezmorf, line 1120: The distinction should rather be done in GMSH reader and one should rather use mesh->add_element_to_vector or bc_mesh->add_element_to_vector.

src/mesh/mesh.h
```diff
@@ -71,11 +71,118 @@ class BoundarySegment {
...
+    {}
+
+    virtual ~MeshBase() {};
+
+    virtual unsigned int n_edges() const = 0;
+    
+    unsigned int n_elements() const
+    { return element_vec_.size(); }
+
+    virtual unsigned int n_nodes() const = 0;
```
jbrezmorf, line 90: Try to eliminate virtual getters.

src/mesh/mesh.h
```diff
@@ -71,11 +71,118 @@ class BoundarySegment {
     static Input::Type::Record input_type;
 };
 
+
+/// Base class for Mesh and BCMesh.
+class MeshBase {
```
jbrezmorf, line 76: Move to a separate file.

src/mesh/accessors.hh
```diff
@@ -179,17 +180,11 @@ public:
...
-        if (boundary_) return ( element_idx_ - mesh_->bulk_size_ );
-        else return element_idx_;
-    }
-
-    /// Return global idx of element in full element vector
-    unsigned int mesh_idx() const {
         return element_idx_;
     }
 
     unsigned int index() const {
```
jbrezmorf, line 186: Make map in bc mash so that find_elem_id works without "boundary_" flag.

src/mesh/accessors.hh
```diff
@@ -210,11 +205,11 @@ public:
     }
 
     bool operator==(const ElementAccessor<spacedim>& other) const {
```
jbrezmorf, line 207: Need not to test boundary_ flag.

src/mesh/accessors.hh
```diff
@@ -244,7 +239,7 @@ private:
     unsigned int dim_;
 
     /// Pointer to the mesh owning the element.
-    const Mesh *mesh_;
+    const MeshBase *mesh_;
     /// True if the element is boundary
     bool boundary_;
```
jbrezmorf, line 244: Remove boundary_ flag both form data as from constructor.

jbrezmorf (changes requested):

jstebel:
Current status:
- some tests have wrong results: 02/23, 05/07, 25/02, 25/03, 50/21, 50/23, 50/24
- some tests fail on 2 procs in ElementCacheMap::element_eval_point(): 03/07, 03/44, 03/45, 26/03, 27/02

src/mesh/mesh.cc
jstebel: `add_element` is a public function used by mesh readers. Mesh class uses the protected method `add_element_to_vector`.

src/mesh/mesh.cc
jstebel: We need to iterate Element objects (not ElementAccessor) to be able to change its properties inside the loop. That is why elements_range() is not used here.

src/mesh/bc_mesh.cc
jstebel: For empty BCMesh the method elements_range() would fail because it then refers to unallocated element vector.

src/mesh/accessors.hh
```diff
@@ -228,8 +223,8 @@ public:
      centre = elm_ac->node_idx(0);            // short format with dereference operator
  @endcode
      */
-    const Element * operator ->() const {
-    	return &(mesh_->element_vec_[element_idx_]);
+    inline const Element * operator ->() const {
+    	return element();
```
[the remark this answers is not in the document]
jstebel: Both methods (element() and operator->()) are inline so I think it is not a bottleneck in this form.

src/mesh/accessors.hh
```diff
@@ -179,17 +180,11 @@ public:
...
-    }
-
-    /// Return global idx of element in full element vector
-    unsigned int mesh_idx() const {
         return element_idx_;
     }
 
     unsigned int index() const {
-    	return (unsigned int)mesh_->find_elem_id(element_idx_);
+    	return (unsigned int)(boundary_ ? mesh_->get_bc_mesh()->find_elem_id(element_idx_) : mesh_->find_elem_id(element_idx_) );
```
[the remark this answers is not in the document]
jstebel: Done

PR #1256, opened 2020-07-09
Df dg optimize

dflanderka:
- different optimizations in DG assembly and relative methods


src/fields/field.hh
```diff
@@ -241,6 +239,17 @@ public:
      */
     void set_field(const RegionSet &domain, const Input::AbstractRecord &a_rec, double time);
```
jbrezmorf, line 240: remove Field::set_field use Field::set instead

src/fields/multi_field.hh
```diff
@@ -260,9 +260,23 @@ public:
...
+     * Use this method only if necessary.
+     */
+    void set(std::vector<typename Field<spacedim, Value>::FieldBasePtr> field_vec,
+            double time,
+            std::vector<std::string> region_set_names = {"ALL"});
+
+    /**
+     * Same as previous but assigns same field to all components of MultiField.
+     */
+    void set(typename Field<spacedim, Value>::FieldBasePtr field,
```
jbrezmorf, line 277: Rather call previous method  as `set({one_field}, ...)` ... in HeatModel.

PR #1258, opened 2020-07-15
Df dg patch size

dflanderka:


src/coupling/generic_assembly.hh
```diff
@@ -136,83 +139,29 @@ public:
...
-                        multidim_assembly_[2_d]->assemble_fluxes_element_element(edge_integral_data_[i].edge_side_range);
-                        break;
-                    case 3:
-                        multidim_assembly_[3_d]->assemble_fluxes_element_element(edge_integral_data_[i].edge_side_range);
-                        break;
-                    }
-                }
-                END_TIMER("assemble_fluxes_elem_elem");
-            }
+            this->add_integrals_of_computing_step(cell);
```
jbrezmorf, line 153: We have to check that we fit into the preallocated value cache size. 
1. move creation of the ElementCacheMap into Generic assembly
2. add triples (i_reg, i_element, i_point) into a list
3. similarly add to integral lists
4. for every list store: `temporary_size` and `final_size`
5. if the new point or new ghost patch element does not fit into preallocated arrays we reset finish the patch using the final_size, otherwise we set final_size to the temporary_size after processing a single cell.

src/coupling/generic_assembly.hh
```diff
@@ -59,6 +59,9 @@ template < template<IntDim...> class DimAssembly>
 class GenericAssembly
 {
 private:
+    /// Holds maximum number of neighbouring element adding in one step to patch (own element and its side, coupling, boundary neighbours).
+    static const unsigned int maximal_neighbouring_elem = 9;
```
jbrezmorf, line 63: We can remove this arbitrary limit, but we have to check that number of ghost elements of the patch fits into ElementCacheMap.

src/coupling/generic_assembly.hh
jbrezmorf: Try to make auxiliary template class for a list with temporary size.

PR #1261, opened 2020-07-21
Df unified output data

dflanderka:


dflanderka:
Dead-end branch, new algorithm is in branch 'Df unified output data 2'

PR #1277, opened 2020-12-14
Df formula dependency

dflanderka:


src/fields/field_formula.cc
```diff
@@ -244,20 +245,30 @@ void FieldFormula<spacedim, Value>::cache_update(FieldValueCache<typename Value:
...
+            x_[i] = coords(0);
+            y_[i] = coords(1);
+            z_[i] = coords(2);
+        }
+        if (has_depth_) {
+            d_[i] = depth_cache->scalar(i);
         }
         res_[i] = 0.0;
     }
+    for (auto it : eval_field_data_) {
```
jbrezmorf, line 264: Try to unify copies of coords and depth using eval_field_data_ as well.

src/fields/field_set.hh
```diff
@@ -341,47 +344,50 @@ public:
...
+    inline FieldValueCache<double> * coords_cache() {
+        return X_.value_cache();
+    }
+
+    /// Return FieldValueCache of Fielddepth (depth_)
+    inline FieldValueCache<double> * depth_cache() {
+        return depth_.value_cache();
+    }
+
+    inline void set_surface_depth(std::shared_ptr<SurfaceDepth> surface_depth) {
```
jbrezmorf, line 364: Try to remove these two getters.

PR #1289, opened 2021-02-22
Df optimize fe eval

dflanderka:
Try to optimize FieldFE evaluation for vector and tensor valued fields. Doing first DOF linear combination and then transform by Jacobian.

### DIFF
--- a/src/fields/field_fe.cc
+++ b/src/fields/field_fe.cc
@@ -299,6 +299,7 @@ void FieldFE<spacedim, Value>::cache_update(FieldValueCache<typename Value::elem
 {
     ASSERT( !boundary_dofs_ ).error("boundary field NOT supported!!\n");
     Armor::ArmaMat<typename Value::element_type, Value::NRows_, Value::NCols_> mat_value;
+    Armor::ArmaMat<double, Value::NRows_, Value::NCols_> dbl_mat_value;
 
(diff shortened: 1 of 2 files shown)


src/fields/field_fe.cc
```diff
@@ -259,6 +259,7 @@ void FieldFE<spacedim, Value>::cache_update(FieldValueCache<typename Value::elem
 {
     ASSERT( !boundary_dofs_ ).error("boundary field NOT supported!!\n");
     Armor::ArmaMat<typename Value::element_type, Value::NRows_, Value::NCols_> mat_value;
```
jbrezmorf, line 261: TODO:
1. measure in proper context, proof if it is worth of 
2. more general implementation for other types of Mappings

src/fields/field_fe.cc
```diff
@@ -259,6 +259,7 @@ void FieldFE<spacedim, Value>::cache_update(FieldValueCache<typename Value::elem
...
+    Armor::ArmaMat<double, Value::NRows_, Value::NCols_> dbl_mat_value;
```
jbrezmorf, line 262: Why we need `dbl_mat_value`? mat_value not sufficient?

dflanderka:
Deprecated changes. Modification of FieldFE will be solved using PatchFeValues.

PR #1290, opened 2021-02-24
Edit assembly_fluxes_element_element() to pass to vectorized operations.

jstebel:


src/transport/assembly_dg.hh
```diff
@@ -841,41 +852,53 @@ public:
                             for (unsigned int i=0; i<fe_values_vec_[sd[n]].n_dofs(); i++)
                                 for (unsigned int j=0; j<fe_values_vec_[sd[m]].n_dofs(); j++)
                                     local_matrix_[i*fe_values_vec_[sd[m]].n_dofs()+j] = 0;
-
+
```
jbrezmorf, line 855: 1. Must avoid dynamically allocated vectors. Allocations kill performance, indirect access isn't good neither.
2. Precompute before n,m loops once for every side.

jstebel: Allocation can be done in advance, before the n,m loops. However, precomputation is done not for every side, but for every PAIR of sides. Do we really want it?

jbrezmorf:
Please, move it to the branch DF_transport_asm. There we would make the performance tests before we merge it finally to the master.

PR #1291, opened 2021-02-26
Preallocate and precompute edge integral data.

jstebel:
Optimize assembly of edge integral by preallocating and precomputing the averages and jumps of shape functions.

### DIFF
--- a/src/transport/assembly_dg.hh
+++ b/src/transport/assembly_dg.hh
@@ -493,7 +493,7 @@ public:
 
                                         local_matrix_[index] += (
                                             // flux due to transport (applied on interior edges) (average times jump)
-                                            transport_flux*jumps[n][k*fe_->n_dofs()+i]*averages[m][k*fe_->n_dofs()+j]
+                                            transport_flux*jumps[n][k*fe_->n_dofs()+i]*averages[sd[m]][k*fe_->n_dofs()+j]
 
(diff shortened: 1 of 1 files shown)


src/transport/assembly_dg.hh
```diff
@@ -423,8 +453,21 @@ public:
...
+                    // precompute jumps and weighted averages of shape functions over the pair of sides (s1,s2)
+                    k=0;
+                    for (auto p1 : data_->stiffness_assembly_->edge_points(edge_side1) )
+                    {
+                        auto p2 = p1.point_on(edge_side2);
+                        for (unsigned int i=0; i<fe_->n_dofs(); i++)
+                        {
+                            for (int n=0; n<2; n++)
+                            {
+                                jumps[n][k*fe_->n_dofs()+i] = (n==0)*fe_values_vec_[s1].shape_value(i,k) - (n==1)*fe_values_vec_[s2].shape_value(i,k);
```
jbrezmorf, line 465: Seems that jumps[0] == jumps[1], optimization possible.
We will further make an abstract storage for precomputed values for pairs of sides in order to hide exact placement in the memory.

jbrezmorf:
This is fine. Thanks. We will continue with David on it.

src/transport/assembly_dg.hh
jstebel: I don't think so. Shape functions can be ordered differently on opposite Edge sides.

PR #1294, opened 2021-03-24
Df optimize asm points

dflanderka:


src/fields/field_set.cc
```diff
@@ -193,9 +194,9 @@ bool FieldSet::is_jump_time() const {
 
 void FieldSet::cache_update(ElementCacheMap &cache_map) {
     ASSERT_GT_DBG(region_field_update_order_.size(), 0).error("Variable 'region_dependency_list' is empty. Did you call 'set_dependency' method?\n");
-    for (unsigned int i_reg=0; i_reg<cache_map.n_regions(); ++i_reg) {
-        unsigned int region_idx = cache_map.eval_point_data( cache_map.region_chunk_by_map_index(i_reg) ).i_reg_;
-        for (const FieldCommon *field : region_field_update_order_[region_idx]) field->cache_update(cache_map, region_idx);
+    for (unsigned int i_reg_patch=0; i_reg_patch<cache_map.n_regions(); ++i_reg_patch) {
+        unsigned int region_idx = cache_map.eval_point_data( cache_map.region_chunk_by_map_index(i_reg_patch) ).i_reg_;
```
jbrezmorf, line 198: Make a mapping method of cache_map and need not to pass region_idx explicitely

src/fields/field.impl.hh
```diff
@@ -771,9 +771,9 @@ void Field<spacedim, Value>::cache_reallocate(const ElementCacheMap &cache_map)
 
 
 template<int spacedim, class Value>
-void Field<spacedim, Value>::cache_update(ElementCacheMap &cache_map, unsigned int i_reg) const {
-    if (region_fields_[i_reg] != nullptr) // skips bounadry regions for bulk fields and vice versa
-        region_fields_[i_reg]->cache_update(value_cache_, cache_map, i_reg);
+void Field<spacedim, Value>::cache_update(ElementCacheMap &cache_map, unsigned int region_idx, unsigned int region_patch_idx) const {
+    if (region_fields_[region_idx] != nullptr) // skips bounadry regions for bulk fields and vice versa
```
jbrezmorf, line 775: Move obtaining of region_idx from region_patch_idx here (appropriate method of cache_map)

src/fields/field_value_cache.hh
```diff
@@ -197,30 +220,40 @@ public:
...
     /// Return end position of element chunk in FieldValueCache
     inline unsigned int element_chunk_end(unsigned int mesh_elm_idx) const {
-        std::map<unsigned int, unsigned int>::const_iterator it = element_to_map_.find(mesh_elm_idx);
+        std::unordered_map<unsigned int, unsigned int>::const_iterator it = element_to_map_.find(mesh_elm_idx);
         if ( it != element_to_map_.end() ) return element_starts_[it->second+1];
         else return ElementCacheMap::undef_elem_idx;
     }
 
+    /// Return begin position of element chunk in FieldValueCache
+    inline unsigned int element_chunk_begin_new(unsigned int elm_patch_idx) const {
```
jbrezmorf, line 236: Remove "old" variants and possibly introduce `find_patch_element(mesh_elem_idx)`.

src/fields/field_value_cache.cc
```diff
@@ -37,7 +37,8 @@ ElementCacheMap::ElementCacheMap()
 : elm_idx_(CacheMapElementNumber::get(), ElementCacheMap::undef_elem_idx),
   ready_to_reading_(false), element_eval_points_map_(nullptr), eval_point_data_(0),
   regions_starts_(2*ElementCacheMap::regions_in_chunk,ElementCacheMap::regions_in_chunk),
-  element_starts_(2*ElementCacheMap::elements_in_chunk,ElementCacheMap::elements_in_chunk) {}
+  element_starts_(2*ElementCacheMap::elements_in_chunk,ElementCacheMap::elements_in_chunk),
+  dh_(nullptr) {}
```
jbrezmorf, line 41: Why we need DH here? It seems inappropriate since there may be multiple DH involved in single assembly.

src/tools/revertable_list.hh
```diff
@@ -83,19 +83,43 @@ public:
...
+     * Method passes argumets of Type constructor.
+     */
+    template<class... Args>
+    inline std::size_t emplace_back(Args&&... args)
+    {
+        ASSERT_DBG((enlardeg_by_ > 0) || (temporary_size_ < reserved_size())).error("Data array overflowed!\n");
+        if (temporary_size_ == reserved_size()) { // enlarge reserved size
+        	this->resize( this->reserved_size() + enlardeg_by_ );
+        }
+        data_[temporary_size_] = Type( std::forward<Args>(args)... );
```
jbrezmorf, line 118: This doesnůt prevent duplicate creation of Type. 
If the reserved size is OK, then you can use `data_.emplace` to create the instance right in the vector.

src/transport/assembly_dg.hh
```diff
@@ -63,9 +64,12 @@ public:
 
 
     /// Assemble integral over element
-    inline void cell_integral(DHCellAccessor cell)
+    inline void cell_integral(unsigned int element_patch_idx, unsigned int dh_local_idx)
     {
-        ASSERT_EQ_DBG(cell.dim(), dim).error("Dimension of element mismatch!");
+        if ((int)dh_local_idx == -1) return;
+        DHCellAccessor cell(data_->dh_.get(), dh_local_idx);
```
jbrezmorf, line 70: Avoid recreation of the cell in cell_integral implementations.

src/transport/concentration_model.hh
```diff
@@ -133,6 +133,46 @@ public:
         { return substances_; }
 
 
+        /// @name Set of methods returning vectors of field names using in different assemblations.
+        // @{
+
+        /// Returns vector of field names of Mass assembly.
+        inline std::vector<string> mass_assembly_subset() const {
```
jbrezmorf, line 140: assembly subsets should be independent of the model since the dependency is handled automatically. We only need to mention fields used directly in assembly methods.

src/transport/transport_dg.cc
```diff
@@ -307,11 +307,11 @@ void TransportDG<Model>::initialize()
 
 
     // create assemblation object, finite element structures and distribute DOFs
-	data_->mass_assembly_ = new GenericAssembly< MassAssemblyDim >(data_.get(), this->balance());
-	data_->stiffness_assembly_ = new GenericAssembly< StiffnessAssemblyDim >(data_.get(), this->balance());
-	data_->sources_assembly_ = new GenericAssembly< SourcesAssemblyDim >(data_.get(), this->balance());
-	data_->bdr_cond_assembly_ = new GenericAssembly< BdrConditionAssemblyDim >(data_.get(), this->balance());
-	data_->init_cond_assembly_ = new GenericAssembly< InitConditionAssemblyDim >(data_.get(), this->balance());
+	data_->mass_assembly_ = new GenericAssembly< MassAssemblyDim >(data_.get(), this->balance(), data_->dh_.get());
```
jbrezmorf, line 310: Why passing DH?

jbrezmorf (commented):
There are possibly some places that get ugly and it is not clear if they improve performance.
These must be confirmed.

src/coupling/assembly_base.hh
```diff
@@ -46,7 +46,7 @@ public:
     }
 
     /// Assembles the volume integrals on cell.
-    inline void cell_integral(FMT_UNUSED DHCellAccessor cell) {}
+    inline void cell_integral(FMT_UNUSED unsigned int element_patch_idx, FMT_UNUSED unsigned int dh_local_idx) {}
```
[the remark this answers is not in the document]
dflanderka: It makes cell_integral method (call of point range) faster. We can't use same optimization in other types of integrals because they need access to equivalent points at lower dim element throught neighbour object etc.

src/fields/eval_subset.hh
```diff
@@ -295,17 +293,14 @@ public:
...
     }
 
     /// Returns range of bulk local points for appropriate cell accessor
-    inline Range< BulkPoint > points(const DHCellAccessor &cell, const ElementCacheMap *elm_cache_map) const {
-        ASSERT_DBG(cell.element_cache_index() != ElementCacheMap::undef_elem_idx)(cell.elm_idx())
-                .error("Undefined element cache index!\n");
-
-        auto bgn_it = make_iter<BulkPoint>( BulkPoint(cell, elm_cache_map, this, eval_points_->subset_begin(dim_, subset_index_)) );
-        auto end_it = make_iter<BulkPoint>( BulkPoint(cell, elm_cache_map, this, eval_points_->subset_end(dim_, subset_index_)) );
+    inline Range< BulkPoint > points(unsigned int element_patch_idx, const ElementCacheMap *elm_cache_map) const {
```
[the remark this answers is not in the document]
dflanderka: Yes, explanation in comment: https://github.com/flow123d/flow123d/pull/1294#discussion_r625236493

src/coupling/assembly_base.hh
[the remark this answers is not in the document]
jbrezmorf: Points range call is fine. Just try to measure passing the cell or cell reference vs. new parameters.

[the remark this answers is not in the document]
dflanderka: OK, new parameters are not faster then passing the cell. Changes partly revert.

src/fields/eval_subset.hh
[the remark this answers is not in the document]
jbrezmorf: I see, element patch index is not part of cell.

PR #1295, opened 2021-03-26
Stress boundary condition in mechanics.

jstebel:


src/coupling/hm_iterative.cc
```diff
@@ -251,11 +264,29 @@ void HM_Iterative::update_potential()
             double potential = -alpha*density*gravity*pressure;
         
             potential_vec_.set( dof_indices[side.side_idx()], potential );
+
+            // The reference potential is applied only on dirichlet and total_flux b.c.,
+            // i.e. where only mechanical traction is prescribed.
+            if (side.side().is_boundary() &&
+                    (flow_->data().bc_type.value(side.centre(), side.cond().element_accessor()) == DarcyMH::EqData::dirichlet ||
```
jbrezmorf, line 271: We can not use .value to evaluate fields anymore. Need to refactor this as an auxiliary assembly class.

src/coupling/hm_iterative.cc
```diff
@@ -102,6 +109,9 @@ void HM_Iterative::EqData::initialize(Mesh &mesh)
     
     potential_ptr_ = create_field_fe<3, FieldValue<3>::Scalar>(mesh, MixedPtr<FE_CR>());
     pressure_potential.set(potential_ptr_, 0.0);
+
+    ref_potential_ptr_ = create_field_fe<3, FieldValue<3>::Scalar>(mesh, MixedPtr<FE_CR>());
+    ref_pressure_potential.set(ref_potential_ptr_, 0.0);
```
jbrezmorf, line 114: What prevents passing a copy of flow bc_pressure to elasticity? We could possibly add setting to zero for the Neumann boundary condition. It should also work for Seepage and River BC in Darcy.

jbrezmorf (commented):
Have to discuss some technicalities. I suggest merging followed by a swift refactor to assembly class in HM iterative and separate issue for passing bc_pressure and exact meaning of ref_pressure.

src/coupling/hm_iterative.cc
jstebel: The reason is to minimize the interface between coupled processes. Mechanics should not be spoiled by external fields from flow or other equations.

jbrezmorf: OK. So copy the comments to a new issue and merge this branch. In a separate branch we can then test what exactly prevents:
1. usage of copy of bc_pressure (without .value call)
2. having FieldModel of the boundary field ref_pressure as a copy of bc_pressure

jstebel:
Ready to merge.

jbrezmorf:
Po sloučení s master nefunguje test 50/01. Může být způsobeno permutací nodů v elementech kvůli odstranění permutací. Prosím, zkontroluj a případně aktualizuj referenční data.

PR #1299, opened 2021-04-12
Df mechanic assembly

dflanderka:
Application of new assembly implementation in Elasticity class.


src/mechanics/elasticity.cc
```diff
@@ -398,34 +480,40 @@ void Elasticity::update_output_fields()
...
-	data_.set_time(time_->step(), LimitSide::right);
+	eq_fields_->mark_input_times( *time_ );
+	eq_fields_->set_time(time_->step(), LimitSide::right);
 	std::stringstream ss; // print warning message with table of uninitialized fields
 	if ( FieldCommon::print_message_table(ss, "mechanics") ) {
 		WarningOut() << ss.str();
 	}
 
-    ( (LinSys_PETSC *)ls )->set_initial_guess_nonzero();
+    ( (LinSys_PETSC *)eq_data_->ls )->set_initial_guess_nonzero();
```
jbrezmorf, line 490: If possible move construction and initialization of the LinSys into single method avoiding typecasts.

src/mechanics/elasticity.cc
```diff
@@ -473,43 +567,49 @@ void Elasticity::next_time()
...
+        || eq_fields_->subset(FieldFlag::in_main_matrix).changed())
     {
         DebugOut() << "Mechanics: Assembling matrix.\n";
-        ls->start_add_assembly();
-        ls->mat_zero_entries();
-        assemble_stiffness_matrix();
-        ls->finish_assembly();
+        eq_data_->ls->start_add_assembly();
+        eq_data_->ls->mat_zero_entries();
+        START_TIMER("assemble_stiffness");
```
jbrezmorf, line 580: Move this timer into the assemble method? As the GenericAssembly is 
specialized we can take the FRAME name from the Assemble<> template.

src/mechanics/elasticity.cc
```diff
@@ -473,43 +567,49 @@ void Elasticity::next_time()
...
+        stiffness_assembly_->assemble(eq_data_->dh_);
+        END_TIMER("assemble_stiffness");
+        //assemble_stiffness_matrix();
+        eq_data_->ls->finish_assembly();
 
         if (stiffness_matrix == NULL)
-            MatConvert(*( ls->get_matrix() ), MATSAME, MAT_INITIAL_MATRIX, &stiffness_matrix);
+            MatConvert(*( eq_data_->ls->get_matrix() ), MATSAME, MAT_INITIAL_MATRIX, &stiffness_matrix);
```
jbrezmorf, line 587: This smells, separate issue #1316.

src/mechanics/elasticity.cc
```diff
@@ -473,43 +567,49 @@ void Elasticity::next_time()
...
+        eq_data_->ls->start_add_assembly();
+        eq_data_->ls->rhs_zero_entries();
+    	//assemble_rhs();
+        START_TIMER("assemble_rhs");
+        rhs_assembly_->assemble(eq_data_->dh_);
+        END_TIMER("assemble_rhs");
+        eq_data_->ls->finish_assembly();
+
+        if (rhs == nullptr) VecDuplicate(*( eq_data_->ls->get_rhs() ), &rhs);
+        VecCopy(*( eq_data_->ls->get_rhs() ), rhs);
```
jbrezmorf, line 606: @jstebel Why we need a copy?

src/mechanics/elasticity.cc
```diff
@@ -523,21 +623,21 @@ void Elasticity::output_data()
...
+//     balance_->calculate_instant(subst_idx, eq_data_->ls->get_solution());
 //     balance_->output();
 //     END_TIMER("MECH-balance");
 
     END_TIMER("MECH-OUTPUT");
 }
 
 
-template<unsigned int dim>
+/*template<unsigned int dim>
```
jbrezmorf, line 640: Please, use rather // line comments as it works better with GIT diffs.

src/mechanics/elasticity.hh
```diff
@@ -188,14 +223,17 @@ public:
...
 
 	const Vec &get_solution()
-	{ return ls->get_solution(); }
+	{ return eq_data_->ls->get_solution(); }
+
+	inline EqFields &eq_fields() { return *eq_fields_; }
+
+	inline EqData &eq_data() { return *eq_data_; }
 
-    inline EqData &data() { return data_; }
```
jbrezmorf, line 198: Where this was used? Do not see declaration of data_. How it is related to the coupling?

src/mechanics/assembly_elasticity.hh
```diff
@@ -0,0 +1,695 @@
...
+
+                for (unsigned int i=0; i<n_dofs_ngh_[n]; i++)
+                {
+                    arma::vec3 vi = (n==0) ? arma::zeros(3) : vec_view_side_->value(i,k);
+                    arma::vec3 vf = (n==1) ? arma::zeros(3) : vec_view_sub_->value(i,k);
+                    arma::mat33 gvft = (n==0) ? vec_view_sub_->grad(i,k) : arma::zeros(3,3);
+
+                    for (int m=0; m<2; m++)
+                        for (unsigned int j=0; j<n_dofs_ngh_[m]; j++) {
+                            arma::vec3 ui = (m==0) ? arma::zeros(3) : vec_view_side_->value(j,k);
```
jbrezmorf, line 210: @jstebel We should avoid tests for individual dofs as it fails to generalize for different FE. And it is also slow as it breaks CPU pipelining.

src/mechanics/assembly_elasticity.hh
```diff
@@ -0,0 +1,695 @@
...
+                            "lame_mu", "lame_lambda", "young_modulus", "poisson_ratio", "bc_type", "bc_displacement",
+		                    "bc_traction", "fracture_sigma"};
+        this->used_fields_ = eq_fields_->subset( sub_names );
+    }
+
+    /// Destructor.
+    ~RhsAssemblyElasticity() {}
+
+    /// Initialize auxiliary vectors and other data members
+    void initialize(FMT_UNUSED std::shared_ptr<Balance> balance, ElementCacheMap *element_cache_map) {
```
jbrezmorf, line 306: Can we make a common function for creation of objects shared between RHS and Stiffness assembly?

src/mechanics/elasticity.cc
```diff
@@ -122,7 +125,7 @@ std::shared_ptr<DOFHandlerMultiDim> FEObjects::dh_tensor() { return dh_tensor_;
 
 
 
-} // namespace Mechanics
+} // namespace Mechanics */
 
 
 double lame_mu(double young, double poisson)
```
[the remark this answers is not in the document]
dflanderka: Used in HM_iterative yet, TODO comment added.

src/mechanics/elasticity.cc
dflanderka: Will be solved together with DG after merge.

PR #1303, opened 2021-04-29
Df revision asm data

dflanderka:


jbrezmorf (approved):
OK. Can merge, but make a branch for DF_transport_asm before the merge so that we can review Model changes later on.

PR #1304, opened 2021-05-03
Df vector mpi

dflanderka:
Replace method data() and operator [] of VectorMPI.
New methods: get, set, add, divisor.


jbrezmorf (commented):
Simplification of the interface is fine. We should consider a similar code for filling the output caches.

jbrezmorf:
Filling of output caches is a form of assembly and should be replaced by more general algorithm for field interpolation from one to other discretization. Then we may only need a local copy of a VecMPI into output cache.

result:
- comment out [] operator
- introduce normalize operation as in output cache

Issue #1305, opened 2021-05-06
Independent assembly subsets.


dflanderka:
Assembly subsets should be independent of the model since the dependency is handled automatically. We only need to mention fields used directly in assembly methods. Methods xxx_assembly_subset in ConcentationModel and HeastModel must be removed.
- Assembly implementation classes would define a FieldSet (subset) of fields directly used in the assembly.
- This subset of used fields will be passed as parameter to EqFields::cache_reallocate(cache_map, used_fieldset).
- cache_reallocate will call set_dependency automatically
- FieldSet::set_dependency iterate through the used fields in the main fields_range loop.
- cache_reallocate will allocate cache for all really used fields detected by set_dependency.
- full EqFields Fieldset will be passed to individual fields, necessary in FieldModel and FieldFormula
- result is stored to region_field_update_order_, this structure will contain only necessary fields for cache_update
- FieldFormula makes parsing step only if the formula is changed
- FieldFormula resets the field pointers and recompile parser at every set_dependency call

jbrezmorf:
Resolving #1282 will possibly need splitting FieldFormula::set_dependency into two methods
1. getting a list of parent fields
2. set reallocated caches of these fields

jbrezmorf:
Some implementation improvements moved to new issue #1320 , otherwise done.

PR #1314, opened 2021-05-18
Df independent subsets

dflanderka:
Solve issue #1305, complete tests of FieldFormula


src/transport/assembly_dg.hh
```diff
@@ -130,8 +134,8 @@ public:
     /// Implements @p AssemblyBase::reallocate_cache.
     void reallocate_cache() override
```
jbrezmorf, line 135: Can we move this method to GenericAssembly? Just name every particular assembly class or use `type(T).name()`
with particular assembly class `T`.

jbrezmorf (commented):
- Substantial suggestions as code comments in the last commit.
- Possible simplification of `Assembly<>::cache_reallocate`.

src/fields/field_formula.hh
```diff
@@ -49,12 +49,17 @@ using namespace std;
...
+ * https://github.com/flow123d/bparser/
+ * http://warp.povusers.org/FunctionParser/ (gradually replaced by BParser)
+ *
+ * Allows parsing:
+ * - base variables: coordinates (x,y,z), time (t), surface depth (d); constants: e, pi
+ * - standard functions: trigonometric functions, min and max function, exponential function, ternary operator etc.
+ * - expressions dependendent on other fields
  *
  * TODO:
  * correct support for discrete functions (use integer parser), actually we just convert double to int
```
dflanderka, line 62: Delete TODO

Issue #1321, opened 2021-06-02
Unit in record TimeValue description.


Paulie14:
The tuple `TimeValue` (e.g. defining the time of input fields of equation) has a key `unit` which is a `String`.
And its description states that the user can use:  "Possible values: 's' seconds, 'min' minutes, 'h' hours, 'd' days, 'y' years."
However, the string is actually parsed by the `UnitConverter`, which allows much more.

Example usage:
```
      - region: .BOUNDARY
        time:
          time: 1
          unit: year; year = 365.2425*d
        bc_type: dirichlet
```
where we defined the correct length of the year due to leap years.

In contrast to that, the records for various types of fields `Field*` has a key `unit`, which
is actually a record with proper description of the unit converter functionality.

Can we move the `unit_record` from `field_algo_base_impl.hh:95` to `unit_si.cc`?
And use it properly also in the TimeValue?

Then it would behave the same naturally and also provide the same description in generated doc.

Paulie14:
Fixed by 832ae70694015811993c188bd3539217ed3a6e9a.

Paulie14:
Finished by PR #1322.

PR #1322, opened 2021-06-02
Pe timeunits

Paulie14:
solving issue #1321


src/tools/time_governor.cc
```diff
@@ -48,8 +49,9 @@ const Tuple & TimeGovernor::get_input_time_type(double lower_bound, double upper
     return Tuple("TimeValue", "A time with optional unit specification.")
         .declare_key("time", Double(lower_bound, upper_bound), Default::obligatory(),
                                     "The time value." )
-		.declare_key("unit", String(), Default::read_time("Common time unit of the equation's Time Governor."),
-				"The time unit. Possible values: 's' seconds, 'min' minutes, 'h' hours, 'd' days, 'y' years.")
+		.declare_key("unit", UnitConverter::get_input_type(),
+                Default::read_time("The time unit."),
```
jbrezmorf, line 53: Keep previous default value description. Related is `TimeGovernor.common_time_unit` that has to be changed as well.

src/tools/time_governor.cc
```diff
@@ -48,8 +49,9 @@ const Tuple & TimeGovernor::get_input_time_type(double lower_bound, double upper
...
+				"User can benefit from the Unit Convertor funcionality, e.g. creating own time units.")
```
jbrezmorf, line 54: Have to check all places accessing "time" and "unit" and call `UnitConverter::convert("s")`.

jbrezmorf: Seems that TimeUnitConversion is the right place.

jbrezmorf (commented):
see comments
+ make an integration test with common_time_unit in 02_gneric_input

Paulie14:
`common_time_unit` is already tested in 02/11, 02/12, 02/17 - I suppose that this is satisfying

Issue #963, opened 2018-06-22
Native IO for conc.


jbrezmorf:
@Paulie14 
- [x]  Create test A for writing a native conc data of TransportDG, three variants: P0, P1, P2.
        Square domain, subsquares with significantly different conductivity, pressure jump.
        BC conc on left conductive part of boundary. Simulation time till steady state.
        Write native conc in steady state.
        After steady state time, turn off conc on inflow BC.
        Write subsequent simulated conc as reference data for the test B.

- [x] Create test B for reading init conc from native data.
        Same geometry and flow as in test A.
        Read init conc from native format.
        Check that the conc simulation is the same as in test B.

@dflanderka 
- [ ]  Test that test A CORNER data output is same as the NATIVE output of test A once it is read in test B and output as CRONER data.
- [ ] @jstebel How to turn off interpolation of initial concentration fi given as NATIVE.
- [ ] Fix lists of Scalars, Vectors, Tensors for the Flow123dData tag of the native output.

jbrezmorf:
Test A probably 03/43 (currently commented) test B probably 03/42 (currently commented).

@Paulie14 
TODO: test it

Paulie14:
The tests were prepared for @dflanderka , so that he could further work on native IO.
I tried the native output of conc in test 43 (testA), but it fails on the assert:
```
Program Error: Violated assert! 
      ##                   > In file: /home/paulie/Workspace/flow123d/src/fields/field_fe.cc(817): Throw in function native_data_to_cache
      ##                   > Expression: 'output_data_cache.n_values() * output_data_cache.n_comp() == dh_->distr()->lsize()'
      ##                   > Values:
      ##                     output_data_cache.n_values() * output_data_cache.n_comp() : '1000'
      ##                     dh_->distr()->lsize() : '3000'
      ##                   
      ##                   ** Stacktrace **
      ##                     0  FieldFE<3, FieldValue_<1, 1, double> >::native_data_to_cache(ElementDataCache<double>&)
      ##                     1  Field<3, FieldValue_<1, 1, double> >::fill_data_cache(OutputTime::DiscreteSpace, std::shared_ptr<OutputTime>, std::shared_ptr<ElementDataCache<double> >)
      ##                     2  Field<3, FieldValue_<1, 1, double> >::compute_field_data(OutputTime::DiscreteSpace, std::shared_ptr<OutputTime>)
[...]

PR #1347, opened 2021-07-02
Df fix native io

dflanderka:
- solves part of #963


src/fields/field.impl.hh
```diff
@@ -723,7 +725,10 @@ void Field<spacedim,Value>::fill_data_cache(OutputTime::DiscreteSpace space_type
 
             if (field_fe_ptr) {
                 auto native_output_data_base = stream->prepare_compute_data<double>(this->name(), space_type,
-                        (unsigned int)Value::NRows_, (unsigned int)Value::NCols_);
+                        (unsigned int)Value::NRows_, (unsigned int)Value::NCols_,
+                        typeid(field_fe_ptr->get_dofhandler()->ds()->fe()[0_d].get()).name(),  // should be used better solution of fe_type setting
+                                                                                               // e.g. method 'name()' of FiniteElement and descendants
```
dflanderka, line 730: Use better setting of fe_type (eg. special method of FiniteElement and its descendants).
Add fe_type to output file and check correct outputs of appropriate tests.

PR #1350, opened 2021-07-08
Df asm field output

dflanderka:
- implement multiple interpolations of output 
- application of new assembly system of output evaluation.


src/transport/transport_dg.cc
```diff
@@ -260,6 +260,7 @@ void TransportDG<Model>::initialize()
     eq_data_->output_vec.resize(eq_data_->n_substances());
     eq_fields_->output_field.set_components(eq_data_->substances_.names());
     eq_fields_->output_field.set_mesh(*Model::mesh_);
+    eq_fields_->output_fields.set_mesh(*Model::mesh_);
```
dflanderka, line 263: Remove duplicate call.

src/flow/darcy_flow_lmh.cc
```diff
@@ -73,6 +74,28 @@ FLOW123D_FORCE_LINK_IN_CHILD(darcy_flow_lmh)
 
 
 
+/*******************************************************************************
```
jbrezmorf, line 77: Move to single header file common for both MH and LMH.

src/flow/darcy_flow_mh_output.cc
```diff
@@ -314,7 +314,7 @@ void DarcyFlowMHOutput::output_internal_flow_data()
         raw_output_file << fmt::format("{} {} ", dh_cell.elm().index(), data->full_solution.get(indices[ele->n_sides()]));
         
         // velocity at element center
-        flux_in_center = data->field_ele_velocity.value(ele.centre(), ele);
+        flux_in_center = data->flux.value(ele.centre(), ele) / data->cross_section.value(ele.centre(), ele); // TODO use new assembly
```
jbrezmorf, line 317: Keep original, Model has value methdo.

src/io/output_mesh.hh
```diff
@@ -132,6 +132,13 @@ public:
 		else return shared_from_this();
 	};
 
+	/// Return local element index to its global index.
+	inline LongIdx get_loc_elem_idx(LongIdx global_idx) {
```
jbrezmorf, line 136: Is just "temporary" solution. comment where it is used, and that this should not be used elsewhere.

src/fields/assembly_output.hh
```diff
@@ -0,0 +1,183 @@
...
+
+
+    /// Assemble integral over element
+    inline void cell_integral(DHCellAccessor cell, unsigned int element_patch_idx)
+    {
+        ASSERT_EQ_DBG(cell.dim(), dim).error("Dimension of element mismatch!");
+
+        auto p = *( this->bulk_points(element_patch_idx).begin() ); // evaluation point (in element center)
+        unsigned int val_idx = stream_->get_output_mesh_ptr()->get_loc_elem_idx(cell.elm_idx());
+        for (FieldListAccessor f_acc : used_fields_.fields_range()) {
```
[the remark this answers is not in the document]
dflanderka: Replace `output_data_cache` by setter solves this problem. We don't need `used_caches_` now.

Issue #208, opened 2014-08-19
doxygen documentation


jbrezmorf:
- [ ] add flow123d version into title
- add and/or update documentation for stable classes:
   - [ ] input classes
   - [ ] field classes
   - [ ] exceptions
   - [ ] profiler
   - [ ] time governor
   - ... append
- [ ] add stable classes to the main page, add hand written documentation

- propose changes in format of pages to enhence readability

jbrezmorf:
Duplicate of #920.

jbrezmorf:
Doxygen doc builds just fine and is published with docker images built on Jenkins.

TODO for ver 3.1.0:
- [ ] fix GIT_DESCRIBE to get correct information about branch and commit on the main page. See doc/CMakeLists.txt.
- [ ] Update authors.
- [ ] Review texts, update.
- [ ] Describe new assembly process, use doc/design/field_evaluate.md

jbrezmorf:
Further improvements are of low priority. Postponed to 5.0.

PR #1356, opened 2021-07-21
Df doxygen update

dflanderka:
- solves #208


jbrezmorf:
Try if it works with doxygen, review formating.

PR #1358, opened 2021-07-21
Pe tests

Paulie14:
- fix test that were not running (accidentally skipped)
- fix runtest script - check that filenames in all the configs exist
- fix and rename tutorial tests


jbrezmorf (approved):
Great.

PR #1364, opened 2021-07-28
Remove dead code concerning CFL in reactions.

Paulie14:
- sofar only removing dead code
(solution of CFL postponed)


jbrezmorf:
Check tests and merge.

PR #1367, opened 2021-08-11
Jb perm 1

jbrezmorf:
- Permute element nodes to get a matching orientation of element faces and edges. 

- Have to introduce inverted elements with negative Jacobian. Must be taken into account in FEM implementation in particular calculation of normals. *In fact not a big deal since normals seem to be computed for boundary elements.*

- Remove explicit permutation of faces.

- Minor changes in intersections via. Plucker coordinates
 
- Adapted reference data (order) in unit tests.

Essential change in mesh.cc, `canonical_faces`. Idea is to orient every edge the node with larger index. 
This automatically leads to common permutation of faces, but request element nodes to be in increased order.

Subsequently there are many changes removing permutation related structures, arguments ...

Changes in `reference_manual` are due to updated generated documentation, including tutorials.

More changes and simplifications in `eval_points` and `eval_subsets`. Ulocking opportunity to further simplifications.

Careful review and fix of `compute_intersections`.

Intorduce `undef_idx` in them main name space not only in the Mesh class.

Simplification of the ElementAccessor data members.

Changes in the reference data, checked manually. 
TODO: revert usage of VTK interpolated comparison, change more reference files.

Add VTK based comparison of the test results, used by default. Support for the comparison of interpolated data (can partially hide mesh permutations)


jbrezmorf:
Merged manually.

Issue #1366, opened 2021-08-10
Reading FieldFE from gmsh is broken - reconsider component indexing approach.


Paulie14:
Issue based on observations in PR #1361

Reconsider component (substance) indexing approach:
We used to support multiple substance values as vectors inside `ElementData` section in MSH format in the past.
This is no longer supported, although some parts of code still exist (especially in `get_element_data` in `msh_basereader.cc`).
This approach was also not appropriate due to the fact that the substance names were not written in the msh format and 
the order of substances must have been kept (equal to order in yaml) while reading.
Currently the MSH format should behave more like VTK - for each component (substance) have its own `ElementData` section.
The field names would then be same as in VTK, e.g. `A_conc`, `B_conc` for substances `A` and `B`.
This approach would also enable to read (in MSH) what is written by Flow123d (in MHS) which does work for substance components at the moment.

TODO:
- [x] change default value of `component_idx` in FieldAlgorithmBase (`.impl.hh`) to zero (currently undef_int, and then checked in heat eq.)
- [x] both VTK and MSH will not support components (substances) - remove `can_have_components_` flag
- [x] then simplify the `get_element_data` function, avoid some of the conditions [Partialy done.]
- [x] prepare tests for writing and reading MSH data (analogical to test 03-44 and 03-45 for VTK)
- [ ] the interface of the reader lacks proper documentation
- [x] consider merging two phase reading: `find_header` followed by `get_element_data` in `FieldFE`
- [ ] consider further the polymorphism in reader classes (what is common and what is format dependent - see `MeshDataHeader` struct)

Paulie14:
@jbrezmorf please confirm the statements and todo list

Paulie14:
- seems that we can remove the `component_idx` from the reader altogether
- then also `CacheData` in `ElementDataCache` would be always vector of size 1 (every substance field would have its own `ElementDataCache`, correct?)

Paulie14:
I removed `component_idx` and `can_have_components_` and the condition from `get_element_data`.
I have now working test on write-read MSH data with `equivalent mesh`.

However using `identic_mesh`, the element data are permuted and I cannot figure out how..
Similar behaviour can be seen with VTK.
If this not the case for using identic mesh then I have no idea when to use it..
Is the mesh optimizer and its permutation somehow involved?

Paulie14:
Update: VTK with identic mesh give different results only when running parallel

Paulie14:
[...]

PR #1369, opened 2021-08-12
Pe fix fieldfe msh

Paulie14:
Trying to simplify mesh reader.
First steps of issue #1366 before vacation..


src/io/msh_vtkreader.cc
```diff
@@ -525,32 +520,25 @@ void VtkMeshReader::create_node_element_caches() {
...
     HeaderQuery con_params("connectivity", 0.0, OutputTime::DiscreteSpace::MESH_DEFINITION);
-    auto & con_header = this->find_header(con_params);
+    auto con_header = this->find_header(con_params);
     con_header.n_entities = offsets_vec[offsets_vec.size()-1];
-    for (unsigned int i=bulk_elements_id_.size(); i<con_header.n_entities; ++i) bulk_elements_id_.push_back(i);
-    this->get_element_data<unsigned int>(con_header.n_entities, con_header.n_components, false, 0 );
+    this->get_element_data<unsigned int>(con_header, con_header.n_entities, con_header.n_components, false);
 
     has_compatible_mesh_ = false;
-	bulk_elements_id_.clear();
```
Paulie14, line 553: this part also seems to me like some dead code...
the assert and then setting `has_compatible_mesh_` to true does no make sense.
it seems to me, that `bulk_elements_id_` is not used anywhere for something useful (definitely in this function)..
then it seems that `has_compatible_mesh_` is also useless
and `find_header` and `get_element_data` can be merge together...

PR #1381, opened 2021-09-20
Js constraint matrix contact

jstebel:


jstebel:
Test case 50/26 with crossing fractures gives different results for different number of processes (error in rtol approx. 3-5 %). The difference in displacement/stress fields is mostly observed in fractures.
I have checked that at 1 and 2 processes:
- constraint matrices are equal
- stiffness matrices have equal eigenvalues/singular values
When contact is switched off, the results are equal on any number of processes.
So the source of the error may be related to PERMON.

Problem with 1 fracture works OK.

PR #1384, opened 2021-10-29
Df asm darcy flow

dflanderka:
- new assembly algorithm in DarcyFlowLMH and RichardsLMH


src/flow/assembly_lmh_old.hh
```diff
@@ -28,598 +47,688 @@
...
 
-            postprocess_velocity(dh_cell, reconstructed_solution_);
+        fe_ = std::make_shared< FE_P_disc<dim> >(0);
+        fe_values_side_.initialize(*this->quad_low_, *fe_, update_normal_vectors);
 
-            bc_fluxes_reconstruted = true;
-        }
+        fe_values_.initialize(quad_rt_, fe_rt_, update_values | update_JxW_values | update_quadrature_points);
+
+        ndofs_ = fe_values_.n_dofs();
```
jbrezmorf, line 167: Rather extract these as local variables where used. No performance hit as these are inlined anyway. Try to keep shared variables to a minimum.

src/flow/assembly_lmh_old.hh
```diff
@@ -28,598 +47,688 @@
...
+        qsize_ = fe_values_.n_points();
+        this->use_dirichlet_switch_ = true;
+
+        // local numbering of dofs for MH system
+        // note: this shortcut supposes that the fe_system is the same on all elements
+        // TODO the function should be DiscreteSpace.fe(ElementAccessor)
+        // and correct form fe(ad_->dh_->own_range().begin()->elm()) (see documentation of DiscreteSpace::fe)
+        auto fe = eq_data_->dh_->ds()->fe()[Dim<dim>{}];
+        FESystem<dim>* fe_system = dynamic_cast<FESystem<dim>*>(fe.get());
+        eq_data_->loc_side_dofs[dim-1] = fe_system->fe_dofs(0);
```
jbrezmorf, line 177: These could rather be local variables of the Assembly or even just local variables.

src/flow/assembly_lmh_old.hh
```diff
@@ -28,598 +47,688 @@
...
-    /// Saves the local system to a map: element index -> LocalSystem.
-    /// Currently used only for seepage BC.
-    void save_local_system(const DHCellAccessor& dh_cell)
-    {
-        // for seepage BC, save local system
-        if(save_local_system_)
-            ad_->seepage_bc_systems[dh_cell.elm_idx()] = loc_system_;
+    /// Setter of use_dirichlet_switch data member
+    inline void set_dirichlet_switch(bool ds) {
+        this->use_dirichlet_switch_ = ds;
```
jbrezmorf, line 190: Dirichlet switch was a hack to turn off a part of assembly, we should rather separate that part into separate methods, have separate assembly class for reconstruct assembly, but derived from the full assembly and reusing all methods, but avoid that particular method.

src/flow/assembly_lmh_old.hh
```diff
@@ -28,598 +47,688 @@
...
-                    //DebugOut().fmt("x: {}, to neum, p: {} f: {} -> f: {}\n", b_ele.centre()[0], bc_pressure, solution_flux, side_flux);
-                    switch_dirichlet=0;
+            if (use_dirichlet_switch_) {  // skip BC change if only reconstructing the solution
+                if (switch_dirichlet) {
+                    // check and possibly switch to flux BC
+                    // The switch raise error on the corresponding edge row.
+                    // Magnitude of the error is abs(solution_flux - side_flux).
+
+                    // try reconstructing local system for seepage BC
+                    load_local_system(cell_side.cell());
```
jbrezmorf, line 321: first make a pure function to assembly the local system we need here, then make it work with its duplicity cal, then possibly we find a new way how to cache it. We may even try to store all local systems as we now have to reassemble during the reconstruct assembly.

jbrezmorf (commented):
Just main points in darcy flow LMH. Have to see Richards yet.

src/flow/richards_lmh.cc
```diff
@@ -223,8 +224,54 @@ void RichardsLMH::assembly_linear_system()
...
+    this->mh_matrix_assembly_richards_->assemble(eq_data_->dh_);
+}
+
+
+void RichardsLMH::reconstruct_schur_asm() {
+    this->reconstruct_schur_assembly_richards_->assemble(eq_data_->dh_);
+}
+
+
+RichardsLMH::~RichardsLMH() {
```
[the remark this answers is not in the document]
dflanderka: There is one assemblation (InitCondPostprocessAssembly) implemented only in Richards. We cacan't remove this destructor. But other data are unified.

src/flow/assembly_lmh_old.hh
dflanderka: These 'dofs' variables are defined in EqData as std::array of size 3 and hold one item for each dimension. Values are 'global' on all elements of one dimension. Data are used in begin() method too where we iterate through elements of different dimensions and therefore we need to preset data of all dimensions.

src/flow/assembly_lmh_old.hh
```diff
@@ -28,598 +47,688 @@
...
-        //     mortar_assembly = std::make_shared<P1_CouplingAssembler>(ad_);
-        // }
+    typedef typename DarcyLMH::EqFields EqFields;
+    typedef typename DarcyLMH::EqData EqData;
 
-        // reconstructed vector for the velocity and pressure
-        // length = local Schur complement offset in LocalSystem
-        schur_offset_ = loc_edge_dofs[0];
-        reconstructed_solution_.zeros(schur_offset_);
+    static constexpr const char * name() { return "ReadInitCondAssemblyLMH"; }
```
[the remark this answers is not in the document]
jbrezmorf: Need for various time frames i GenericAssembly, no way to get `const char *` from `typeid::name()`.
Postponed.

src/flow/assembly_lmh_old.hh
```diff
@@ -22,604 +35,783 @@
...
+        for ( DHCellAccessor dh_cell : eq_data_->dh_->local_range() ) {
+            const ElementAccessor<3> ele = dh_cell.elm();
+            const DHCellAccessor dh_cr_cell = dh_cell.cell_with_other_dh(eq_data_->dh_cr_.get());
+
+            elm_dim = ele.dim();
+            size = (elm_dim+1) + 1 + (elm_dim+1); // = n_sides + 1 + n_sides
+            loc_size = size + ele->n_neighs_vb();
+            loc_size_schur = ele->n_sides() + ele->n_neighs_vb();
+            LocDofVec dofs(loc_size);
+            LocDofVec dofs_schur(loc_size_schur);
```
jbrezmorf, line 597: Check that `dofs[schur_offset:] == dofs_schur`. Get rid of `dofs_schur`.

src/flow/assembly_lmh_old.hh
```diff
@@ -22,604 +35,783 @@
...
-    void assemble_local_system(const DHCellAccessor& dh_cell, bool use_dirichlet_switch)
+
+
+    /// Common code of end method of MH matrix assembly (Darcy and Richards)
+    void end_mh_matrix()
     {
-        set_dofs(dh_cell);
+        for ( DHCellAccessor dh_cell : eq_data_->dh_->own_range() ) {
+            eq_data_->loc_system_[dh_cell.local_idx()].compute_schur_complement(
+                    eq_data_->schur_offset_[dh_cell.dim()-1], eq_data_->loc_schur_[dh_cell.local_idx()], true);
```
jbrezmorf, line 295: Make loc_schur a temporary system which is not saved for every element.

PR #1388, opened 2021-11-15
Df replace add potential

dflanderka:


jbrezmorf (commented):
Add remaining fields in similar fassion.

src/fields/field_add_potential.hh
```diff
@@ -22,6 +22,87 @@
...
+    {}
+
+    typename Field<spacedim,Value>::FieldBasePtr create_field(Input::Record rec, const FieldCommon &) override {
+        Input::AbstractRecord field_a_rec;
+        if (rec.opt_val(field_name_, field_a_rec)) {
+
+            FieldAlgoBaseInitData init_data(field_name_, Value::NRows_, UnitSI::dimensionless());
+            auto inner_field_ptr = FieldAlgorithmBase<spacedim, Value>::function_factory( field_a_rec, init_data );
+
+            // get domain specification
```
jbrezmorf, line 66: Deduplicate processing of the region and rid, use correct time.
- Attribute field_input_name_ in FactoryBase, pass in constructor.
- Drop the parameter input_name of method is_active_field_descriptor.
- change parameters of FieldBase::create_field(descriptor_rec, time, const RegionSet &domain, ?? FieldCommon)
- check if passing field is necessary
- deduplicate create_field and Update history using these new parameters

jbrezmorf: What exactly is the problem? The shared->input_name_ is set in FieldCommon::name() which is set during creation process of FieldCommon. Give more details.

dflanderka: Yes, shared->input_name_ is set in FieldCommon::name(), but FactoryBase is create in default constructor of Field<> (see field.impl.hh, line 48) and 'input_name' parameter is not to disposition in this moment. FieldCommon::name() method is called later.

PR #1399, opened 2021-12-06
Pe fix fieldfe msh

Paulie14:
- Leave the concept of components (e.g. vector of substances).
- Initial problem - could not write and again read msh data. (new tests 03-46, 03-47)
- Solves issue #1366 up to some details.


src/io/msh_vtkreader.cc
```diff
@@ -523,32 +518,22 @@ void VtkMeshReader::create_node_element_caches() {
...
-    for (unsigned int i=bulk_elements_id_.size(); i<offset_header.n_entities; ++i) bulk_elements_id_.push_back(i);
-    std::vector<unsigned int> &offsets_vec = *( this->get_element_data<unsigned int>(offset_header.n_entities, offset_header.n_components, false, 0 ) );
+	HeaderQuery offsets_query("offsets", 0.0, OutputTime::DiscreteSpace::MESH_DEFINITION);
+    std::vector<unsigned int> &offsets_vec =
+        *( this->get_element_data<unsigned int>(offsets_query, false) );
 
     // read connectivity data section
-    HeaderQuery con_params("connectivity", 0.0, OutputTime::DiscreteSpace::MESH_DEFINITION);
-    auto & con_header = this->find_header(con_params);
-    con_header.n_entities = offsets_vec[offsets_vec.size()-1];
```
Paulie14, line 546: after merging `find_header` and `get_element_data`, this abuse of the header is not possible and it causes problems..
not sure, how to overcome this.

src/fields/field_algo_base.impl.hh
```diff
@@ -47,7 +47,7 @@ template <int spacedim, class Value>
 FieldAlgorithmBase<spacedim, Value>::FieldAlgorithmBase(unsigned int n_comp)
 : value_(r_value_),
   field_result_(result_other),
-  component_idx_(std::numeric_limits<unsigned int>::max()),
+  component_idx_(0),
```
jbrezmorf, line 50: Point is that it should be set to undef_idx.

src/io/msh_basereader.cc
```diff
@@ -86,57 +86,49 @@ std::vector<int> const & BaseMeshReader::get_element_vector(bool boundary_domain
...
-		                tok_.f_name(), data_section_name_, field_name, actual_header_.time, actual_header_.n_components, n_components);
-	    		actual_header_.n_components=n_components;
-	    	}
-	    }
-
-    	(*element_data_values_)[field_name]
-					= std::make_shared< ElementDataCache<T> >(field_name, actual_header_.time, size_of_cache, n_components*n_entities);
-    	this->read_element_data(*(it->second), actual_header_, n_components, boundary_domain );
+        // check that the header is valid, try to correct n_components
+        if (header.n_components != expected_n_components) {
```
jbrezmorf, line 118: This should rather throw an exception.

jbrezmorf (commented):
Fix those two points and merge it.

src/io/msh_vtkreader.cc
Paulie14: reverted

src/fields/field_algo_base.impl.hh
Paulie14: done

src/io/msh_basereader.cc
Paulie14: done

Paulie14: and tested in unit test

PR #1405, opened 2021-12-13
Staz sb

jstebel:


jbrezmorf:
Following steps are necessary before merge:
- general interface to the Newton solver from PETSC
- usage of new assembly algorithm
- support for consistent residuum assembly and its differentiation, 
  can use just differentiation of parts of the residuum assembly.
- move the implementation to DarcyFlowLMH (DarcyFlowMH is obsolete)
- use preferably FieldModel if the basic numerical scheme is preserved
- tests

Issue #1422, opened 2022-01-28
Segfault in parallel run on boundary FieldFE.


dflanderka:
Parallel run of tests with boundary FieldFE throws Segfault in method DOFHandlerMultiDim::init_status. Problematic line is calling of Side::edge_idx(). Edges are not set on boundary mesh. The problem is in tests 02 / 23 and 02 / 24.

- suggest how we can switch off the loop over ghost_range on boundary mesh
- check other calling of Side::edge_idx() in DofHandler class whether they can cause an error

PR #1423, opened 2022-01-28
Add flag in dof handler to switch on/off distribution of edge dofs.

jstebel:
Small fix to make code more readable.


jstebel:
Resolves #1422.

PR #1432, opened 2022-03-11
Df asm observe

dflanderka:
- implementation of new assembly algorithm in observe output


src/fields/equation_output.hh
```diff
@@ -93,9 +97,29 @@ public:
      */
     void output(TimeStep step);
 
+    /// Getter for set of observe fields.
+    inline const std::unordered_set<string> &observe_fields() const {
```
jbrezmorf, line 101: We should not pass the whole EquationOutput to the observe assembly and implement those getters, rather pass necessary data through the constructor, possibly wrapping them into an auxiliary struct.

src/fields/field.impl.hh
```diff
@@ -398,8 +398,9 @@ void Field<spacedim, Value>::observe_output(std::shared_ptr<Observe> observe)
 
     if (observe->point_ds()->size() == 0) return;
```
jbrezmorf, line 400: observe_output would be necessary only for DarcyMH, so we should remove DarcyMH first (in master) and remove this method as well.

src/fields/field.hh
```diff
@@ -424,6 +433,9 @@ protected:
     /// ElementDataCache used during field output, object is shared with OutputTime
     std::shared_ptr<ElementDataCache<typename Value::element_type>> output_data_cache_;
 
+    /// Same as previous but used during observe output, object is shared with Observe
+    std::shared_ptr<ElementDataCache<typename Value::element_type>> observe_data_cache_;
```
jbrezmorf, line 437: Should not be in Field, rather copy directly to the Observe object.

src/fields/equation_output.hh
```diff
@@ -93,9 +97,29 @@ public:
...
+    /// Getter for OutputTime object.
+    inline Mesh *eq_mesh() {
+        return mesh_;
+    }
+
     /// Selects the error control field out of output field set according to input record.
     typename OutputMeshBase::ErrorControlFieldFunc select_error_control_field();
     
+    /// Getter from patch_point_data.
+    inline PatchPointVec &patch_point_data() {
```
jbrezmorf, line 119: Can we move this into Observe ?

src/fields/equation_output.cc
```diff
@@ -318,12 +323,20 @@ void EquationOutput::output(TimeStep step)
...
 
     // observe output
-    for(FieldCommon * field : this->field_list) {
-        if ( field->flags().match( FieldFlag::allow_output) ) {
-            if (observe_fields_.find(field->name()) != observe_fields_.end()) {
-                field->observe_output( observe_ptr );
+    {
+        FieldSet used_fields;
+        uint n_fields = 0;
+        for(FieldCommon * f_acc : this->field_list) {
```
jbrezmorf, line 329: Just form the output field set here and pass it to the assembly together with Observe instance.

src/fields/assembly_observe.hh
```diff
@@ -0,0 +1,235 @@
...
+
+
+private:
+    /// Calls cache_reallocate method on set of used fields
+    inline void reallocate_cache() {
+        multidim_assembly_[1_d]->eq_fields_->cache_reallocate(this->element_cache_map_, multidim_assembly_[1_d]->used_fields_);
+        // DebugOut() << "Order of evaluated fields (" << DimAssembly<1>::name() << "):" << multidim_assembly_[1_d]->eq_fields_->print_dependency();
+    }
+
+    std::shared_ptr<EvalPoints> eval_points_;                     ///< EvalPoints object
```
jbrezmorf, line 126: Aren't these in GenericAssembly? Why we have to declare them again?

src/fields/assembly_observe.hh
```diff
@@ -0,0 +1,235 @@
...
+struct PatchPointData {
+    /// Default constructor
+    PatchPointData() {}
+
+    /// Constructor with data mebers initialization
+    PatchPointData(unsigned int elm_idx, arma::vec loc_coords)
+    : elem_idx(elm_idx), local_coords(loc_coords), i_quad(0), i_quad_point(0) {}
+
+    /// Copy constructor
+    PatchPointData(const PatchPointData &other)
```
jbrezmorf, line 40: No constructors necessary for POD struct types.

Constructed simply by:
`PatchPointData x = {elm_idx, loc_coords, 0,0}`

or `emplace_back(idx, loc_coords, 0,0)`

jbrezmorf (changes requested):
Deserves some simplifications.

src/fields/assembly_observe.hh
jbrezmorf: Move to common parent class ?

src/fields/assembly_observe.hh
```diff
@@ -0,0 +1,235 @@
...
+    : AssemblyBase<dim>(), eq_fields_(eq_fields), eq_data_(eq_data) {
+        this->active_integrals_ = ActiveIntegrals::bulk;
+        offsets_.resize(1.1 * CacheMapElementNumber::get());
+
+        for (auto observe_field : eq_data_->observe_fields()) {
+            auto found_field = eq_fields_->field(observe_field);
+            used_fields_ += *found_field;
+        }
+
+        if (dim==1) {
```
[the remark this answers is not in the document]
jbrezmorf: Create `patch_point_data` in EquationOutput and pass to the Assembly as well.

src/fields/equation_output.hh
jbrezmorf: Move `patch_point_data_` vector into Observe class.

src/fields/field.impl.hh
dflanderka: observe_output is not used now, method will be removed, assembly classes used in equations and AssemblyObserveOutput class are independent.

dflanderka: observe_output is not used now, method will be removed, assembly classes used in equations and AssemblyObserveOutput class are independent.

PR #1447, opened 2022-05-11
Pe fix gmsh offset

Paulie14:
- start gmsh output node and element numbering at 1 for discontinuous output mesh (corner data present)
- update test 03-22 to involve different interpolation types at once


src/io/output_mesh.cc
```diff
@@ -123,7 +123,7 @@ void OutputMeshBase::create_id_caches()
 	OutputElementIterator it = this->begin();
 	for (unsigned int i = 0; i < this->n_elements(); ++i, ++it) {
         if (mesh_type_ == MeshType::orig) elm_idx[0] = orig_mesh_->find_elem_id(it->idx());
-		else elm_idx[0] = it->idx();
+		else elm_idx[0] = it->idx() + 1;    // GMSH starts at 1
```
jbrezmorf, line 126: When is this branch active? I.e. mesh_type != MeshType::orig.
It could be in the case of disconnected mesh or refined mesh.

Paulie14: yes, if `orig`, then the numbering is transfered from orig mesh, otherwise (disconnected or refined) we use new numbering, starting at 1

PR #1448, opened 2022-05-16
Df unit tests segfaults

dflanderka:
- fix unit tests after change version to 4.0.0


unit_tests/output/parallel_output_test.cpp
```diff
@@ -209,8 +209,8 @@ TEST_F(TestParallelOutput, discontinuous_mesh)
     make_output_mesh();
 
     /* Simulate field output */
-    data.init_scalar.field_output(stream, OutputTime::ELEM_DATA);
```
jbrezmorf, line 212: Why this change?

PR #1453, opened 2022-06-07
Js fix hm coupling source

jstebel:
Fix assembly of HM source - integrate over edges instead of elements. Should slightly improve the convergence of HM iterations.


jbrezmorf:
It would be better to refactor this into bulk assembly in a similar fashion as the Richards is done. It is not will require some direct DOF vector access as we currently dos not have appropriate support in the assembly algorithm. 
We should discuss that more precisely if there is a similarity with what is done in Richards.

jstebel:
> It would be better to refactor this into bulk assembly in a similar fashion as the Richards is done. It is not will require some direct DOF vector access as we currently dos not have appropriate support in the assembly algorithm. We should discuss that more precisely if there is a similarity with what is done in Richards.

There is a problem in bulk assembly, namely how to evaluate fields in different points than the one provided. This is the reason why I proposed to use edge_integral instead.

jstebel:
First, resolve issue #1475 before merging this pull-request.

Issue #1450, opened 2022-06-03
User fields


jbrezmorf:
- [ ] implement them for all equation
- [ ] create all user fields in `FieldSet::set_user_fields_map`, rename to `init_user_fields`, remove `user_field_input_map`
- [ ] Rename `FieldSet::get_user_field`, to `make_user_field_type`
- [ ] simplify FieldSet, merge `field_list` and `user_field_list`, 
    - check user field list for name conflict with the field_list
    - change ASSERT to ASSERT_PERMANENT in FieldSet::operator +=, improve error message
    -  to have common access to all fields, including user fields
- [ ] Allow output of user fields.
- [ ] In particular review DarchyMH Output and its output_fields and evaluation of a FieldFormula dependent

PR #1456, opened 2022-06-17
Df user fields

dflanderka:
Solves #1450


src/fields/field_set.hh
```diff
@@ -387,9 +389,11 @@ public:
     std::string print_dependency() const;
 
     /**
-     * Return pointer to the field defined by user given by name @p field_name. Return nullptr if not found.
+     * Collective interface to @p FieldCommon::set_default_fieldset().
      */
```
jbrezmorf, line 393: Describe meaning of  default field set.

src/mechanics/elasticity.cc
```diff
@@ -395,6 +399,12 @@ void Elasticity::initialize()
     eq_fields_->lame_lambda.set(Model<3, FieldValue<3>::Scalar>::create(fn_lame_lambda(), eq_fields_->young_modulus, eq_fields_->poisson_ratio), 0.0);
     eq_fields_->dirichlet_penalty.set(Model<3, FieldValue<3>::Scalar>::create(fn_dirichlet_penalty(), eq_fields_->lame_mu, eq_fields_->lame_lambda), 0.0);
 
+    // read optional user fields
+    Input::Array user_fields_arr;
+    if (input_rec.opt_val("user_fields", user_fields_arr)) {
+       	eq_fields_->init_user_fields(user_fields_arr, time_->step());
```
jbrezmorf, line 405: How to deal with the time step?

jbrezmorf: It seems that we have tests for reading the user fields in different equations. We should also test output of the user fields and test user fields of different shape. This common test would rather be in 03_output

src/flow/darcy_flow_lmh.cc
```diff
@@ -122,19 +121,22 @@ const it::Record & DarcyLMH::get_input_type() {
 
     DarcyLMH::EqFields eq_fields;
     
-    return it::Record("Flow_Darcy_LMH", "Lumped Mixed-Hybrid solver for saturated Darcy flow.")
+    return it::Record(equation_name(), "Lumped Mixed-Hybrid solver for saturated Darcy flow.")
 		.derive_from(DarcyFlowInterface::get_input_type())
         .copy_keys(EquationBase::record_template())
         .declare_key("gravity", it::Array(it::Double(), 3,3), it::Default("[ 0, 0, -1]"),
                 "Vector of the gravity force. Dimensionless.")
+        .declare_key("user_fields", it::Array(DarcyLMH::EqFields().make_user_field_type(equation_name())),
```
jbrezmorf, line 129: Move key definition into EquationBase.
Create there a method for processing this key and calling init_user_fields.

jbrezmorf (changes requested):

src/flow/darcy_flow_mh_output.cc
```diff
@@ -172,18 +170,29 @@ DarcyFlowMHOutput::DarcyFlowMHOutput(DarcyFlowInterface *flow, Input::Record mai
     }
 }
 
-void DarcyFlowMHOutput::prepare_output(Input::Record in_rec)
+void DarcyFlowMHOutput::prepare_output(Input::Record main_mh_in_rec)
 {
   	// we need to add data from the flow equation at this point, not in constructor of OutputFields
 	output_fields += darcy_flow->eq_fieldset();
```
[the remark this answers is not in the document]
dflanderka: Will be solved in #1483

src/fields/field_set.cc
```diff
@@ -39,7 +39,7 @@ const Input::Type::Record & FieldSet::get_user_field(const std::string &equation
                      "Name of user defined field.")
         .declare_key("is_boundary", Input::Type::Bool(), Input::Type::Default("false"),
                      "Type of field: boundary or bulk.")
-        .declare_key("scalar_field", scalar_field.get_input_type(), Input::Type::Default::obligatory(),
+        .declare_key("scalar_field", scalar_field.get_input_type(),
```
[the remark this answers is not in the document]
dflanderka: Will be solved in #1484

PR #1460, opened 2022-06-30
Df remove darcy mh

dflanderka:
Solves #1456.


src/flow/darcy_flow_lmh.cc
```diff
@@ -1362,6 +1530,21 @@ DarcyLMH::~DarcyLMH() {
 }
 
 
+/// Helper method fills range (min and max) of given component
```
jbrezmorf, line 1533: Check usage of this and the next method. Seems that it was used only in DarcyFlowMH.

jbrezmorf (commented):
- need to fix problem with dim =0 in FieldFE and FeValues
- detailed investigation of the numerical differences in other tests

PR #1489, opened 2022-11-07
Df asm darcy output

dflanderka:


src/flow/darcy_flow_mh_output.cc
```diff
@@ -230,6 +230,55 @@ void DarcyFlowMHOutput::prepare_specific_output(Input::Record in_rec)
...
+        unsigned int dim = mesh_->region_db().get_dim(i);
+        ASSERT_GT(dim, 0).error("Bulk region with dim==0!\n");
+        reg_by_dim[dim-1].push_back( mesh_->region_db().get_label(i) );
+    }
+
+    // Create instances of FieldPython and set them to Field objects
+    FilePath source_file( "analytical_module.py", FilePath::input_file);
+    double t = darcy_flow->time().t(); // initial time
+
+    std::shared_ptr< ScalarSolution > pressure_algo_1d = std::make_shared< ScalarSolution >();
```
jbrezmorf, line 255: template<FieldType>
void set_ref_solution(source_file, string python_method, output_field, reg) {
    std::shared_ptr< FieldType > algo = std::make_shared< FieldType >();
   algo->set_python_field_from_file( source_file, python_method);
    output_field.set(algo, t, reg);
}

PR #1500, opened 2022-12-15
Df fix read element cache

dflanderka:


src/io/msh_gmshreader.cc
```diff
@@ -265,30 +270,52 @@ void GmshMeshReader::read_element_data(ElementDataCacheBase &data_cache, MeshDat
...
+                // bulk
+                data_cache.read_ascii_data(tok_, header.n_components, (bulk_id_iter - bulk_el_ids.begin()) );
+                ++n_bulk_read;  ++bulk_id_iter;
+            } else if (*bdr_id_iter == (int)id) {
+            	// boundary
+                unsigned int bdr_shift = data_cache.get_boundary_begin();
+                data_cache.read_ascii_data(tok_, header.n_components, (bdr_id_iter - bdr_el_ids.begin() + bdr_shift) );
+                ++n_bdr_read;  ++bdr_id_iter;
+            } else {
+                if ( (*bulk_id_iter != imax) | (*bdr_id_iter != imax) )
```
jbrezmorf, line 289: Possibly should be (*bulk_id_iter == imax) | (*bdr_id_iter == imax). Test.

src/fields/field_fe.cc
```diff
@@ -671,50 +668,69 @@ void FieldFE<spacedim, Value>::interpolate_intersection(ElementDataCache<double>
...
+            if ( std::isnan(default_value_) )
+                THROW( ExcUndefElementValue() << EI_Field(field_name_) );
+            for (unsigned int i=0; i<loc_dofs.n_elem; ++i) {
+                ASSERT_LT(loc_dofs[i], (LongIdx)data_vec_.size());
+                data_vec_.add( loc_dofs[i], default_value_ * this->unit_conversion_coefficient_ );
+                ++count_vector[ loc_dofs[i] ];
+            }
+        } else {
+            if (native) data_vec_i = (source_idx + shift) * loc_dofs.n_elem;
+            else data_vec_i = (source_idx + shift) * dh_->max_elem_dofs();
```
jbrezmorf, line 702: Seems there is a bug in original code, that there should be dh_->max_elem_dofs() for the native case as well.

src/fields/field_fe.cc
```diff
@@ -671,99 +668,124 @@ void FieldFE<spacedim, Value>::interpolate_intersection(ElementDataCache<double>
...
-			}
-		}
-	}
+    for (auto cell : dh_->own_range()) {
+        LocDofVec loc_dofs = cell.get_loc_dof_indices();
+        //DebugOut() << cell.elm_idx() << " < " << source_target_vec.size() << "\n";
+        int source_idx = source_target_vec[cell.elm_idx()];
+
+        if (source_idx == (int)(Mesh::undef_idx)) { // undefined value in input data mesh
+            if ( std::isnan(default_value_) )
```
jbrezmorf, line 693: Test defualt is nan before the main loop.

PR #1508, opened 2023-04-17
Fix clearing observe_values_time_ for empty points_ in Observe.

jbrezmorf:


jbrezmorf:
This change helps to fix some issues with empty observe point lists. Yet the Observe point output is a bad design.

PR #1523, opened 2023-07-18
Df single bparser

dflanderka:
- remove matrix of parsers in FieldFormula
- fix yaml test files


tests/25_solute_dg_bc/01_inflow.yaml
```diff
@@ -32,7 +32,7 @@ problem: !Coupling_Sequential
           bc_conc: 3
         - region: domain
           diff_m: !FieldFormula
-            value: ' ((0) if (y>0.5) else (1e-2)) '
+            value: '[ [((0) if (y>0.5) else (1e-2)), 0, 0], [0, ((0) if (y>0.5) else (1e-2)), 0], [0, 0, ((0) if (y>0.5) else (1e-2))] ]'
```
jbrezmorf, line 35: Rewrite as "scalar * eye(3)", use manual unit tensor until `eye` is supported by bparser.

unit_tests/fields/field_algo_base_test.cpp
```diff
@@ -534,11 +534,11 @@ string field_input = R"INPUT(
    init_conc=[ 10, 20, 30],    // FieldConst
    conductivity={ //3x3 tensor
```
jbrezmorf, line 535: Revisit. Should be tensor here. Does the test pass?

dflanderka:
Fix field_python_test after merge of master

unit_tests/fields/field_algo_base_test.cpp
dflanderka: Input string is used in init_from_input test without assemblation. Test passes with both (vector and tensor) formula. But I changed input formula to tensor according to the comment.

Issue #1490, opened 2022-11-14
BParser uncatched exception


jbrezmorf:
With master branch, having FieldFormula: "<pressure_grad> @ X" failed since FieldFormula still using FParser.
however it fails with BParser error: 

"ERROR.Bparser, /home/jb/workspace/flow123d/third_party/bparser/include/ast.hh:246
                  Error: Undefined var: X"

Not sure how that exactly happens, but we should definitely catch any problems with unknown variables in FieldFormula.

jbrezmorf:
- `X` is coordinate vector
- could be used to prescribe a pressure gradient boundary condition as the pressure_head field
- - introduce a test for this kind of formula

jbrezmorf:
Possible input snippet:
```
flow_equation:
    ...
    input_fields:
    - piezo_head : "pressure_grad = [1, 2, 3]; pressure_grad @ X"
```

dflanderka:
Solved and used in test 02 / 31

PR #1524, opened 2023-07-26
Df coords vector

dflanderka:
- replace x, y, z coordinates with X


jbrezmorf (commented):
- modify test 31 to use flow_equation only, see comment in #1490
- use the correct shape for dependent fields in Fieldformula

PR #1528, opened 2023-08-18
Df benchmark asm

dflanderka:


unit_tests/coupling/assembly_speed_test.hh
```diff
@@ -0,0 +1,716 @@
+#ifndef ASSEMBLY_SPEED_TEST_HH_
```
jbrezmorf, line 1: Rename:
assembly_benchmark to DG_mockup.*
assembly_speed_test.hh to DG_mockup_assembly.hh

PR #1530, opened 2023-08-20
fix boost dependancy

jbrezmorf:
- fix only boost include for patch


jbrezmorf:
Seems already merged.

PR #1540, opened 2023-11-10
Df patch fevalues

dflanderka:


src/coupling/assembly_base.hh
```diff
@@ -69,6 +69,9 @@ public:
     /// Method finishes object after assemblation (e.g. balance, ...).
     virtual void end() {}
 
+    /// Method prepares object before computing on patch (typically reinitialize PatchFEValues objects).
+    virtual void patch_reinit(FMT_UNUSED const std::vector<unsigned int> &elm_idx_vec) {}
```
jbrezmorf, line 73: No need for virtual method, just one FeValues instance for bulk and one for side.

jbrezmorf:
Branch gets deleted.

PR #1567, opened 2024-09-19
Df eq data base

dflanderka:
- implement common parent of all EqData classes


jbrezmorf:
This issue was intended to streamline the setting order of quadrature that are specific for assembly, however are created by GenericAssembly. Properly deal with this issue needs rethink assembly creation process and rather try to create the assembly objects more explicitly, yet in a compact code. 

However, supporting multiple quadrature orders efficiently means having more granular control over which fields and FeValues are updated at which QPs.

PR #1576, opened 2024-11-06
Df patch fe optim

dflanderka:
- improvement of ElQ and FeQ

### DIFF
--- a/src/fem/patch_fe_values.hh
+++ b/src/fem/patch_fe_values.hh
@@ -427,15 +427,15 @@ public:
         uint n_points = patch_point_vals_->get_quadrature()->size();
         uint n_dofs = fe_component->n_dofs();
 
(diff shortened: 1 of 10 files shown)


src/fem/patch_fe_values.hh
```diff
@@ -140,24 +168,32 @@ public:
...
 
-    inline bool is_high_dim() const {
-        return (join_idx_ >= n_dofs_low_);
-    }
-
-    /// Iterates to next item.
-    inline void inc() {
-        join_idx_++;
+    /// Return local index of DOF (on low / high-dim) - should be private method
+    inline unsigned int local_idx(unsigned int i_join_idx) const {
```
jbrezmorf, line 172: See where this is used. It seems that most calls were eliminated.

PR #1599, opened 2025-07-31
Df accessors interface

dflanderka:
- application of new interface of Integral and operation accessors


src/fem/eval_subset.hh
```diff
@@ -27,6 +27,9 @@
 #include "mesh/range_wrapper.hh"
```
jbrezmorf, line 27: Split or Rename to "integral_acc" just before merge.

src/fem/op_accessors.hh
```diff
@@ -31,8 +31,9 @@ using Tensor = arma::mat33;
 template <class ValueType>
 class ElQ {
 public:
-    /// Forbidden default constructor
-    ElQ() = delete;
+    /// Ddefault constructor
```
jbrezmorf, line 34: Comment why default constructor is needed. We can allow it, but must be carefull since that leads to uninitialized instances.

PR #1621, opened 2026-01-21
Df patch fe optim

dflanderka:
- solves optimization of storing global and patch data of integral accessor objects

### DIFF
--- a/src/coupling/assembly_base.hh
+++ b/src/coupling/assembly_base.hh
@@ -212,8 +212,10 @@ public:
      * Method is called from GenericAssembly::assembly method.
      */
     virtual inline void assemble_cell_integrals() {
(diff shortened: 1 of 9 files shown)


src/coupling/assembly_base.hh
```diff
@@ -229,8 +222,11 @@ public:
      * Method is called from GenericAssembly::assembly method.
      */
     inline void assemble_boundary_side_integrals() {
-        for (unsigned int i=0; i<integral_data_.boundary_[0].permanent_size(); ++i) {
-            this->boundary_side_integral(integral_data_.boundary_[0][i].side);
+        uint n_patch_boundaries = integrals_.n_patch_boundaries();
```
jbrezmorf: Move patch_boundary_list = into loop avoiding the if and simlpify the code.

PR #1622, opened 2026-01-26
Df patch fe remaining

dflanderka:
- replacement of remaining FeValues usage in equations


jbrezmorf (commented):
Try to remove use_patch_fe_values flag in generic assembly.

PR #1627, opened 2026-05-07
Js docker image 4.0.5

jstebel:
Upgrade PETSc and other libraries in docker image.
Necessary fixes:
- Change PETSC_NULL to PETSC_NULLPTR.
- Fix VectorMPI::zero_entries() to properly clear VectorMPI values. Without it many tests get wrong results.


jbrezmorf (commented):
Reapply those commits with introduced explicit begin, end steps. 
We merge that and DF will then do a review of this pattern through the code base.

jstebel:
> Reapply those commits with introduced explicit begin, end steps. We merge that and DF will then do a review of this pattern through the code base.

Done.

PR #1637, opened 2026-07-01
Add water_source_sigma and water_source_ref_pressure fields.

jstebel:


src/flow/assembly_richards.hh
```diff
@@ -278,15 +283,15 @@ protected:
...
+                                            -diagonal_coef_*eq_fields_->water_source_sigma(p)-mass_diagonal_/eq_data_->time_step_,
                                             -source_diagonal_ - mass_rhs_);
             }
 
             eq_data_->balance_->add_mass_values(eq_data_->water_balance_idx, cell, {local_side},
                                                {0.0}, diagonal_coef_*water_content_vec.get(local_side));
             eq_data_->balance_->add_source_values(eq_data_->water_balance_idx, ele.region().bulk_idx(),
                                                 {this->eq_data_->loc_system_[cell.local_idx()].row_dofs[eq_data_->loc_edge_dofs[dim-1][i]]},
-                                                {0},{source_diagonal_});
+                                                {diagonal_coef_*eq_fields_->water_source_sigma(p)},{source_diagonal_});
```
jbrezmorf: both diagonal term and RHS term is assembled to the main system, but only diagonal term goes into balance. Is that correct?

src/flow/assembly_richards.hh
```diff
@@ -278,15 +283,15 @@ protected:
                              << "\n");
                 */
                 eq_data_->loc_system_[cell.local_idx()].add_value(eq_data_->loc_edge_dofs[dim-1][i], eq_data_->loc_edge_dofs[dim-1][i],
```
jbrezmorf, line 285: We should refactor it into a single function for both basic Darcy and Richards case.

jbrezmorf (commented):
Need to create a test for that.

src/flow/assembly_richards.hh
```diff
@@ -278,15 +283,15 @@ protected:
...
+                                            -diagonal_coef_*eq_fields_->water_source_sigma(p)-mass_diagonal_/eq_data_->time_step_,
                                             -source_diagonal_ - mass_rhs_);
             }
 
             eq_data_->balance_->add_mass_values(eq_data_->water_balance_idx, cell, {local_side},
                                                {0.0}, diagonal_coef_*water_content_vec.get(local_side));
             eq_data_->balance_->add_source_values(eq_data_->water_balance_idx, ele.region().bulk_idx(),
                                                 {this->eq_data_->loc_system_[cell.local_idx()].row_dofs[eq_data_->loc_edge_dofs[dim-1][i]]},
-                                                {0},{source_diagonal_});
+                                                {-diagonal_coef_*eq_fields_->water_source_sigma(p)},{source_diagonal_});
```
jstebel, line 294: The robin-type source enters the balance matrix (through diagonal_coef) as well as rhs (source_diagonal_).

src/flow/assembly_richards.hh
jstebel: The duplicated calculations of e.g. sources indeed should be unified. But as a whole, I think the assermbly of Richards has some major differences to Darcy and should be treated carefully.

PR #1638, opened 2026-07-10
Df reallocate asm

dflanderka:
- re-creating of assembly objects at each time step
- that allows to construct FieldFE operations correctly


jbrezmorf (commented):
turn assembly objects to temporary variables

PR #1641, opened 2026-08-13
Df field fe fix tests

dflanderka:


src/fem/eval_points.hh
```diff
@@ -193,26 +198,38 @@ private:
...
     private:
         Armor::Array<double> local_points_;                           ///< Local coords of points vector
         std::array<int, EvalPoints::max_subsets+1> subset_starts_;    ///< Indices of subsets data in local_points_ vector, used size is n_subsets_ + 1
-        std::vector<points_domain> points_domains_;                   ///< Flags hold if quadrature points are bulk or side, temporary data member of FieldFE patch operations
         unsigned int n_subsets_;                                      ///< Number of subset
         unsigned int dim_;                                            ///< Dimension of local points
 
+        // Following data members are temporary and they are used in evaluation of FieldFeOps.
+        // Data members will be removed after merge of Field and ParchFeValues caches.
+        std::vector<points_domain> points_domains_;                   ///< Flags hold if quadrature points are bulk or side
```
jbrezmorf, line 232: Mark by a comment tag e.g. [Remove after cache merge] the temporary code.

src/fem/patch_op_impl.hh
```diff
@@ -117,14 +117,16 @@ inline Tensor PatchOp<3>::point_value<Tensor>(uint point_idx, uint i_dof) const
 
 template <>
 template <>
-inline unsigned int PatchOp<3>::point_value<unsigned int>(FMT_UNUSED uint point_idx, FMT_UNUSED uint i_dof) const {
-    return 0;
+inline unsigned int PatchOp<3>::point_value<unsigned int>(uint point_idx, uint i_dof) const {
+	PatchPointValues<3> &ppv = patch_fe_->patch_point_vals_[domain_][dim_];
+    return result_(i_dof)(ppv.points_map_[point_idx]);
```
jbrezmorf, line 132: Add comment that this default conversion to int should not be done and these should return from "int_result_". Allocate only for FieldFeOp. In evaluation, fill both result_ and
int_result_ = (floor(result_)).

